PR #11, opened 2026-04-02
Add provider scheduling for time-based availability

Gajesh2007:
## Summary
- Providers can configure time windows when their machine serves inference (e.g., "weekday nights 10pm-8am", "all weekend")
- New `[schedule]` section in `provider.toml` with day-of-week, start/end time, overnight support
- Provider disconnects from coordinator and frees GPU memory outside schedule windows, reconnects automatically when the next window opens
- macOS app gets schedule editor UI in Settings > Availability tab with day toggles and time pickers

### Config example
```toml
[schedule]
enabled = true

[[schedule.windows]]
days = ["mon", "tue", "wed", "thu", "fri"]
start = "22:00"
end = "08:00"

[[schedule.windows]]
days = ["sat", "sun"]
start = "00:00"
end = "23:59"
```

### Files
- `provider/src/scheduling.rs` — New module: schedule parsing, `is_active_now()`, `duration_until_next_active()`, overnight window support
- `provider/src/config.rs` — Added `schedule: Option<ScheduleConfig>` to provider config
- `provider/src/main.rs` — Schedule-aware outer loop wrapping coordinator lifecycle
- `app/DGInf/Sources/DGInf/SettingsView.swift` — Schedule UI in Availability tab
- `app/DGInf/Sources/DGInf/StatusViewModel.swift` — Schedule state + config sync to TOML

## Test plan
- [x] All 131 provider tests pass (`cargo test`)
- [x] macOS app compiles (`swift build`)
- [x] Rust formatting clean (`cargo fmt`)
- [ ] Manual: set schedule, verify provider goes offline/online at window boundaries
- [ ] Manual: overnight window (22:00-08:00) crosses midnight correctly
- [ ] Manual: app schedule UI persists settings and writes to provider.toml

### DIFF
--- a/app/DGInf/Sources/DGInf/SettingsView.swift
+++ b/app/DGInf/Sources/DGInf/SettingsView.swift
@@ -134,11 +134,99 @@ private struct AvailabilityTab: View {
                 Text("Idle Detection")
                     .font(.headline)
             }
+
+            Section {
+                Toggle("Enable schedule", isOn: $viewModel.scheduleEnabled)
+
+                if viewModel.scheduleEnabled {
+                    ForEach($viewModel.scheduleWindows) { $window in
+                        ScheduleWindowRow(window: $window)
+                    }
+                    .onDelete { indices in
+                        viewModel.scheduleWindows.remove(atOffsets: indices)
+                    }
+
+                    Button {
+                        viewModel.scheduleWindows.append(ScheduleWindowModel.defaultWindow())
+                    } label: {
+                        Label("Add Window", systemImage: "plus")
+                    }
+                    .buttonStyle(.borderless)
+                }
+
+                Text("Set when your machine serves inference. Outside these windows, the provider disconnects and frees GPU memory. Requires provider restart to take effect.")
+                    .font(.caption)
+                    .foregroundColor(.secondary)
+            } header: {
+                Text("Schedule")
+                    .font(.headline)
+            }
         }
         .padding()
         .onAppear {
             selectedTimeout = viewModel.idleTimeoutSeconds
         }
+    }
+}
+
+// MARK: - Schedule Window Row
+
+private struct ScheduleWindowRow: View {
+    @Binding var window: ScheduleWindowModel
+
+    private let hours: [String] = (0..<24).map { String(format: "%02d:00", $0) }
+        + (0..<24).map { String(format: "%02d:30", $0) }
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 8) {
+            // Day selector
+            HStack(spacing: 4) {
+                ForEach(ScheduleWindowModel.allDays, id: \.self) { day in
+                    let isActive = window.activeDays.contains(day)
+                    Button {
+                        if isActive {
+                            window.activeDays.removeAll { $0 == day }
+                        } else {
+                            window.activeDays.append(day)
+                        }
+                    } label: {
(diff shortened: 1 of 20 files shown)


hankbobtheresearchoor:
## Hermes Agent Review
**Verdict:** Needs changes

### [CRITICAL]
- **TOML config sync uses regex to remove sections**: syncScheduleToConfig in StatusViewModel.swift uses a regex to strip existing [schedule] sections from the TOML file before rewriting them. This approach is fragile -- it could accidentally remove content if a section name partially matches, or break if the TOML file has comments or formatting that the regex does not account for. A single malformed edit could corrupt the entire provider config. Consider using a proper TOML parser/writer (like the Rust side already does) or a round-trip parse-modify-serialize approach.

### [WARNING]
- **Schedule evaluation uses local time**: The scheduling module uses Local::now() for all time checks, which depends on the system timezone. If a provider machine changes timezones or has its clock adjusted, the schedule behavior could change unexpectedly. For a distributed system, consider using UTC with an optional timezone offset configured by the user.

- **AtomicProviderStats uses Relaxed ordering**: The AtomicU64 counters use Ordering::Relaxed which does not provide synchronization guarantees with other memory operations. While acceptable for approximate counters, the inference_active AtomicBool also uses Relaxed ordering in some paths. If inference_active is used to gate scheduling decisions, consider using Ordering::Acquire/Release to ensure visibility across threads.

- **warmup_backend sends real inference request**: The warmup function sends a chat completion request (hi, max_tokens=1) to prime GPU caches. This consumes tokens and could have side effects on model state. Consider documenting that warmup is a best-effort operation and adding a config option to disable it.

- **check_model_loaded falls back to true on parse failure**: If the /health response cannot be parsed as JSON, check_model_loaded returns true, assuming the backend is healthy. This could mask a misconfigured backend that returns a non-JSON health response but is not actually ready for inference.

### [SUGGESTION]
- **Overnight window end boundary**: The is_active_now check for overnight windows uses time < w.end (strict less-than). This means a window of 22:00-08:00 becomes inactive exactly at 08:00:00. Consider whether the boundary should be inclusive (time <= w.end) to match user expectations.

- **duration_until_inactive for overnight is imprecise**: When computing duration_until_inactive for an overnight window that started yesterday, the code adds secs_until_midnight + time_to_secs(w.end). However, this calculation does not account for the current seconds-within-the-minute, which could cause the duration to be off by up to 60 seconds. Not critical for scheduling but worth noting.

- **SwiftUI schedule UI lacks validation**: The ScheduleWindowRow allows arbitrary time selection but does not validate that start differs from end, or that at least one day is selected. A window with no days or identical start/end would be silently ignored by the Rust parser.

- **Large PR scope**: This PR combines scheduling, atomic stats refactor, backend warmup, model-loaded health checks, and substantial code formatting changes. Consider splitting into smaller PRs for easier review -- e.g., one for the scheduling module, one for the stats refactor, one for warmup.

### [LOOKS GOOD]
- **Schedule module design**: The ParsedWindow with overnight detection and the is_active_now logic with yesterday-fallback for overnight windows is well thought out.
- **Backend warmup concept**: Priming GPU caches before serving real traffic is a good operational practice that eliminates first-token latency penalties.
- **Graceful schedule transitions**: The duration_until_inactive and duration_until_next_active methods enable smooth connect/disconnect behavior.
- **Config roundtrip test**: The TOML serialization roundtrip test for ScheduleConfig ensures backward compatibility.

PR #24, opened 2026-04-15
fix(stt): support tokenizer.json for cohere-transcribe-03-2026 (mlx_audio ≤ 0.4.2 compat)

guglxni:
## Summary

- `mlx_audio` ≤ 0.4.2 `post_load_hook` hardcodes `tokenizer.model` (SentencePiece), but `CohereLabs/cohere-transcribe-03-2026` ships only `tokenizer.json` (HuggingFace fast tokenizer), causing a crash at server startup.
- Adds `_CohereAsrTokenizerHF` — a drop-in shim over HuggingFace `tokenizers` implementing the same interface (`eos_token_id`, `build_prompt_tokens`, `batch_decode`) as the existing `CohereAsrTokenizer`.
- Monkey-patches `Model.post_load_hook` inside `load_model()` to probe `tokenizer.model` first, then fall back to `tokenizer.json` — same pattern already used for `load_buffers_from_checkpoint`.

## Root cause

```python
# mlx_audio/stt/models/cohere_asr/cohere_asr.py  line 705
tokenizer_path = model_path / "tokenizer.model"   # ← always SentencePiece
model._tokenizer = CohereAsrTokenizer(str(tokenizer_path), ...)
# CohereAsrTokenizer.__init__ calls spm.SentencePieceProcessor().load(path)
# → OSError / RuntimeError because the file doesn't exist
```

`CohereLabs/cohere-transcribe-03-2026` (released March 2026) moved to the HuggingFace `tokenizers` JSON format; `mlx_audio` 0.4.2 was not updated to match.

## Test plan

- [ ] `python stt_server.py --model CohereLabs/cohere-transcribe-03-2026 --port 8101` starts without error
- [ ] `POST /v1/audio/transcriptions` returns a transcript for a WAV file
- [ ] Models that still ship `tokenizer.model` (SentencePiece) continue to load via the original path

## Notes

The R2 bundle ships `mlx_audio` 0.2.7 (no `cohere_asr` module at all). This fix is only exercised on installs that have upgraded to 0.4.2 and are targeting `cohere-transcribe-03-2026`. A proper upstream fix in `mlx_audio` would be preferable long-term.

### DIFF
--- a/provider/stt_server.py
+++ b/provider/stt_server.py
@@ -320,6 +320,71 @@ async def stats():
 
 
 # ---------------------------------------------------------------------------
+# HuggingFace fast-tokenizer shim for cohere-transcribe-03-2026
+# ---------------------------------------------------------------------------
+
+class _CohereAsrTokenizerHF:
+    """Drop-in replacement for CohereAsrTokenizer when the model ships
+    tokenizer.json (HuggingFace fast tokenizer) instead of tokenizer.model
+    (SentencePiece).  CohereLabs/cohere-transcribe-03-2026 is the first
+    release to do this; mlx_audio ≤ 0.4.2 doesn't handle it.
+    """
+
+    def __init__(self, tokenizer_json_path: str):
+        from tokenizers import Tokenizer  # pip install tokenizers
+
+        self._tok = Tokenizer.from_file(tokenizer_json_path)
+        self.vocab_size = self._tok.get_vocab_size()
+
+    def _id(self, token: str) -> int:
+        tid = self._tok.token_to_id(token)
+        if tid is None:
+            raise KeyError(f"Token not in vocab: {token!r}")
+        return tid
+
+    @property
+    def eos_token_id(self) -> int:
+        return self._id("<|endoftext|>")
+
+    @property
+    def bos_token_id(self) -> int:
+        return self._id("<|startoftranscript|>")
+
+    @property
+    def pad_token_id(self) -> int:
+        return self._id("<pad>")
+
+    @property
+    def unk_token_id(self) -> int:
+        return self._id("<unk>")
+
+    def build_prompt_tokens(self, language: str, punctuation: bool = True) -> list:
+        tokens = [
+            "<|startofcontext|>",
+            "<|startoftranscript|>",
+            "<|emo:undefined|>",
+            f"<|{language}|>",
+            f"<|{language}|>",
+            "<|pnc|>" if punctuation else "<|nopnc|>",
+            "<|noitn|>",
+            "<|notimestamp|>",
+            "<|nodiarize|>",
+        ]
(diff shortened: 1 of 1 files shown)


guglxni:
@Gajesh2007 could you approve the workflow run for this PR? Thanks!

hankbobtheresearchoor:
## Hermes Agent Review

### ⚠️ Warning

1. **`_id()` raises `KeyError` on missing tokens — will crash the server at runtime** — Properties like `eos_token_id`, `bos_token_id`, `pad_token_id`, and `unk_token_id` call `self._id(token)` which raises `KeyError` if the token is not in the vocabulary. If the tokenizer.json uses different special token names than expected (which is likely for future models), the STT server will crash on startup rather than falling back gracefully. Consider returning `None` / a sentinel value and logging a warning instead.

2. **Monkey-patching `post_load_hook` on the class object is a global mutation** — The patch `_cohere_asr_mod.Model.post_load_hook = _patched_post_load_hook` modifies the class in-place for the entire process lifetime. If `load_model` is called multiple times (e.g., for different models), the patch is re-applied each time. While functionally idempotent (it replaces with the same patched version), this is fragile if the upstream class structure changes. Consider patching once at module import time with a guard.

3. **No integrity validation of `tokenizer.json` before loading** — `Tokenizer.from_file()` deserializes an untrusted JSON file from the model directory. If a compromised model repository serves a malicious tokenizer.json (e.g., one that causes excessive memory allocation or infinite loops during tokenization), there is no size or schema check before loading. While this is a pre-existing pattern from the upstream code, the new code path extends the attack surface.

### 💡 Suggestion

4. **`build_prompt_tokens` has hardcoded language token duplication** — The list includes `f"<|{language}|>"` twice in a row. If this is intentional (the original SentencePiece tokenizer may require it), add a comment explaining why. If not, it is a bug that could cause unexpected decoding behavior.

5. **Add a startup log line when the HF tokenizer path is taken** — When the patched `post_load_hook` falls back to `tokenizer.json`, log the path taken so operators can verify which tokenizer is active. This aids debugging in heterogeneous model deployments.

6. **Consider contributing upstream to `mlx_audio`** — The PR description notes this is a local workaround. Filing an issue and PR against `mlx_audio` would benefit the broader community and reduce maintenance burden here.

### ✅ Looks Good

7. The fallback pattern (try `tokenizer.model` first, then `tokenizer.json`) preserves backward compatibility with existing SentencePiece models.

8. The `_CohereAsrTokenizerHF` class faithfully mirrors the `CohereAsrTokenizer` interface, which is the right approach for a drop-in shim.

9. The negative ID filter in `decode` and `batch_decode` (`int(i) for i in ids if int(i) >= 0`) is a reasonable safeguard against padding/special token IDs.

PR #31, opened 2026-04-16
fix: validate provider-reported token counts to prevent billing inflation

### LINKED ISSUE #30: Security: Provider can inflate token counts to overcharge consumers (billing fraud)
## Summary

A malicious provider can report inflated `completion_tokens` in the `InferenceComplete` WebSocket message, causing consumers to be overcharged and the provider to receive inflated payouts. The coordinator trusts the provider's self-reported token counts for billing without any validation.

## Severity

High — financial exploit allowing systematic billing fraud.

## Root Cause

`coordinator/internal/api/provider.go`, function `handleComplete()`:

```go
totalCost := payments.CalculateCostWithOverrides(
    pr.Model, msg.Usage.PromptTokens, msg.Usage.CompletionTokens, ...)
providerPayout := payments.ProviderPayout(totalCost)
```

The provider-reported `PromptTokens` and `CompletionTokens` are used directly for cost calculation. There is no coordinator-side validation — no chunk counting, no comparison with `max_tokens`, no sanity check.

## Confirmed Exploitation

Against a local coordinator at `trust=hardware` (production setting):

- Sent: `"hello"` (1 word)
- Real response: `"Echo: hello"` (~2 tokens)
- **Provider reported: 100,000 completion_tokens**
- Consumer charged: **$0.02** (should be $0.0001)
- Overcharge factor: **200x**
- 6 requests drained **$0.12** from $10,000 balance

At scale: 1,000 requests/day = $20/day stolen per victim. Provider keeps 95%.

The coordinator logged the inflated token counts without warning. The consumer has no visibility into the discrepancy.

## Affected Versions

All versions including v0.3.9 (checked latest master).

## Proposed Fix

Branch with patch: [`farazshaikh/d-inference:fix/token-count-validation`](https://github.com/farazshaikh/d-inference/tree/fix/token-count-validation)

Commit: `8f18a80a`

The fix adds coordinator-side validation:
1. Counts content-bearing SSE chunks forwarded during streaming (each ≈ 1 token)
2. On `InferenceComplete`, caps reported `completion_tokens` at `max(observed_chunks * 10, requested_max_tokens)`
3. Logs a warning and records a job failure (degrades provider reputation) when inflation is detected

For non-streaming requests, the consumer's `max_tokens` serves as the upper bound.

## Additional Recommendations

- Add token count validation for `prompt_tokens` too (provider could inflate input billing)
- Consider independent token estimation using a fast tokenizer on the coordinator
- Rate-limit or flag providers with repeated billing anomalies
- Add consumer-facing usage alerts when charges significantly exceed estimates

farazshaikh:
## Summary

Adds coordinator-side validation of provider-reported token counts to prevent billing inflation attacks.

Closes #30

## The Problem

Providers self-report `prompt_tokens` and `completion_tokens` in the `InferenceComplete` message. The coordinator uses these directly for billing with zero validation. A malicious provider can report 100,000 tokens for a 2-token response and the consumer gets overcharged 200x.

## The Fix

Two files changed:

**`coordinator/internal/registry/registry.go`**
- Added `ChunksForwarded int64` counter to `PendingRequest`

**`coordinator/internal/api/provider.go`**
- `handleChunk()`: increments `ChunksForwarded` for each content-bearing SSE chunk
- `handleComplete()`: validates reported `completion_tokens` against `max(observed_chunks * 10, requested_max_tokens)`, caps if exceeded, logs warning, records job failure

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| Honest provider (100 tokens) | Charged 100 | Charged 100 |
| Inflated report (100K tokens, 5 chunks) | Charged 100K | **Capped to max_tokens or 50** |
| Non-streaming (no chunks) | Charged whatever reported | Capped to requested max_tokens |

The 10x tolerance on chunk counting accounts for tokenization variance (sub-word tokens, batch chunks). Anything beyond 10x is clearly malicious.

## Test plan

- [ ] Existing billing tests pass
- [ ] Honest providers unaffected (reported tokens < max_allowed)
- [ ] Inflated token count gets capped and logged
- [ ] Provider reputation degraded on inflation detection
- [ ] Non-streaming requests capped at requested max_tokens

Made with [Cursor](https://cursor.com)

### DIFF
--- a/coordinator/internal/api/provider.go
+++ b/coordinator/internal/api/provider.go
@@ -30,7 +30,9 @@ import (
 	"encoding/json"
 	"log/slog"
 	"net/http"
+	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/eigeninference/coordinator/internal/attestation"
@@ -624,6 +626,14 @@ func (s *Server) handleChunk(providerID string, provider *registry.Provider, msg
 		s.logger.Warn("chunk for unknown request", "provider_id", providerID, "request_id", msg.RequestID)
 		return
 	}
+	// Count content-bearing chunks for billing validation.
+	// Each SSE "data:" line with a delta.content field represents roughly
+	// one output token. This is an approximation but sufficient to detect
+	// order-of-magnitude inflation by malicious providers.
+	if strings.Contains(msg.Data, `"content"`) {
+		atomic.AddInt64(&pr.ChunksForwarded, 1)
+	}
+
 	// Non-blocking send — if consumer is gone the chunk is dropped.
 	select {
 	case pr.ChunkCh <- msg.Data:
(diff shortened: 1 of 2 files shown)


hankbobtheresearchoor:
## Hermes Agent Review
**Verdict:** Needs changes

### 🔴 Critical

1. **Silently ignoring errors on financial operations** — Both `s.store.Credit(...)` and `s.store.RecordProviderEarning(...)` use `_ =` to discard errors. These are **billing-impacting operations** — a failed credit means a provider does real work and never gets paid. The old code logged errors; the new code silently drops them. Same issue with `s.ledger.CreditProvider(...)` in the wallet path and image generation handler. **Every financial operation must log errors at minimum, and ideally should retry or enter a dead-letter queue.**

2. **Lost atomicity in credit + earning record** — The old `CreditProviderAccount` performed the credit and earning record in a single store transaction. The new code splits this into two independent calls (`Credit` then `RecordProviderEarning`), creating a consistency gap: if `Credit` succeeds but `RecordProviderEarning` fails (or vice versa), the books diverge. For a billing system, this is a data integrity risk. Either restore the atomic transaction or add compensating logic.

### ⚠️ Warning

3. **Chunk-based billing validation is easily gamed** — `strings.Contains(msg.Data, "\"content\"")` detects "content-bearing" chunks, but a malicious provider can inject `"content"` into non-content fields (e.g., tool call arguments, metadata) to inflate the observed chunk count, or structure output so content is spread across fewer chunks. The 10x tolerance makes this even more permissive. This heuristic provides a *lower* bound on detection, not a reliable one.

4. **`TrustHardware` attestation in PoC mode is misleading and dangerous** — `provider.SetAttested(true, registry.TrustHardware)` is set without actual hardware-rooted MDM verification. Any code downstream that checks `TrustHardware` to make routing or billing decisions will trust a provider that only has self-signed attestation. If this is truly PoC-only, it should be behind a build flag or config toggle, not silently committed to the trust level.

5. **Non-streaming billing cap uses `requested_max_tokens` as upper bound** — For non-streaming requests, `maxAllowed` falls back to `pr.RequestedMaxTokens` (default 4096). A malicious provider could set `max_tokens` very high and then inflate tokens up to that cap. The coordinator should also validate the *consumer-requested* max_tokens against reasonable model limits.

6. **Image generation handler no longer credits Privy-linked accounts** — The `handleImageGenerationComplete` function now only credits via `s.ledger.CreditProvider(providerWallet, ...)`, skipping the `p.AccountID != ""` path entirely. Providers linked via Privy will not receive earnings for image generation jobs. This appears to be an unintended regression.

### 💡 Suggestion

7. **Add metrics for billing validation** — When token inflation is detected, emit a counter metric (not just a log line) so that alerting can trigger on aggregate patterns across providers.

8. **Consider per-model max_token limits** — Hardcoding 4096 as a default cap may be too low for some models (e.g., long-context models) and too high for others. Model-specific limits would be more precise.

### ✅ Looks Good

9. The general approach of coordinator-side validation of self-reported token counts is sound and necessary. The `ChunksForwarded` counter on `PendingRequest` is a good primitive.

10. `RecordJobFailure` on inflation detection properly degrades provider reputation, creating a deterrent.

Gajesh2007:
Fixed

PR #51, opened 2026-04-17
fix(provider): resolve FLUX.2 Klein pipeline by s3_name, not legacy cache path

### LINKED ISSUE #23: v0.3.8: Not able to download FLUX.2 Klein 4B, getting "The requested URL returned error: 404"
Below are the logs: 
  
Hardware:
    Chip:       Apple M2 Pro
    Memory:     16 GB total, 12 GB available
    GPU:        19 cores
    Bandwidth:  200 GB/s

  Security: System Integrity Protection is enabled
    SIP:            ✓ Enabled
    Secure Enclave: ✓ Available
    Enclave key:    ✓ Generated
    MDM enrolled:   ✓ Yes (hardware trust)

  Account:
    Linked:   ✓ Yes

Select models to download (12 GB available):

  [1]   8.1 GB  FLUX.2 Klein 4B
  [✗]  17.4 GB  FLUX.2 Klein 9B (too large)
  [✗]  27.0 GB  Qwen3.5 27B Claude Opus Distilled (too large)
  [✗]  28.0 GB  Gemma 4 26B (too large)
  [✗]  26.0 GB  Trinity Mini (too large)
  [✗] 122.0 GB  Qwen3.5 122B (too large)
  [✗] 243.0 GB  MiniMax M2.5 (too large)
  [✓]   4.2 GB  Cohere Transcribe (already downloaded)

  Enter numbers to download (comma-separated, e.g. 1,3):
  1

  Downloading FLUX.2 Klein 4B...
curl: (56) The requested URL returned error: 404                                                           

  ⚠ Unknown image model

lukevenediger:
## Summary
- `download_ckpt_model_from_cdn` matched the FLUX.2 Klein pipeline against `cache_dir.contains("flux_2_klein_{4,9}b_q8p")`, but `cmd_models` writes image models to `~/.darkbloom/models/<s3_name>` (e.g. `flux-klein-4b-q8`). The match always fell through to the `None` arm, so users saw `⚠ Unknown image model` after the tarball fallback 404'd.
- Match on a combined key of `base_url` (carries `s3_name`) and `cache_dir`, accepting both dashed (`klein-4b`/`klein-9b`) and underscored (`klein_4b`/`klein_9b`) forms so the legacy `~/.cache/huggingface/hub/models--flux_2_klein_4b_q8p.ckpt/...` layout still resolves.

Fixes #23.

## Test plan
- [x] `darkbloom models` → `1` (Download) → `1` (FLUX.2 Klein 4B): VAE completes, diffusion model begins streaming from R2 (`flux-klein-4b-q8`) at ~70+ MB/s on Apple M4 Pro.
- [x] Verified R2 URLs return `200 OK` for `flux_2_klein_4b_q8p.ckpt`, `flux_2_vae_f16.ckpt`, `qwen_3_4b_q8p.ckpt`.
- [ ] Confirm FLUX.2 Klein 9B downloads through the same path.
- [ ] Smoke-check legacy HF-cache layout still resolves (`klein_4b`/`klein_9b` branch).

### DIFF
--- a/provider/src/main.rs
+++ b/provider/src/main.rs
@@ -510,7 +510,11 @@ fn download_ckpt_model_from_cdn(
         name: &'static str,
     }
 
-    let pipeline = if cache_dir.to_string_lossy().contains("flux_2_klein_9b_q8p") {
+    // Match on base_url (contains s3_name) or cache_dir (legacy HF cache layout),
+    // so both `~/.darkbloom/models/flux-klein-9b-q8` and
+    // `~/.cache/huggingface/hub/models--flux_2_klein_9b_q8p.ckpt/...` resolve correctly.
+    let key = format!("{} {}", base_url, cache_dir.to_string_lossy());
+    let pipeline = if key.contains("klein-9b") || key.contains("klein_9b") {
         Some(ImagePipeline {
             model_file: "flux_2_klein_9b_q8p.ckpt",
             text_encoder: "qwen_3_8b_q8p.ckpt",
@@ -518,7 +522,7 @@ fn download_ckpt_model_from_cdn(
             version: "flux2_9b",
             name: "FLUX.2 [klein] 9B",
         })
-    } else if cache_dir.to_string_lossy().contains("flux_2_klein_4b_q8p") {
+    } else if key.contains("klein-4b") || key.contains("klein_4b") {
         Some(ImagePipeline {
             model_file: "flux_2_klein_4b_q8p.ckpt",
             text_encoder: "qwen_3_4b_q8p.ckpt",


hankbobtheresearchoor:
## Hermes Agent Review

### ⚠️ Warning

1. **Substring matching is fragile and could match unintended model names** — The check `key.contains("klein-9b")` would also match a hypothetical future model like `flux-klein-9b-lora` or `klein-9b-quantized`. Similarly, the combined `key = format!("{} {}", base_url, cache_dir.to_string_lossy())` means a URL containing `klein-9b` in a path segment unrelated to the model (e.g., a bucket name or CDN prefix) could trigger a false match. Consider using more precise matching (e.g., exact segment match or regex with word boundaries).

2. **No test coverage for the matching logic** — The pipeline selection is a critical dispatch path (wrong pipeline = wrong inference behavior), yet there are no unit tests verifying that various URL/cache-dir combinations resolve to the correct pipeline. A few test cases covering `~/.darkbloom/models/flux-klein-4b-q8`, `~/.cache/huggingface/hub/models--flux_2_klein_4b_q8p.ckpt/...`, and a non-matching path would go a long way.

### 💡 Suggestion

3. **Consider a lookup table instead of chained `if`** — As more models are added, the `if/else if` chain will grow. A `HashMap` or slice of `(pattern, ImagePipeline)` would be more maintainable and easier to test in isolation.

4. **Log when pipeline is None** — The original code already had an `Unknown image model` warning path, but it would be helpful to log the `key` value that failed to match, making it easier to diagnose future path resolution issues.

### ✅ Looks Good

5. The core fix is correct — `cache_dir` alone was insufficient because `cmd_models` stores models under `~/.darkbloom/models/<s3_name>`, not the HuggingFace cache layout. Including `base_url` in the match key resolves the root cause.

6. Supporting both dashed and underscored forms ensures backward compatibility with existing HuggingFace cache directories.

PR #52, opened 2026-04-17
Fix STT-only model selection in darkbloom start

### LINKED ISSUE #42: MacBook Air M4 16GB - Cohere Transcribe Model
ModuleNotFoundError: No module named 'mlx_audio.stt.cohere_asr'
"No models selected" loop

Image

cryptosingheth:
## Summary

Fixes #42 for the `No models selected` loop when selecting only `Cohere Transcribe [transcription]`.

The issue was reproduced on a Mac Mini, but the root cause is not Mac Mini-specific. `hardware::detect()` only controls which catalog entries fit the machine's RAM; the failure happens later in generic startup selection validation after the picker has already returned selected entries. Any Mac provider that selects only an STT/transcription model can hit the same error.

The separate `ModuleNotFoundError: No module named 'mlx_audio.stt.cohere_asr'` symptom mentioned in #42 is a runtime/package issue and is not covered by this PR.

## Root Cause

`cmd_start` already split selected picker entries into:

- text models for `--model`
- image models for `--image-model`
- transcription/STT models for `EIGENINFERENCE_STT_MODEL`

However, the final validation only checked text and image selections:

```rust
if selected_models.is_empty() && final_image_model.is_none() {
    anyhow::bail!("No models selected");
}
```

So `picked_stt` could be populated correctly, but STT-only startup still failed before launchd service installation.

## Changes

- Adds a small `has_start_selection` helper that treats text, image, and STT selections as valid startup workloads.
- Updates the final `darkbloom start` guard to include `picked_stt`.
- Prints the selected STT model in the startup summary.
- Adds focused unit tests for text-only, image-only, STT-only, and empty selections.
- Keeps the branch on current upstream `master`; while validating that base, also applies formatter-only cleanup in `crypto.rs` / `secure_enclave_key.rs` and gates the in-process engine scratch variable behind the existing `python` feature so `--no-default-features` compiles.

## Validation

- `cd provider && cargo fmt --check`
- `cd provider && cargo test --no-default-features`

Note: an initial sandboxed test run failed because local mock-server socket binding was denied with `Operation not permitted`; the same test command passed when rerun outside the sandbox.

### DIFF
--- a/provider/src/crypto.rs
+++ b/provider/src/crypto.rs
@@ -150,13 +150,10 @@ pub fn legacy_enclave_e2e_key_paths() -> Vec<std::path::PathBuf> {
 }
 
 fn purge_legacy_e2e_files() {
-    for path in [
-        legacy_node_key_paths(),
-        legacy_enclave_e2e_key_paths(),
-    ]
-    .into_iter()
-    .flatten()
-    .filter(|path| path.exists())
+    for path in [legacy_node_key_paths(), legacy_enclave_e2e_key_paths()]
+        .into_iter()
+        .flatten()
+        .filter(|path| path.exists())
     {
         match std::fs::remove_file(&path) {
             Ok(()) => tracing::info!("Removed legacy E2E secret file: {}", path.display()),
--- a/provider/src/main.rs
+++ b/provider/src/main.rs
@@ -3465,6 +3465,7 @@ async fn cmd_serve(
                                         .map(|s| (s.model_id.clone(), s.model_path.clone(), s.port, s.pid, s.healthy, s.restarting))
                                 };
 
+                                #[cfg(feature = "python")]
                                 let mut inprocess_engine = None;
                                 if let Some((slot_model_id, slot_model_path, slot_port, slot_pid, slot_healthy, slot_restarting)) = slot_info {
                                     if is_inprocess {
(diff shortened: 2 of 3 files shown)


hankbobtheresearchoor:
## Hermes Agent Review

### ⚠️ Warning

1. **`#[cfg(feature = "python")]` on `inprocess_engine` declaration only** — The `mut inprocess_engine` variable is gated behind the feature flag, but if any code *uses* `inprocess_engine` below without a matching `#[cfg(feature = "python")]` guard, the `--no-default-features` build will fail. The PR mentions `cargo test --no-default-features` passes, which is encouraging, but worth double-checking that every usage site (reads, writes, moves) is also conditionally compiled.

2. **Formatter-only changes mixed with functional changes** — The `crypto.rs` and `secure_enclave_key.rs` diffs are pure reformatting (line wrapping). This makes the functional diff harder to review. Consider splitting formatting commits into a separate PR or at least a separate commit for cleaner bisectability.

### 💡 Suggestion

3. **`has_start_selection` is clean but could be a method** — Since it operates on the same data that `cmd_start` already decomposes, consider making it a method on a small struct (e.g., `ModelSelection { text, image, stt }`) to reduce the chance of parameter list drift as more model types are added.

### ✅ Looks Good

4. The root cause fix is correct — STT-only selection was impossible because the guard only checked `selected_models` and `final_image_model`. The new `has_start_selection` properly includes `picked_stt`.

5. Unit tests for all four cases (text-only, image-only, STT-only, empty) provide good coverage of the validation logic.

6. The startup summary now prints the selected STT model, which improves observability for operators.

PR #54, opened 2026-04-17
fix: install.sh PBS fallback never triggers when coordinator 404s

### LINKED ISSUE #45: Install script: Python runtime 404 → PBS fallback never triggers
## Bug

The install script (`install.sh`) fails to install the Python runtime because both coordinator download URLs return **HTTP 404**, and the python-build-standalone (PBS) fallback is never reached due to a logic gap.

## Environment

- macOS 26.3.1 / 26.4.1 (tested on 3 machines)
- Apple M3 Max 128 GB, Apple M4 16 GB × 2
- Darkbloom v0.3.8 (latest as of 2026-04-17)

## Steps to Reproduce

```bash
curl -fsSL https://api.darkbloom.dev/install.sh | bash
```

## Observed

```
→ [3/7] Verifying inference runtime...
  Downloading Python runtime (~105 MB)...
  ⚠ Python runtime download failed — inference won't work without it
```

`darkbloom doctor` then reports: `Inference runtime not found`.

### 1. Coordinator URLs return 404

```bash
$ curl -sI https://api.darkbloom.dev/dl/eigeninference-python-runtime.tar.gz
HTTP/2 404

$ curl -sI https://api.darkbloom.dev/dl/dginf-python-runtime.tar.gz
HTTP/2 404
```

### 2. PBS fallback logic gap (lines ~210–240)

The PBS fallback only triggers when `$PYTHON_BIN` **exists but doesn't execute**:

```bash
if [ -f "$PYTHON_BIN" ] && ! "$PYTHON_BIN" -c "print('ok')" 2>/dev/null; then
```

When both coordinator downloads fail, `$PYTHON_BIN` **doesn't exist at all**, so this condition is `false` and the PBS download is skipped entirely.

## Fix

Change line ~210 from:

```bash
if [ -f "$PYTHON_BIN" ] && ! "$PYTHON_BIN" -c "print('ok')" 2>/dev/null; then
```

to:

```bash
if ! { [ -f "$PYTHON_BIN" ] && "$PYTHON_BIN" -c "print('ok')" 2>/dev/null; }; then
```

This triggers the PBS fallback whenever Python is missing **or** broken. I verified this fix works — after patching, the script downloads Python 3.12.13 from PBS, installs site-packages from R2, and `darkbloom doctor` reports 8/8 checks passed.

## Additional Context

- The PBS URL itself is fine (GitHub returns 302 → valid download)
- R2 CDN site-packages for v0.3.8 also work
- This affects every fresh install since v0.3.8 was released (2026-04-15)

ethenotethan:
## Summary

Fixes the PBS (python-build-standalone) fallback in `install.sh` so it triggers when Python is **missing**, not just when it's broken. Every fresh install where the coordinator Python runtime download 404s silently skipped the fallback, leaving the provider without a working Python.

Adds an E2E CI job on `macos-latest` that runs the real installer hosted by the coordinator and verifies `darkbloom` works.

Fixes #45

## Details

The old condition on line 217:
```bash
if [ -f "$PYTHON_BIN" ] && ! "$PYTHON_BIN" -c "print('ok')" 2>/dev/null; then
```
requires `$PYTHON_BIN` to **exist** before checking execution. When coordinator downloads fail (HTTP 404), the file is never written, so `-f` is false and the entire PBS fallback block is skipped.

The fix negates the entire compound test:
```bash
if ! { [ -f "$PYTHON_BIN" ] && "$PYTHON_BIN" -c "print('ok')" 2>/dev/null; }; then
```
so the fallback fires whenever Python is missing **or** broken.

## Changes

- **`scripts/install.sh`** and **`coordinator/internal/api/install.sh`**: Fix PBS fallback condition to handle missing Python binary
- **`.github/workflows/ci.yml`**: Add `test-install` job that runs the real installer on `macos-latest` and verifies `darkbloom --version` and `darkbloom doctor`


hankbobtheresearchoor:
## Hermes Agent Review
**Verdict:** LGTM

Two focused changes: (1) an E2E install test in CI, and (2) a shell logic bug fix in the Python runtime check. Both are well-scoped.

### ✅ Looks Good

1. **Shell logic fix is correct**: The original `if [ -f "$PYTHON_BIN" ] && ! "$PYTHON_BIN" -c "print('ok')" 2>/dev/null; then` would only enter the fallback branch if the file exists AND it fails. When the file doesn't exist at all, the short-circuit `&&` evaluation means the `!` test is never reached, and the condition evaluates to false (skipping the fallback). The fix `if ! { [ -f "$PYTHON_BIN" ] && "$PYTHON_BIN" -c "print('ok')" 2>/dev/null; }; then` correctly enters the fallback when the file is missing OR broken. This is a real reliability fix.

2. **Error message improvement**: "Python runtime missing or broken" is more accurate than "Downloaded Python binary doesn't run on this machine" since it covers the missing-file case.

3. **E2E install test**: Running `curl -fsSL https://api.darkbloom.dev/install.sh | bash` + `darkbloom --version` + `darkbloom doctor` on `macos-latest` is a good smoke test for the install flow.

### 💡 Suggestion

1. **E2E install test has no branch/target control**: The install test always hits the production install endpoint (`api.darkbloom.dev`). If the install script or the released binary is broken in production, this CI test won't catch it before merge — it tests the current production, not the PR's changes. Consider whether this test should be a scheduled/nightly job instead of per-PR, or add a way to test the PR's install.sh changes specifically.

2. **`curl | bash` in CI**: While this mirrors real user behavior, it means the CI test depends on the availability of `api.darkbloom.dev`. If the endpoint is down, CI will fail spuriously. Consider adding a retry or timeout annotation.

3. **No artifact from the install test**: The test installs but doesn't verify the installed binary matches the PR's version. Adding a version check (e.g., confirming the installed version matches the expected commit) would make this more robust.

PR #60, opened 2026-04-18
Fix provider model path leakage in completion responses

Gajesh2007:
## Summary
- preserve the public requested model ID before provider requests are rewritten to local backend paths, and stop completion responses from leaking local filesystem paths
- split public model IDs from source repo IDs so the external API uses clean model names while providers still resolve and launch backing artifacts from source IDs
- normalize legacy `mlx-community/...` IDs at coordinator/provider edges for backward compatibility, migrate/dedupe legacy catalog rows, and update app/UI/tooling surfaces to the new public ID contract
- merge latest `origin/master` into this branch and resolve the provider-side conflict by combining master's default coordinator URL constant with the source-aware public/source model ID logic

## Verification
- `go test ./...` in `coordinator/`
- `cargo +nightly fmt --check` in `provider/`
- `PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 cargo +nightly check --target aarch64-apple-darwin --no-default-features` in `provider/`
- `PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 cargo +nightly check --target aarch64-apple-darwin --no-default-features --tests` in `provider/`
- merge conflict reviewed and resolved; repeated parallel review passes after fixing runtime/source-ID and catalog-migration edge cases

<div><a href="https://cursor.com/agents/bc-5c3d0585-1d0c-451f-8c6e-0c4876d18cd0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-5c3d0585-1d0c-451f-8c6e-0c4876d18cd0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/app/EigenInference/Sources/EigenInference/ModelCatalog.swift
+++ b/app/EigenInference/Sources/EigenInference/ModelCatalog.swift
@@ -10,12 +10,17 @@ enum ModelCatalog {
 
     struct Entry: Identifiable {
         let id: String
+        let sourceID: String?
         let name: String
         let modelType: String   // "text", "image", "transcription"
         let sizeGB: Double
         let architecture: String
         let description: String
         let minRAMGB: Int
+
+        var resolvedSourceID: String {
+            sourceID ?? id
+        }
 
         /// Whether this model fits on a machine with the given RAM.
         func fitsInMemory(totalGB: Int) -> Bool {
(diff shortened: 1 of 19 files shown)


hankbobtheresearchoor:
## Hermes Agent Review

This is a critical security and usability PR that decouples public API model IDs from internal source (HuggingFace) IDs, preventing local filesystem path leakage in API responses and errors. The migration logic for renaming legacy model IDs is thorough, and the response sanitization layer is well-implemented with good test coverage.

### 🔴 Critical

1. **Migration deletes legacy rows then re-inserts — no transaction**: In `seedModelCatalog`, the migration loop calls `SetSupportedModel` (upsert to new ID) then `DeleteSupportedModel` (delete old ID). If the process crashes between these two operations, the database could have both old and new rows, or the migration could be re-attempted with partial state. For PostgreSQL, wrap the rename in a single transaction. For MemoryStore, the risk is lower but still exists if the coordinator panics mid-seed.

2. **`canonicalModelID` silently falls back to the input**: The new `canonicalModelID` method returns the input model ID unchanged if no mapping is found. This means if a consumer sends a legacy ID (e.g., `mlx-community/Qwen3.5-122B-A10B-8bit`) after migration, the request may fail with "model not found" because the catalog now uses `Qwen3.5-122B-A10B-8bit`. Consider maintaining a reverse mapping (legacy ID → canonical ID) for a transition period, or at minimum logging when an unmapped ID is used.

### ⚠️ Warning

1. **`sanitize_public_model_error` uses string replacement**: The error sanitization does `error_body.replace(request_model, &public_model)`. If `request_model` is a short string that appears elsewhere in the error message (e.g., a partial path or common word), this could corrupt the error message. Consider using more targeted replacement (e.g., only replacing within JSON `"model"` fields, or using a regex that matches the model value in context).

2. **`rewrite_response_model` rewrites ALL `"model"` keys in the response**: The function replaces any top-level `"model"` string field and any `"response.model"` nested field. If the backend returns a response with a `"model"` field in a different context (e.g., a fine-tuned model reference inside tool call metadata), it would be incorrectly rewritten. Consider scoping the rewrite to known response shapes only.

3. **Source ID exposure in `/v1/models/catalog`**: The handler includes `SourceID` in the public catalog response. If the source ID contains internal repository paths or organizational details, this could leak information. Verify this is intentional.

### 💡 Suggestion

1. **Migration should be idempotent and logged**: Add structured logging for each migrated model ID (legacy → new) so operators can verify the migration in production logs.

2. **`find_catalog_model_by_public_or_source_id` is O(n) per lookup**: In the provider's model listing, this function is called for each downloaded model. With a large catalog, this is O(n*m). A `HashMap<String, &CatalogModel>` indexed by both public and source IDs would be more efficient.

3. **Consider a `"model_aliases"` field in the API**: Rather than silently remapping, expose model aliases so consumers can discover that `mlx-community/Qwen3.5-122B-A10B-8bit` is now `Qwen3.5-122B-A10B-8bit`. This improves the developer experience for migration.

### ✅ Looks Good

1. **Path leakage prevention** — the core security fix is excellent. Leaking `/Users/provider/.cache/huggingface/hub/...` paths in API responses is a real information disclosure vulnerability, and the multi-layer approach (response rewriting, error sanitization, in-process response building) is comprehensive.

2. **Test coverage** — the integration tests for streaming, non-streaming, Responses API, and error responses all verify that leaked paths are scrubbed. Well done.

3. **`sourceID` concept in the Swift app** — clean approach to maintain backward compatibility with on-disk HuggingFace cache paths while presenting clean public IDs.

4. **Console UI model ID updates** — consistent update of example code and pricing tables to use the new canonical IDs.

PR #62, opened 2026-04-18
Optional sender→coordinator request encryption

Gajesh2007:
## Summary

Adds optional sender→coordinator request encryption. Senders NaCl-Box-seal each HTTP request body to the coordinator's long-lived X25519 public key (HKDF-derived from the existing BIP39 mnemonic, published at `GET /v1/encryption-key`). The coordinator decrypts inside the TEE via a `sealedTransport` middleware, runs normal routing/billing on plaintext, and seals the response (buffered JSON or per-event SSE) back to the sender's per-request ephemeral pubkey.

- **Optional and off by default.** Triggered by `Content-Type: application/eigeninference-sealed+json` (case-insensitive, parameter-tolerant). Plaintext SDK / curl flows keep working unchanged.
- **No silent fallback.** A sealed request that fails to decrypt returns 400 — never falls through to plaintext.
- **Threat model.** Protects against everything in front of the coordinator's TEE (CDN, Caddy, proxies, network observers beyond TLS). Coordinator still sees plaintext — true sender→provider E2E (double-sealed inner blob to the provider's SE pubkey) is a follow-up.

### Coordinator (Go)
- `internal/e2e/coordinator_key.go` — HKDF-SHA256 derivation with domain `eigeninference-coordinator-e2e-v1`. `kid` = first 16 hex chars of SHA-256(pubkey) so rotations are detectable.
- `internal/api/sender_encryption.go` — `GET /v1/encryption-key` endpoint plus `sealedTransport` middleware. The sealing `ResponseWriter` handles both buffered JSON and per-SSE-event sealing with status-code preservation.
- Wraps `/v1/chat/completions`, `/v1/responses`, `/v1/completions`, `/v1/messages`, `/v1/images/generations`. Transcribe is intentionally excluded (multipart binary).

### Console-UI (Next.js + tweetnacl)
- `src/lib/encryption.ts` — seal/unseal helpers; coordinator key fetched via `/api/encryption-key` with the user-selected coordinator URL forwarded as `x-coordinator-url`. Cache is keyed per coordinator URL so switching coordinators in Settings transparently picks up the right pubkey.
- `src/lib/api.ts` — `streamChat` seals when toggle on, decrypts each SSE event back, and unseals 4xx/5xx error envelopes so `model_not_found` / `insufficient_funds` stay readable.
- `src/app/settings/page.tsx` — opt-in toggle with eager key fetch on enable.
- `src/app/api/chat/route.ts` — forwards sealed body bytes verbatim and propagates `X-Eigen-Sealed*` headers.
- `src/app/api/encryption-key/route.ts` — proxy that honors `x-coordinator-url`.

### Tests
- Coordinator: key-derivation determinism + cross-mnemonic distinctness + NaCl-Box round trip; integration tests against `httptest.NewServer` for endpoint, happy path, tampered ciphertext, wrong kid, mixed-case `Content-Type`, plaintext bypass, per-event SSE sealing.
- Console-UI: 11 vitest cases including a Go-produced sealed fixture that the TS unseals (cross-compat regression guard).

### Reviewers ran
Both `general-purpose` and `codex-rescue` quality gates ran on this diff. Two High issues from codex are addressed:
1. Coordinator URL was not forwarded to the key-fetch and the cache was global — now per-URL with `x-coordinator-url` threading.
2. Sealed 4xx/5xx error bodies were opaque to the user — `streamChat` now unseals them before parsing the message.

## Test plan

- [ ] `cd coordinator && go test ./...` (the new `internal/e2e` and `internal/api` tests pass; existing suite has no regressions)
- [ ] `cd console-ui && npx vitest run __tests__/encryption.test.ts` (11/11)
- [ ] `cd console-ui && npm run build` (clean; new `/api/encryption-key` route registered)
- [ ] In a deployed environment with a mnemonic configured: hit `GET /v1/encryption-key` and verify the published kid is stable across restarts
- [ ] In console-ui: toggle "Encrypt to coordinator" in Settings, send a chat message, verify the request body on the wire is a sealed envelope (DevTools) and the response decrypts correctly (chat works end-to-end)
- [ ] With encryption enabled, request a model not in the catalog → user sees a readable `model_not_found` (not an opaque sealed envelope)
- [ ] Switch the coordinator URL in Settings while encryption is on → the next request fetches the new coordinator's pubkey instead of using the stale cached one

## Follow-up (separate PR)
- Replay protection: signed timestamp inside the seal + short-window check on the coordinator
- Observability: `sealed_requests_total`, `kid_mismatches_total`, `decryption_failures_total` counters
- Phase 2: double-seal inner blob to provider SE pubkey for true sender→provider E2E (coordinator becomes a blind router)


Gajesh2007:
@claude review this

hankbobtheresearchoor:
## Hermes Agent Review

### Warning

1. **No replay protection on sealed requests**: The PR description acknowledges this as a follow-up, but it is worth highlighting the risk. A sealed request captured from the wire can be replayed indefinitely because there is no timestamp or nonce checked server-side. An attacker who captures one sealed request (even over TLS, via a compromised CDN node) could replay it to drain account balance. Consider adding replay protection before merging, or at minimum documenting the risk and committing to a follow-up timeline.

2. **Coordinator private key in process memory** (`coordinator_key.go`): The X25519 private key lives in a Go struct for the lifetime of the process. Unlike the provider SE keys, this key is not protected by a Secure Enclave or hardware boundary. A memory dump or core capture exposes the key, compromising all past and future sealed requests. Consider deriving the key on demand and zeroing after use, or using a KMS/TEE sealing mechanism.

3. **Sealed transport silently falls through when coordinatorKey is nil** (`sender_encryption.go`): If the coordinator key is not configured, a sealed request hits the middleware, finds coordinatorKey == nil, and returns 503. But there is no metric or log distinguishing "sealed request rejected because encryption not configured" from other 503s. This could confuse operators debugging why a client with encryption enabled is failing.

### Suggestion

4. **Cross-compat fixture uses hardcoded test mnemonic**: The encryption tests include a hardcoded BIP39 mnemonic. While this is test-only, any developer who accidentally uses this mnemonic for a real deployment would have their coordinator encryption key known to anyone who reads the test file. Consider using a clearly fake/invalid mnemonic or adding a prominent warning comment.

5. **SSE sealing buffers until double-newline** -- edge case with very large single events: The sealedResponseWriter buffers SSE data until it sees a double newline. If a provider sends a single SSE event with multi-megabyte content (e.g., a very long completion), the entire event is buffered in memory before sealing. Consider adding a size limit on the buffer to prevent OOM.

6. **/api/encryption-key proxy caches for 5 minutes** (`route.ts`): The proxy sets Cache-Control: public, max-age=300. If the coordinator rotates its key (e.g., after a mnemonic change), clients will use the stale key for up to 5 minutes, causing decryption failures. The coordinator-side kid mismatch handling returns 400, but the client experience would be confusing errors for up to 5 minutes. Consider a shorter TTL or adding a cache-bust mechanism on kid change.

7. **Missing x-coordinator-url forwarding on /api/encryption-key**: The proxy reads coordUrl from the header but does not forward it in the upstream fetch call. The coordinator key is per-coordinator, so the wrong key would be cached if the user switches coordinators. The client-side getCoordinatorKey in encryption.ts does forward it, but the server-side proxy at api/encryption-key/route.ts does not. This is inconsistent.

### Looks Good

8. **No silent fallback to plaintext**: A sealed request that fails decryption returns 400, never falls through to the plaintext handler. This is the correct security posture -- fallback would undermine the entire encryption feature.

9. **HKDF domain separation**: Using eigeninference-coordinator-e2e-v1 as HKDF info ensures the encryption key is cryptographically independent from the Solana billing keypair derived from the same mnemonic. Well-designed.

10. **Per-SSE-event sealing**: Sealing each SSE event individually (rather than buffering the entire stream) is the right approach for streaming inference -- it maintains low latency and allows the client to decrypt incrementally.

11. **Per-URL key caching in the console-UI**: Switching coordinator URLs correctly invalidates the cache and fetches a new key. This prevents cross-coordinator key confusion.

PR #63, opened 2026-04-18
Routing: observability + 4-phase scheduler improvements + reliability hardening

Gajesh2007:
## Summary

Three layered changes to how the coordinator matches inference requests to providers:

1. **Reliability hardening** — Prometheus metrics package, per-account rate limiter, panic-safe goroutines, registry clamps on provider-reported stats, status-canonical SE signing.
2. **Routing observability** — per-decision metrics (`RoutingDecisions`, `ProviderSelected`, `RoutingCostMs`, `EffectiveDecodeTPS`, `AttestationChallenges`), cost-breakdown debug log, structured `RoutingDecision` returned by `ReserveProviderEx`.
3. **4-phase scheduler algorithm changes**, each gated by tests that fail without the change:
   - **P1** — free-memory admission gate (reject provider when `freeGB < SizeGB + kvEstimate`)
   - **P2** — tighter `maxConcurrency` tiers (4/8/16/24/32 → 2/4/6/8/12) to match real MLX batch degradation
   - **P3** — stronger queue penalty + wider tie window so one queued request ≈ one slow-provider decode
   - **P4** — load-scaled effective decode TPS: `effective = static / (1 + k·batch)`

Both routing constants (`effectiveTPSLoadFactor`, `kvCacheBytesPerToken`) were **measured empirically** against a live `vllm-mlx` backend on M4 Max + Qwen2.5-7B-4bit via `scripts/calibrate-routing.sh`. Both prior guesses were materially wrong (47% and 48% off respectively).

> Note: an earlier draft of this PR included a "Phase 5: model-swap penalty" based on a wrong mental model of the provider runtime. The provider runs **one `vllm-mlx` process per configured model on its own port** (`provider/src/main.rs:2371`); models coexist, no swap ever happens. Phase 5 has been reverted (`40d760d4`).

## Why

Production symptom: all traffic funnels to the biggest machine. A single M3 Ultra was taking ~100 % of requests while M3 Pros sat idle, because the cost function's decode-TPS gap dwarfed the queue penalty. Providers could also OOM when advertising models that didn't fit — no free-memory check. And a heavily-batched Ultra looked infinitely faster than an idle Pro, because the scheduler used single-request benchmarked TPS regardless of actual batch size.

## What's in this PR

15 commits across three logical groupings:

**Reliability foundation (6 commits):**
- `b9abf010` metrics package, ratelimit, saferun, clamps, status-canonical signing
- `5b33e316` routing observability metrics
- `ae011ea1` queue-dispatch metric propagation (reviewer fix)
- `dd7d4f14` `over_capacity` vs `queued` outcome distinction (reviewer fix)
- `d37ed649` HTTP-level metric assertion tests + clarified status-field enforcement comment

**Algorithm scenario tests (1 commit):**
- `45d70173` algorithm scenarios: phase goals + regression guards, red→green TDD signal across phases

**4 phases + reverted P5 + reviewer fixes + calibration (8 commits):**
- `afb47135` P1: `CatalogEntry.SizeGB` + `freeMemoryAdmits` gate in `buildCandidate`
- `c7dfeab1` P2: lower `maxConcurrency` tiers
- `b79a4795` P3: queue penalty 1000→3000, window 750→3000, pending 250→750
- `0daa4d53` P4: `effectiveDecodeTPS` helper, used in decode + backlog terms
- `bee82414` P5: model-swap penalty (later reverted — wrong premise)
- `f65a75b8` tie-break order-dependence fix (two-pass selection), gofmt, KV int64-overflow guard
- `4333c50a` `scripts/calibrate-routing.sh`
- `e8d921df` calibrated constants from M4 Max measurement
- `40d760d4` revert P5 — providers run one vllm-mlx per model, no swap exists

## Measured vs coded constants

| Constant | Coded | Measured (M4 Max, Qwen 7B 4-bit) | Shipped |
|---|---|---|---|
| `kvCacheBytesPerToken` | 524,288 (0.5 MB) | 357,615 (0.34 MB) | 400,000 (0.38 MB, slack for larger models) |
| `effectiveTPSLoadFactor` | 0.40 | 0.27 | 0.27 |

## Test plan

- [x] `go test ./...` green across all coordinator packages
- [x] `go test ./... -race -timeout 180s` green × 3 stress runs
- [x] 12/12 algorithm scenarios green (`TestAlgorithm_*` in `registry/algorithm_scenarios_test.go`)
- [x] 5 HTTP-level metric wiring tests (`api/metrics_integration_test.go`)
- [x] Calibration script run against live vllm-mlx, confirms constants
- [x] Two reviewer passes (Claude Code subagent + Codex rescue): caught tie-break order-dependence (fixed), KV-overflow concern (defensive cap added), and the wrong-premise Phase 5 (reverted)
- [ ] Canary against one test-rack provider pair before fleet rollout
- [ ] Watch `ProviderSelected` distribution in prod for 24 h post-merge — distribution skew >80 % on any single provider means the constants need another pass
- [ ] Re-run `scripts/calibrate-routing.sh` on Ultra and Pro tiers; `effectiveTPSLoadFactor` may need per-tier tuning (only M4 Max measured so far)

## Carried-over findings not fixed here

- `provider.go:510-609` enforces `SIPEnabled`/`BinaryHash`/etc. even when `statusFieldsTrusted == false`. Intentional defense-in-depth during the v0.3.11 rollout window; comment clarified in `d37ed649`. TODO already in code to require status sig in v0.3.13+.
- Per-chip-tier `effectiveTPSLoadFactor` lookup: if k varies meaningfully by chip, a map beats one global constant. Deferred until Ultra/Pro data exists.

## Breaking changes

None. All added fields on `RoutingDecision` are additive. Existing `ReserveProvider` signature preserved as a thin wrapper around `ReserveProviderEx`. No protocol changes between coordinator and provider.


hankbobtheresearchoor:
## Hermes Agent Review
**Verdict:** LGTM

### ⚠️ Warning

1. **MemoryStore pruner is a silent data loss vector** (`main.go`): The pruner runs every 15 minutes and caps each slice to `DefaultPruneMaxEntries`. For usage/ledger/earnings, this means historical billing records are silently discarded. While the PR correctly gates this behind `EIGENINFERENCE_ALLOW_MEMORY_STORE`, the pruner gives a false sense of safety — an operator running in-memory "just for testing" could lose usage records they intended to reconcile later. Consider logging the number of pruned entries per category at WARN level, or making the pruner opt-in separately from the memory store itself.

2. **`envFloat`/`envInt` silently ignore malformed values** (`main.go`): If someone sets `EIGENINFERENCE_RATE_LIMIT_RPS=abc`, the helper silently falls back to the default with no log warning. In a security-sensitive configuration (rate limiting), a typo could silently disable or misconfigure protection. Consider logging a warning when a non-empty env var fails to parse.

3. **Rate limiter pruner uses `saferun.Recover` in a deferred closure** (`main.go`): The `StartPruner` call passes `func() { saferun.Recover(logger, "ratelimit_pruner") }` as the panic callback. This means a panic in the pruner is recovered but not re-raised or re-scheduled — the pruner silently stops. Over time, the rate limiter map grows unboundedly if the pruner is dead. Consider logging a CRITICAL-level message if the pruner panics, or re-starting it.

### 💡 Suggestion

4. **Financial rate limit defaults are very strict**: 0.2 RPS with burst 3 means an account can do about 1 financial operation every 5 seconds. For an operator managing many providers (deposits, withdrawals, key rotations), this may be too restrictive. The env var override exists, but the default should perhaps be documented with guidance on scaling.

5. **Metrics fleet poller interval is hardcoded**: The 15-second poll interval for `ProvidersConnected`/`ProvidersIdle`/`QueueDepth` is not configurable. For large fleets (1000+ providers), 15-second polling may be too frequent and create lock contention on the registry snapshot. Consider making the interval configurable or switching to event-driven metric updates.

6. **Cancel write timeout is 2 seconds but no metric tracks failures**: `sendProviderCancel` logs at debug level when the cancel write fails, but there is no Prometheus counter. Failed cancels mean providers continue processing orphaned requests — a counter would help detect systemic WebSocket health issues.

### ✅ Looks Good

7. **MemoryStore guard is excellent**: Refusing to boot without `EIGENINFERENCE_DATABASE_URL` unless explicitly opted in prevents the catastrophic scenario of running production billing on a non-durable store. This is a critical safety net.

8. **Cancel message ordering fix**: The comment at `consumer.go:490-498` explaining why `RemovePending` must precede `sendProviderCancel` is exactly right — this prevents a double-charge race where the original provider sends `InferenceComplete` after the retry has already been billed.

9. **Prometheus metrics are well-structured**: `RoutingDecisions` with outcome labels (`selected`, `queued`, `over_capacity`, `no_provider`) gives exactly the observability needed to diagnose routing skew. `EffectiveDecodeTPS` per provider enables per-machine performance dashboards.

10. **saferun package**: Centralized goroutine lifecycle management with panic recovery and a configurable observer is the right pattern for a long-running coordinator process.

PR #64, opened 2026-04-18
Add /providers/me dashboard for account-scoped fleet view

Gajesh2007:
## Summary

Adds a "My Machines" dashboard at `/providers/me` so logged-in users can see every provider machine linked to their account — status, earnings, and the full list of reasons a machine might not be earning.

## Coordinator

- `Store.ListProvidersByAccount(ctx, accountID)` — memory + postgres impls + partial `idx_providers_account` index
- `GET /v1/me/providers` (Privy auth) — merges persisted `ProviderRecord` with the live `registry.Provider` snapshot, dedupes per-session reconnect rows by `SerialNumber` / `SEPublicKey` so one physical machine surfaces once, attaches per-machine lifetime earnings
- `GET /v1/me/summary` (Privy auth) — lifetime / 7d / 24h earnings, pending balance, fleet counts (online / serving / offline / untrusted / hardware / needs-attention), latest + min provider version
- `httptest`-based handler tests for auth, account isolation, offline/live merge, reconnect dedupe, earnings attach, summary buckets
- Postgres store test for `ListProvidersByAccount` gated on `DATABASE_URL`

## Console-UI

- Proxy routes `/api/me/{providers,summary}` forwarding the Privy bearer
- `/providers/me` page: summary header, payout banner, per-machine card with status, hardware, live metrics, warm models, reputation, and expandable Trust & attestation / Backend slots sections
- Pure `computeWarnings()` derives 17 signals across **blocking / degrading / info** severities mirroring `ScoreProvider` + `FindProviderWithTrust` exclusion rules:
  - **Blocking**: offline, untrusted, runtime mismatch, version below min, stale attestation challenge, trust below routing threshold (self-signed / none), no catalog models, thermal critical
  - **Degrading**: crashed backend (0.05×), idle-shutdown cold start (0.1×), MDA missing, thermal serious/fair, memory pressure >90%, low success rate
  - **Info**: no payout configured, outdated provider version, awaiting first challenge
- 29 vitest assertions covering the warning matrix + semver helper
- Existing `/providers` page's wallet-match card replaced with a CTA linking to the new dashboard

## Test plan

- [x] `cd coordinator && go test ./internal/store/ ./internal/api/ -count=1` passes
- [x] `cd console-ui && npx vitest run providers-me-warnings` → 29/29
- [x] `cd console-ui && npm run build` — `/providers/me`, `/api/me/providers`, `/api/me/summary` register cleanly
- [x] `gofmt -l` clean on all touched Go files
- [x] TypeScript clean for new files (pre-existing errors in `EarningsContent.tsx` / `PrivyClientProvider.tsx` / `vitest.config.ts` are not from this PR)
- [ ] Manual: log in, confirm the dashboard renders with real fleet data and warnings fire for a machine that's offline / self-signed / outdated
- [ ] Manual: sanity-check `/v1/me/providers` and `/v1/me/summary` against a deployed coordinator with real account linkage

## Review

Two independent reviewers (Codex rescue + general-purpose Claude subagent) caught and have been addressed:
- Reconnect duplication (persisted records from old sessions) → dedup by stable identity
- Trust level warnings had wrong severity (production `MinTrustLevel=hardware` excludes sub-hardware entirely)
- `backend_crashed` was overstated as blocking → actually a 0.05× scoring penalty
- Missing stale-challenge warning (the 6-minute freshness window)
- Missing no-catalog-models warning
- TypeScript regression on `challenge_max_age_seconds` in the `Pick<>` shape

### DIFF
--- a/console-ui/__tests__/providers-me-warnings.test.ts
+++ b/console-ui/__tests__/providers-me-warnings.test.ts
@@ -1,0 +1,345 @@
+import { describe, it, expect } from "vitest";
+import {
+  computeWarnings,
+  highestSeverity,
+  semverLess,
+} from "@/app/providers/me/warnings";
+import type { MyProvider, MyProvidersResponse } from "@/app/providers/me/types";
+
+const ctx: Pick<
+  MyProvidersResponse,
+  "latest_provider_version" | "min_provider_version" | "heartbeat_timeout_seconds" | "challenge_max_age_seconds"
+> = {
+  latest_provider_version: "0.3.10",
+  min_provider_version: "0.3.10",
+  heartbeat_timeout_seconds: 90,
+  challenge_max_age_seconds: 360,
+};
+
+function baseProvider(overrides: Partial<MyProvider> = {}): MyProvider {
+  return {
+    id: "p1",
+    account_id: "acct-1",
+    status: "online",
+    online: true,
+    hardware: { chip_name: "Apple M3 Max", memory_gb: 64 },
+    models: [{ id: "mlx-community/Qwen3.5-9B-MLX-4bit" }],
+    trust_level: "hardware",
+    attested: true,
+    mda_verified: true,
+    acme_verified: true,
+    se_key_bound: true,
+    secure_enclave: true,
+    sip_enabled: true,
+    secure_boot_enabled: true,
+    authenticated_root_enabled: true,
+    runtime_verified: true,
+    failed_challenges: 0,
+    pending_requests: 0,
+    max_concurrency: 8,
+    reputation: {
+      score: 0.85,
+      total_jobs: 0,
+      successful_jobs: 0,
+      failed_jobs: 0,
+      total_uptime_seconds: 0,
+      avg_response_time_ms: 0,
+      challenges_passed: 5,
+      challenges_failed: 0,
+    },
+    lifetime_requests_served: 0,
+    lifetime_tokens_generated: 0,
+    earnings_total_micro_usd: 0,
+    earnings_count: 0,
+    last_challenge_verified: new Date().toISOString(),
+    version: "0.3.10",
+    ...overrides,
+  };
+}
+
+describe("semverLess", () => {
+  it("compares numeric segments", () => {
+    expect(semverLess("0.3.9", "0.3.10")).toBe(true);
+    expect(semverLess("0.3.10", "0.3.9")).toBe(false);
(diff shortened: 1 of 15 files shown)


hankbobtheresearchoor:
## Hermes Agent Review
**Verdict:** LGTM

### 💡 Suggestion

1. **Proxy routes pass auth tokens to upstream without validation** (`api/me/providers/route.ts`, `api/me/summary/route.ts`): The proxy reads the `privy-token` cookie and forwards it as a Bearer token. This is fine for the current architecture, but if the coordinator ever adds token revocation or rotation, the cached browser token could be stale. Consider adding a short `max-age` or `stale-while-revalidate` cache header on the proxy response so the UI does not show permanently stale data between refreshes.

2. **Warning severity does not influence provider card ordering**: `highestSeverity()` is computed but the `MyMachinesContent` component sorts by `status` then by `id`. Providers with blocking warnings might appear below healthy ones. Consider sorting by severity (blocking first) within each status group so the user sees the most actionable issues at the top.

3. **`semverLess` does not handle pre-release tags**: The semver comparison handles numeric segments but would treat `"0.3.10-beta"` as equal to `"0.3.10"` (string comparison on the third segment). If provider versions ever include pre-release suffixes, the comparison could misclassify. Low risk for now since the codebase appears to use clean semver.

4. **`computeWarnings` is pure but not memoized**: Called on every render for every provider. With many machines (50+), this recomputes 17 checks per machine per render cycle. The function is lightweight, but wrapping in `useMemo` keyed by the provider data would be more idiomatic.

### ✅ Looks Good

5. **Warning matrix is comprehensive and well-aligned with coordinator routing**: The 17 warning signals mirror the actual `ScoreProvider` / `FindProviderWithTrust` exclusion rules. This means the UI tells users exactly why their machine is not earning, which is excellent UX for a distributed inference network.

6. **Reconnect deduplication by SerialNumber/SEPublicKey**: Correctly handles the case where a provider reconnects and creates duplicate persisted records — deduping by stable identity prevents showing the same physical machine twice.

7. **Test coverage**: 29 vitest assertions covering the full warning matrix, semver helper, severity ranking, and edge cases (small sample size, offline machine, double-warn suppression) is thorough.

8. **Security: auth-forwarding proxy pattern**: The BFF route correctly validates auth before proxying to the coordinator, keeping the Privy token server-side. The `x-coordinator-url` header forwarding allows multi-coordinator support without leaking tokens to the wrong backend.

PR #65, opened 2026-04-18
Allow free self-hosted provider use while model is running

Gajesh2007:
## Summary
- add owner-pinned self-hosted routing for text requests when a linked provider already has the requested model actively running and is E2E-ready
- keep self-hosted requests free by recording zero consumer cost, zero provider payout, and zero platform fee while still tracking usage
- preserve normal marketplace behavior by falling back to the paid path when self-hosted routing is unavailable or fails before commit, and add focused registry/api regression coverage

## Testing
- `cd coordinator && go test ./internal/registry ./internal/api`

<div><a href="https://cursor.com/agents/bc-c9d71705-34c3-4d15-ac56-aaeee0bcfc12"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c9d71705-34c3-4d15-ac56-aaeee0bcfc12"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/coordinator/internal/api/billing_integration_test.go
+++ b/coordinator/internal/api/billing_integration_test.go
@@ -11,6 +11,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"io"
 	"log/slog"
 	"net/http"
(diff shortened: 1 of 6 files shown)


hankbobtheresearchoor:
## Hermes Agent Review

### ⚠️ Warning

1. **Self-hosted billing bypass creates abuse surface** (`consumer.go` `dispatchSelfHostedTextRequest`): The self-hosted path sets `pr.SelfHosted = true`, which causes `handleComplete` to skip all billing — consumer cost, provider payout, and platform fee. While the intent is clear (owner using own machine), a compromised account could link many cheap providers and route unlimited free traffic. Consider adding a per-account self-hosted request rate limit or quota, separate from the marketplace rate limiter.

2. **120-second polling loop with no backoff** (`dispatchSelfHostedTextRequest`): The self-hosted wait path polls every 200ms for up to 120s. This is 600 iterations of acquiring the registry mutex. Under high self-hosted load, this creates contention. Consider exponential backoff or a condition-variable / channel-based signal instead of polling.

3. **Self-hosted fallback race condition**: If `ReserveOwnedProvider` returns nil but `HasOwnedRunningProvider` returns true (the owned provider becomes busy between the two calls), the code enters the polling loop. But if the provider crashes or goes offline during polling, the loop will spin for the full 120s before returning an error. There is no re-check of `HasOwnedRunningProvider` inside the loop to detect permanent loss of the owned provider.

### 💡 Suggestion

4. **Double-write risk on retry**: In `handleChatCompletions`, the self-hosted dispatch path is attempted first, then falls back to marketplace. If the self-hosted dispatch succeeds (provider reserved) but the WebSocket send fails and the code retries, the retry goes through the full dispatch again including self-hosted. The `providerDispatch.selfHosted` field is set but the retry loop in the outer function does not check it — a self-hosted dispatch could be retried as a marketplace dispatch on the second attempt. Consider propagating the self-hosted flag into the retry loop.

5. **No observability for self-hosted vs marketplace split**: The `handleComplete` log now includes `"self_hosted"` which is good, but there are no metrics counters distinguishing self-hosted from marketplace requests. Add `metrics.RoutingDecisions.WithLabelValues(model, "self_hosted")` and `metrics.RoutingDecisions.WithLabelValues(model, "marketplace")` so the traffic split is visible in dashboards.

6. **Integration test `time.Sleep` fragility**: The billing integration tests rely on `time.Sleep(300ms)` for state propagation. This is inherently flaky under load or CI slowdown. Consider using a synchronization primitive (channel, condition variable) or polling with assertion instead.

### ✅ Looks Good

7. **Tight self-hosted eligibility**: `providerCanSelfHostLocked` correctly requires `slot.State == "running"` and matching `AccountID` — this prevents self-hosted traffic from hitting cold or crashed backends, which is the right security boundary.

8. **Zero-cost accounting**: Recording usage with `CostMicroUSD=0` while still tracking prompt/completion tokens is the right approach — usage visibility without billing leakage.

9. **Scheduler refactoring**: Extracting `pickBestCandidate` from `selectBestCandidateLocked` is clean and enables reuse for both marketplace and owned-provider selection.

PR #71, opened 2026-04-19
Disaggregated compute: put small Macs to work serving embeddings + rerank

Gajesh2007:
## Summary

Decoder LLMs are memory-bandwidth bound — to serve them you need a model that fits in RAM. That's why a 16 GB Air or base Mini was previously useless to the network. But **embeddings and reranking** are compute-bound, the models are 100 MB – 2 GB (fit comfortably in 8 GB), and the responses are tiny. Every retrieval/RAG/search system needs them. Tons of demand, perfect match for low-RAM Apple Silicon.

This PR ships the disaggregated-compute layer: small Macs earn revenue serving compute-bound jobs while big-RAM Macs stay free for memory-bandwidth-bound LLM decode. Two new OpenAI-/Cohere-shaped endpoints, tier-aware routing, end-to-end encrypted, with billing and reputation flowing through the existing ledger.

## Linked issue

Closes #

## What ships

**Provider tier classification** (`coordinator/internal/registry/registry.go`)
- `tiny` (≤16 GB) — embeddings, rerank, future spec-decode draft
- `small` (18–24 GB) — embeddings + small chat
- `standard` (≥32 GB) — full text/image/STT decode
- Tier is derived from advertised hardware on `Register` so providers cannot misrepresent capability

**Tier-aware routing** (`coordinator/internal/registry/scheduler.go`)
- Catalog entries gain `ModelType`; embedding/rerank requests apply a 50 000 ms tier-mismatch penalty so the smallest capable provider wins, but work overflows to bigger machines if the small fleet is busy

**New endpoints** (`coordinator/internal/api/embeddings.go`)
- `POST /v1/embeddings` — OpenAI shape
- `POST /v1/rerank` — Cohere shape (the de-facto rerank API used by LangChain / LlamaIndex / Haystack)
- E2E encryption mandatory (NaCl box, X25519 + XSalsa20-Poly1305), same trust boundary as chat
- Pre-flight worst-case reservation, refund unused, clamp at reservation so a misbehaving provider can't overcharge
- **Cross-provider retry** (3 attempts, excludes failed provider) so a single flaky small-tier provider doesn't 502 the consumer
- Provider may seal the response vectors back over the session key (symmetric with chat completions)

**Wire protocol** (`coordinator/internal/protocol/messages.go` ↔ `provider/src/protocol.rs`)
- 4 new message types: `embedding_request`, `embedding_complete`, `rerank_request`, `rerank_complete`
- Vectors travel inline (typical payload ≤ 256 KB, well under 10 MB WS frame limit)

**Provider proxy** (`provider/src/proxy.rs`, `coordinator.rs`, `main.rs`)
- New `handle_embedding_request` / `handle_rerank_request` that forward to a local HTTP backend on `127.0.0.1:$EIGENINFERENCE_EMBEDDING_PORT` (default `image_port + 1`)
- Same E2E session-key encryption flow as chat
- 6 new catalog entries (bge-m3, mxbai-embed-large, Qwen3-Embedding-{0.6B,4B}, bge-reranker-v2-m3, Qwen3-Reranker-0.6B)

**Pricing** (`coordinator/internal/payments/pricing.go`)
- Embeddings: $0.005 / 1 M tokens (half of Together $0.008)
- Reranking: $0.010 / 1 M tokens
- Same 95 % provider / 5 % platform split as chat

**Catalog seeding** (`coordinator/cmd/coordinator/main.go`)
- Adds the 6 embedding/rerank models to `seedModelCatalog`

**Docs** (`docs/disaggregated-compute.md`) — full spec of the system, including the future spec-decode draft offload direction.

## Review-round-2 fixes

Independent reviewers caught real bugs in the first revision; commit `e173a826` fixes all of them:

| Bug | Severity | Fix |
|-----|----------|-----|
| Refund minted free credit when billing was disabled | **CRITICAL** | Gate both reservation + refund on `s.billing != nil` (mirrors chat) |
| `RemovePending` skipped on encryption-error early returns → stuck `Serving` provider | **MEDIUM** | Single `cleanup()` closure in new dispatch helpers |
| No cross-provider retry → single flaky provider fails consumer | **MEDIUM** | `embeddingMaxAttempts=3`, excludes failed providers (mirrors chat) |
| `RecordJobSuccess` called with token count instead of duration → reputation always 1.0 | MEDIUM | Use `result.DurationSecs * Second` with 100 ms floor |
| `encoding_format=base64` silently ignored | NIT | Now returns 400 with clear error |
| `AllowedTiersForModelType` was dead code (never wired into scheduler) | NIT | Removed |

Each fix has a regression test:
- `TestEmbeddingsNoFreeCreditWhenBillingDisabled` — issues an embedding request with no providers and asserts the consumer's ledger balance is unchanged and no `LedgerRefund` entries are created.
- `TestEmbeddingsRetriesAcrossProviders` — registers two providers that both fail, asserts both were attempted before the consumer-visible error.
- `TestEmbeddingsRejectsBase64EncodingFormat` — asserts 400 instead of silent shape drift.

## Test plan

- [x] `cd coordinator && go test ./...` — all green (incl. 6 new e2e tests + 3 regression tests + 4 new tier tests)
- [x] `cd coordinator && gofmt -l .` — clean
- [x] `cd coordinator && go vet ./...` — clean
- [x] Standalone Rust roundtrip test of the new types vs Go-emitted JSON — 7/7 passing
- [x] `cargo fmt --check` on the files I touched — clean

## Components touched

- [x] coordinator (Go)
- [x] provider (Rust)
- [ ] console-ui (Next.js)
- [ ] image-bridge (Python)
- [ ] app (macOS Swift)
- [ ] enclave (Swift)
- [ ] infra / CI / release
- [x] docs

- [ ] No protocol/interface changes
- [x] Yes — described above and matching side updated

Both sides updated symmetrically (`coordinator/internal/protocol/messages.go` ↔ `provider/src/protocol.rs`); roundtrip tests on both sides catch drift.

No bundle script changes — the embedding sidecar runs at a deterministic port (`image_port + 1` or `EIGENINFERENCE_EMBEDDING_PORT`) so the launcher and the proxy agree without extra IPC. A first-party sidecar based on `mlx_embeddings` will ship with the next provider bundle; until then the proxy will return `503` (connect refused) and the coordinator will route to another small-tier provider via the new retry loop, keeping the failure local.

## Notes for reviewers

**First-principles design:**

1. Decode is memory-bandwidth bound; small Macs can't host quality LLMs. Trying to disaggregate decode itself across the WAN is a non-starter (KV-cache transfer saturates internet bandwidth).
2. Embeddings/rerank are compute-bound, small models, small payloads — they map perfectly onto small Apple Silicon and there is huge market demand.
3. So the smallest meaningful intervention is: classify providers by tier, add tier-preferring routing, ship the two endpoints. That's exactly what this PR does.

**Future work (not in this PR):**

The same protocol scaffolding (small request, small response, compute bound) is what speculative-decoding draft offload needs — a tiny-tier provider runs the draft model, a standard-tier provider verifies tokens in parallel. Quality is mathematically identical (rejection sampling), big provider's GPU does the parallelizable work. That's left as a follow-up because it requires changes inside `vllm-mlx` (in `.external/`).

**E2E encryption is mandatory** for both endpoints — even short query text can be sensitive (medical, legal, financial). Same X25519 + XSalsa20-Poly1305 (NaCl box) as chat, same `e2e` package, same provider keypair.

<div><a href="https://cursor.com/agents/bc-4a6ac94b-fdef-4dfd-819a-edee318440ff"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-4a6ac94b-fdef-4dfd-819a-edee318440ff"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/coordinator/cmd/coordinator/main.go
+++ b/coordinator/cmd/coordinator/main.go
@@ -421,6 +421,18 @@ func seedModelCatalog(st store.Store, logger *slog.Logger) {
 		{ID: "mlx-community/gemma-4-26b-a4b-it-8bit", S3Name: "gemma-4-26b-a4b-it-8bit", DisplayName: "Gemma 4 26B", ModelType: "text", SizeGB: 28.0, Architecture: "26B MoE, 4B active", Description: "Fast multimodal MoE", MinRAMGB: 36, Active: true},
 		{ID: "mlx-community/Qwen3.5-122B-A10B-8bit", S3Name: "Qwen3.5-122B-A10B-8bit", DisplayName: "Qwen3.5 122B", ModelType: "text", SizeGB: 122.0, Architecture: "122B MoE, 10B active", Description: "Best quality", MinRAMGB: 128, Active: true},
 		{ID: "mlx-community/MiniMax-M2.5-8bit", S3Name: "MiniMax-M2.5-8bit", DisplayName: "MiniMax M2.5", ModelType: "text", SizeGB: 243.0, Architecture: "239B MoE, 11B active", Description: "SOTA coding, 100 tok/s", MinRAMGB: 256, Active: true},
+
+		// --- Disaggregated compute: embeddings + reranking on small Macs ---
+		// These models give low-RAM Mac providers (Air, base Mini) revenue.
+		// They are routed preferentially to tiny/small tier providers so big
+		// Macs stay free for memory-bandwidth-bound LLM decode. See
+		// coordinator/internal/registry/registry.go: PreferredTiersForModelType.
+		{ID: "mlx-community/bge-m3", S3Name: "bge-m3", DisplayName: "BGE-M3 Embeddings", ModelType: "embedding", SizeGB: 1.2, Architecture: "568M multilingual encoder", Description: "Multilingual dense + sparse embeddings", MinRAMGB: 8, Active: true},
+		{ID: "mlx-community/Qwen3-Embedding-0.6B", S3Name: "Qwen3-Embedding-0.6B", DisplayName: "Qwen3 0.6B Embeddings", ModelType: "embedding", SizeGB: 0.7, Architecture: "0.6B Qwen3 encoder", Description: "Tiny multilingual embeddings", MinRAMGB: 8, Active: true},
+		{ID: "mlx-community/Qwen3-Embedding-4B", S3Name: "Qwen3-Embedding-4B", DisplayName: "Qwen3 4B Embeddings", ModelType: "embedding", SizeGB: 4.5, Architecture: "4B Qwen3 encoder", Description: "High-quality multilingual embeddings", MinRAMGB: 16, Active: true},
+		{ID: "mlx-community/mxbai-embed-large-v1", S3Name: "mxbai-embed-large-v1", DisplayName: "mxbai Large Embeddings", ModelType: "embedding", SizeGB: 0.7, Architecture: "335M BERT-large", Description: "English embeddings, top of MTEB", MinRAMGB: 8, Active: true},
+		{ID: "mlx-community/bge-reranker-v2-m3", S3Name: "bge-reranker-v2-m3", DisplayName: "BGE Reranker v2-m3", ModelType: "rerank", SizeGB: 1.2, Architecture: "568M cross-encoder", Description: "Multilingual cross-encoder reranker", MinRAMGB: 8, Active: true},
+		{ID: "mlx-community/Qwen3-Reranker-0.6B", S3Name: "Qwen3-Reranker-0.6B", DisplayName: "Qwen3 0.6B Reranker", ModelType: "rerank", SizeGB: 0.7, Architecture: "0.6B cross-encoder", Description: "Tiny multilingual reranker", MinRAMGB: 8, Active: true},
 	}
 
 	added := 0
--- a/coordinator/internal/api/embeddings.go
+++ b/coordinator/internal/api/embeddings.go
@@ -1,0 +1,703 @@
+package api
+
+// Disaggregated compute endpoints — embeddings and reranking.
+//
+// These endpoints exist so low-RAM Mac providers (Air, 16 GB Mini) — which
+// cannot host a quality decoder LLM in memory — still earn revenue and add
+// useful capacity to the network. Embeddings and rerankers run on small
+// (100 MB – 2 GB) models that comfortably fit in 8–16 GB and are bottlenecked
+// by GPU compute, not memory bandwidth, so a small Mac is genuinely fast at
+// them. Routing prefers small/tiny tier providers (see
+// registry.PreferredTiersForModelType) so big-RAM machines stay free for
+// memory-bandwidth-bound decode of large language models.
+//
+// API surface mirrors the OpenAI /v1/embeddings shape; rerank mirrors
+// Cohere's /v1/rerank, which is the de-facto standard supported by
+// LangChain, LlamaIndex, Haystack, and most retrieval stacks.
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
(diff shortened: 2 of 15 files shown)


hankbobtheresearchoor:
## Hermes Agent Review

This PR adds disaggregated compute endpoints (embeddings + reranking) for low-RAM providers — a thoughtful architectural choice to monetize small Mac hardware. The implementation mirrors the chat completion path closely (E2E encryption, billing reservation/refund, retry logic, provider routing), which is good for consistency.

### ⚠️ Warning

1. **Billing over-charge clamping logs but does not penalize**: When `actualCost > reservedMicroUSD`, the code clamps and logs an error but does not call `RecordJobFailure` or take any reputation action against the provider. A consistently over-reporting provider will keep getting routed to. Consider recording a job failure or incrementing a reputation counter for over-reporting providers.

2. **`estimateEmbeddingPromptTokens` falls back to `len(input) / 4`**: When the input is neither a string, array of strings, nor array of token arrays, the fallback `len(input) / 4` operates on raw JSON bytes, which includes quotes, brackets, and commas. This could significantly over-estimate tokens for small inputs, leading to higher-than-necessary reservations.

3. **No input size limit on embedding/rerank requests**: The code does not limit the number of documents in a rerank request or the size of the input array in an embedding request. A request with thousands of documents could OOM a small-tier provider or exceed the 60s timeout. Consider adding a maximum batch size.

4. **`reqwest::Client::builder().build().unwrap_or_default()` per request**: Both `do_embedding` and `do_rerank` create a new HTTP client per request. This creates a new connection pool and TLS state each time. Consider sharing a client (e.g., via `Arc<reqwest::Client>`) for connection reuse.

### 💡 Suggestion

1. **`computeEmbeddingCost` / `calculateEmbeddingCost` / `CalculateRerankCost` are referenced but not in the diff**: Ensure these billing functions properly handle the new model IDs and have corresponding unit tests. The cost model for embeddings (per-token) vs. reranking (per query-document pair) should be clearly documented.

2. **Rerank `return_documents` handling**: The `RerankRequestBody` parses `return_documents` but the coordinator's response formatting should strip documents when the consumer didn't request them. Verify the coordinator doesn't leak document text back when `return_documents` is false.

3. **Embedding port derivation**: `embedding_port = image_port + 1` is a reasonable default, but if `image_port` is at the upper bound of valid ports, this could wrap or conflict. A more robust approach would be a dedicated env var with a validated default.

4. **Timeout asymmetry**: The coordinator uses `embeddingTimeout = 60s` while the provider's `reqwest` client also uses 60s. If the coordinator's context timeout fires first, the provider may still be processing. The cancel token should propagate to the provider's HTTP request to avoid orphaned work.

### ✅ Looks Good

1. **E2E encryption on embedding/rerank responses** — the provider can encrypt vectors back to the coordinator's session key, maintaining the same trust boundary as chat completions.

2. **Billing reservation + refund pattern** — correctly mirrors the chat path with the `s.billing != nil` guard preventing accidental credits in billing-disabled deployments.

3. **Cross-provider retry with exclusion list** — 3 attempts with provider exclusion is the right resilience pattern for small-tier providers that may be flaky.

4. **Protocol message design** — clean separation of `EmbeddingComplete` and `RerankComplete` message types with proper serde attributes for optional fields.

PR #73, opened 2026-04-19
Add internal contributor models for speculative compute

Gajesh2007:
## Summary

Add an internal-only contributor model tier plus coordinator-driven deterministic speculative execution so smaller Macs can contribute draft text compute without changing the authoritative target-model output. This also wires the contributor model through the coordinator/provider/app catalogs, tightens routing so internal models are never consumer-routable, and fixes generic inference cleanup/passthrough paths touched by the new control flow.

## Linked issue

Closes #

## Test plan

- [x] `cd coordinator && go test ./internal/registry ./internal/store ./internal/api -count=1`
- [x] `cd coordinator && go test ./internal/api -run 'TestChatCompletionsNoProvider|TestStreamingE2E' -count=1 -v -timeout 30s`
- [x] `cd coordinator && go test ./internal/registry -run 'TestReserveProviderSkipsInternalContributorModels|TestReserveSpecificProviderAllowsInternalContributorModels' -count=1 -v -timeout 20s`
- [ ] `cd provider && cargo test ...` could not be run in this environment because the installed Cargo toolchain is too old for the repo's Edition 2024 manifest

## Components touched

- [x] coordinator (Go)
- [x] provider (Rust)
- [ ] console-ui (Next.js)
- [ ] image-bridge (Python)
- [x] app (macOS Swift)
- [ ] enclave (Swift)
- [ ] infra / CI / release
- [ ] docs

## Notes for reviewers

- Internal contributor models are present in the provider-facing catalog but hidden from consumer-facing model lists and direct routing.
- The speculative path is intentionally narrow: deterministic non-streaming Qwen ChatML requests only.
- The provider default in-process path was extended so raw `/v1/completions` requests used by internal speculation can return the `echo`/`logprobs` shape the coordinator verifier expects.
- `seedModelCatalog` now refreshes built-in contributor/speculation metadata on upgrade while preserving existing operator activation choices and stored weight hashes.

<div><a href="https://cursor.com/agents/bc-2ab5814d-4ced-4541-b6c0-91210f8726bb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2ab5814d-4ced-4541-b6c0-91210f8726bb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/app/EigenInference/Sources/EigenInference/ModelCatalog.swift
+++ b/app/EigenInference/Sources/EigenInference/ModelCatalog.swift
@@ -33,6 +33,7 @@ enum ModelCatalog {
         Entry(id: "flux_2_klein_9b_q8p.ckpt", name: "FLUX.2 Klein 9B", modelType: "image", sizeGB: 13.0, architecture: "9B diffusion", description: "Higher quality image gen", minRAMGB: 24),
 
         // Text generation
+        Entry(id: "mlx-community/Qwen3.5-4B-4bit", name: "Qwen3.5 4B Contributor", modelType: "text", sizeGB: 4.0, architecture: "4B dense, draft contributor", description: "Internal draft model for disaggregated compute", minRAMGB: 16),
         Entry(id: "qwen3.5-27b-claude-opus-8bit", name: "Qwen3.5 27B Claude Opus", modelType: "text", sizeGB: 27.0, architecture: "27B dense, Claude Opus distilled", description: "Frontier quality reasoning", minRAMGB: 36),
         Entry(id: "mlx-community/Trinity-Mini-8bit", name: "Trinity Mini", modelType: "text", sizeGB: 26.0, architecture: "27B Adaptive MoE", description: "Fast agentic inference", minRAMGB: 48),
         Entry(id: "mlx-community/Qwen3.5-122B-A10B-8bit", name: "Qwen3.5 122B", modelType: "text", sizeGB: 122.0, architecture: "122B MoE, 10B active", description: "Best quality", minRAMGB: 128),
@@ -44,8 +45,7 @@ enum ModelCatalog {
         if ramGB >= 256 { return models.first { $0.id.contains("MiniMax") } }
         if ramGB >= 128 { return models.first { $0.id.contains("Qwen3.5-122B") } }
         if ramGB >= 36  { return models.first { $0.id.contains("qwen3.5-27b-claude-opus") } }
-        if ramGB >= 24  { return models.first { $0.id.contains("flux_2_klein_9b") } }
-        if ramGB >= 16  { return models.first { $0.id.contains("flux_2_klein_4b") } }
+        if ramGB >= 16  { return models.first { $0.id.contains("Qwen3.5-4B-4bit") } }
         return models.first { $0.id.contains("cohere-transcribe") }
     }
 
--- a/coordinator/cmd/coordinator/main.go
+++ b/coordinator/cmd/coordinator/main.go
@@ -397,14 +397,15 @@ func envOr(key, fallback string) string {
 	return fallback
 }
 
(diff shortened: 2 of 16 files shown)


hankbobtheresearchoor:
## Hermes Agent Review

This is a substantial feature adding speculative decoding via a 4B draft model contributor. The architecture is well-structured with clean separation between internal dispatch, contributor payout logic, and the consumer-facing path. Good test coverage on the scheduler routing rules and store round-trips.

### ⚠️ Warning

1. **Speculation eligibility bypass surface is wide**: `isDeterministicSpeculationEligible` is the gatekeeper for whether speculation runs, but it checks `temperature == 0` via float64 equality (`temperature != 0`). Floating-point consumers sending `temperature: 0.0` as a JSON number may hit edge cases depending on deserialization. Consider an epsilon threshold (e.g., `temperature > 1e-9`) to avoid accidentally speculating on near-zero-but-not-zero requests where output divergence could occur.

2. **Contributor payout at 20% with no acceptance verification**: `speculationContributorPayoutBps = 2000` grants 20% of the normal payout to the contributor, but the acceptance rate of draft tokens is not verified against the contributor's reported usage. A malicious contributor could report inflated `AcceptedTokens` and get paid for tokens the target model never actually accepted. The target provider's usage should be the authoritative source for billing.

3. **`dispatchInternalInference` has duplicated cleanup paths**: The select statement has multiple branches with nearly identical `RemovePending` + `SetProviderIdle` + error return logic. A single `defer`-based cleanup (set provider idle, remove pending) would reduce the risk of a missed cleanup on future edits.

4. **`formatQwenChatMLPrompt` only supports string content**: The function explicitly rejects non-string message content (e.g., multimodal content arrays that some chat APIs accept). This is documented but could cause silent failures if a consumer sends a valid OpenAI-format request that happens to have array content in a message — it will fail the internal dispatch and fall back to the non-speculative path. Worth logging when this happens for observability.

### 💡 Suggestion

1. **Speculation family is a free-form string**: `SpeculationFamily: "qwen-chatml"` is compared for equality but has no validation. A typo in seeding (e.g., `"qwen-chatm1"`) would silently break speculation routing. Consider defining these as constants or an enum.

2. **`internalCompletionRequestBody` uses `_eigen_internal: true` marker**: This is a reasonable guard, but consider documenting what the provider-side handler does with this flag — does it bypass rate limits? Skip logging? The security implications of a consumer sending this flag directly should be considered.

3. **The `seedModelCatalog` sync behavior change**: The change from "add-only" to "sync" (updating existing models' metadata) is good, but the log message changed from distinguishing "added N new" vs "loaded all existing" to a single "synced" count. If a deployment has many models, it would be useful to log how many were actually updated vs. unchanged for operational visibility.

### ✅ Looks Good

1. **`InternalOnly` model filtering in `ReserveProvider`** — clean approach to prevent consumer traffic from hitting contributor models while still allowing `ReserveSpecificInternalProvider` for internal dispatch.

2. **E2E encryption on the internal dispatch path** — nice that the contributor request still goes through the full E2E encryption flow, maintaining the trust boundary.

3. **Model catalog migration logic** — preserving `Active` and `WeightHash` while refreshing built-in metadata is the right approach.

4. **Comprehensive test coverage** on the scheduler routing, store round-trips, and speculation eligibility checks.

PR #1, opened 2026-03-15
DGInf: Decentralized private AI inference platform

Gajesh2007:
## Summary

Full implementation of DGInf — a platform for private, decentralized AI inference on Apple Silicon Macs.

**Four components, four languages:**
- **Provider Agent** (Rust) — hardware detection, backend management (mlx-lm/vllm-mlx), WebSocket client, request proxy, Secure Enclave attestation
- **Coordinator** (Go) — OpenAI-compatible API, provider registry, job routing, attestation verification, challenge-response, payment ledger, PostgreSQL storage
- **Consumer SDK** (Python) — OpenAI-compatible client, streaming, CLI, payments
- **Secure Enclave** (Swift) — CryptoKit P-256 identity keys, signed attestations, FFI bridge

**Trust & Privacy:**
- Apple Secure Enclave hardware-bound identity (P-256 keys that never leave the silicon)
- Signed attestations verified by coordinator (hardware model, SIP, Secure Boot)
- Periodic challenge-response (every 5 min, 3 failures = untrusted)
- Trust levels: none / self_signed / hardware (MDA scaffolded)
- Coordinator runs in GCP Confidential VM (AMD SEV-SNP)

**Payments:**
- Tempo blockchain (pathUSD stablecoin, sub-cent fees)
- Micro-USD ledger with immutable audit trail
- Per-token pricing, 90/10 provider/platform split

**Tested end-to-end** with real GPU inference (Qwen3.5-4B on Apple M3 Max Metal GPU). 150+ tests across all components.

## Test plan
- [x] Swift: 15 tests — Secure Enclave key gen, sign/verify, attestation, FFI bridge
- [x] Rust: 88 tests — hardware, backend, models, WebSocket, proxy, crypto, attestation
- [x] Go: 6 packages — API, attestation, payments, protocol, registry, store
- [x] Python: 51 tests — client, types, config, CLI, crypto, payments, trust
- [x] Integration: 20 tests — full GPU inference loop, streaming, reconnect, failure scenarios
- [x] Cross-language: Swift Secure Enclave signature verified by Go coordinator

PR #74, opened 2026-04-19
Smart prefill: disaggregating prefill across the Mac fleet

Gajesh2007:
## Summary

Stacks on top of #71 (the embeddings/rerank disaggregated-compute layer). Adds **attention-based prompt compression** so low-RAM Macs can absorb a meaningful chunk of the prefill load that would otherwise pin the big-RAM fleet.

A small-tier provider runs a tiny draft LLM (default Qwen3-0.6B, 700 MB), captures attention scores over the consumer's prompt, and returns the top-K% of tokens in order. The big-tier provider then runs **normal prefill on a 4× shorter prompt** with no engine modifications required. Expected: **2-3× lower TTFT** today, **5×+ at 128 K context** once we graduate to true sparse prefill via vllm-mlx#179 in phase 2.

This is the answer to "how do we disaggregate prefill across consumer internet?" The naive answer (ship KV cache) is dead — 8 GB per 32 K-token Qwen 27B prefill vs 125 MB/s residential pipe. The right answer is **don't ship KV, ship a shorter prompt**. Same compute leverage, ~6 orders of magnitude less bandwidth.

## Linked issue

Closes #

## What ships

**Two surfaces, one shared dispatch path:**

1. **`POST /v1/compress`** — explicit. Useful for pre-compressing a RAG corpus once and storing the compressed chunks.
2. **`smart_prefill` field on `/v1/chat/completions`** — opt-in middleware. Coordinator dispatches the longest user/system message to a tiny-tier compressor, swaps the result back in, then routes to the consumer's chosen big model. Surfaces stats via `X-SmartPrefill-*` response headers.

Both call into the same `runCompression` helper so retry, billing, and E2E encryption are guaranteed identical.

**Wire protocol** (`coordinator/internal/protocol/messages.go` ↔ `provider/src/protocol.rs`)
- New message pair: `prompt_compression_request` / `prompt_compression_complete`
- E2E encrypted under the same NaCl box session-key flow as embeddings/chat — coordinator never sees plaintext on either leg
- Round-trip tests on both sides

**Tier routing** (`coordinator/internal/registry/registry.go`)
- New `compressor` model_type joins `embedding` / `rerank` in `PreferredTiersForModelType` — routes to tiny/small tier first, falls back to standard

**Catalog** (`coordinator/cmd/coordinator/main.go` + `provider/src/main.rs`)
- `mlx-community/Qwen3-0.6B` (700 MB, 8 GB min RAM)
- `mlx-community/Qwen3-1.7B` (1.8 GB, 8 GB min RAM, higher quality)
- Cross-family drafts work (arXiv 2603.02631) → a single 0.6 B serves the entire big-tier catalog. No per-target compressors needed.

**Pricing** (`coordinator/internal/payments/pricing.go`)
- $0.004 / 1 M tokens (Qwen3-0.6B), $0.006 / 1 M (1.7B)
- Net economic impact for the consumer: ~19× positive — compression cost is dwarfed by the prefill cycles it saves on the big-tier model
- Same 95 / 5 provider / platform split

**Provider proxy** (`provider/src/proxy.rs`, `coordinator.rs`, `main.rs`)
- New `handle_compression_request` forwards to a local HTTP sidecar at `EIGENINFERENCE_COMPRESSOR_PORT` (default `embedding_port + 1`) exposing `POST /v1/compress`
- Same E2E session-key flow as embeddings — provider seals the compressed prompt back over the coordinator's session pubkey
- A first-party MLX-based sidecar will ship with the next provider bundle; until then the proxy will return 502 (connect refused) and the coordinator will route to another tiny provider via the existing retry loop

**Docs** (`docs/smart-prefill.md`) — full design, citing the recent literature: SpecPrefill (NVIDIA/MS 2025), Cross-Family Speculative Prefill (arXiv 2603.02631, March 2026), PrfaaS (arXiv 2604.15039, April 2026), BEAVER (arXiv 2603.19635, March 2026), and the open vllm-mlx#179 prototype on our exact runtime that demonstrates 5.45× TTFT reduction at 128 K.

## Review-round-2 fixes

Two independent reviewers caught real bugs in the first revision; commit `6de0a851` fixes all of them:

| Bug | Severity | Fix | Regression test |
|-----|----------|-----|-----------------|
| `smart_prefill` field leaked into chat-provider request body when middleware fell through AND consumer set explicit `max_tokens` | **Real** (vllm-mlx would reject the unknown field on the fall-through path) | Always re-marshal `rawBody` after `applySmartPrefill` — the middleware always strips the field, so the re-marshal must be unconditional too | `TestSmartPrefillStripsFieldOnFallThrough` |
| Reservation not refunded when compression succeeded but produced empty result or swap failed → consumer billed for compression they got no benefit from | **Real** | Single `refundCompression` closure called from every post-billing fall-through path (compression error, empty result, swap failure) | `TestSmartPrefillRefundsOnEmptyResult` |
| `max_keep_tokens` defined in protocol but never plumbed through middleware settings | NIT | Added to `smartPrefillSettings`, parsed from object form, forwarded to compressor request body | covered by structural tests |

Reviewer 2 also questioned the cleanup pattern in `dispatchCompression`'s decrypt-failure path; verified against the embeddings/rerank canonical flow — it's correct (pending was already removed by `handlePromptCompressionComplete` before the channel send, decrypt failure only needs `SetProviderIdle` + `RecordJobFailure`, both of which are called). Reviewer 1 independently reached the same conclusion.

## Test plan

- [x] `cd coordinator && go test ./...` — all green (8 new tests, including 2 regression tests)
- [x] `cd coordinator && gofmt -l .` — clean
- [x] `cd coordinator && go vet ./...` — clean
- [x] `rustfmt --check --edition 2024` on the Rust files I touched — clean

New tests:
- `TestCompressE2E` — full encrypted round-trip on the standalone endpoint through a 16 GB tiny provider
- `TestSmartPrefillMiddlewareSwapsLongestMessage` — full encrypted round-trip through the middleware: register a tiny compressor + a standard chat provider, assert the chat provider receives the *compressed* prompt and the response carries the `X-SmartPrefill-*` headers
- `TestSmartPrefillFallsThroughOnShortPrompt` — middleware no-ops below the threshold; chat provider sees the original prompt
- `TestSmartPrefillStripsFieldOnFallThrough` — **regression test for review-round-2 fix #1**
- `TestSmartPrefillRefundsOnEmptyResult` — **regression test for review-round-2 fix #2**, with the billing service active
- `TestCompressNoFreeCreditWhenBillingDisabled` — same regression test as embeddings (refund cannot mint balance when billing was never charged)
- `TestCompressInvalidRatio` — input validation
- `TestPreferredTiersIncludesCompressor` — tier preference wired up
- 4 new Rust protocol round-trip tests in `provider/src/protocol.rs`

## Components touched

- [x] coordinator (Go)
- [x] provider (Rust)
- [ ] console-ui (Next.js)
- [ ] image-bridge (Python)
- [ ] app (macOS Swift)
- [ ] enclave (Swift)
- [ ] infra / CI / release
- [x] docs

- [x] Yes — described above and matching side updated

Both sides updated symmetrically; round-trip tests on both sides catch drift. The `smart_prefill` field on `/v1/chat/completions` is a new request extension; we strip it from the body before forwarding to the provider so the existing OpenAI-compatible providers never see it.

No bundle script changes — the compressor sidecar runs at a deterministic port (`embedding_port + 1` or `EIGENINFERENCE_COMPRESSOR_PORT`) so the launcher and the proxy agree without extra IPC.

## Notes for reviewers

**Why now and why this design:**

I asked Exa for the latest literature before writing a line of code. The field moved past LLMLingua-2 — the new winner is attention-based "Speculative Prefill" using a small draft LLM (NVIDIA/MS 2025), with cross-family drafts proven viable in March 2026. There's an open prototype of exactly this on the vllm-mlx fork our providers run (waybarrios/vllm-mlx#179) that reports **5.45× TTFT reduction on Qwen3.5-122B at 128 K context** on M2 Ultra. So this PR ships the protocol envelope + tier routing + billing now, and the next iteration graduates from text-level compression (phase 1, this PR) to true sparse prefill against the original positional schema (phase 2, follow-up). Clients don't change between phases — the same `smart_prefill` flag governs both.

**What this is NOT:**

- Not bit-exact prefill disaggregation. That requires shipping KV cache, which is bandwidth-bound across consumer internet and dead on arrival.
- Not speculative *decoding*. That accelerates decode using a draft on the *same* machine. Both will be supported, neither replaces the other.
- Not lossless. Phase 1 drops ~75 % of input tokens. Quality on LongBench/RULER at 4× is consistently >90 % across the cited research, but consumers needing verbatim recall (legal, code, exact-quote retrieval) should leave `smart_prefill` off.

**Stacking note:** This PR's base branch is `cursor/disaggregated-compute-small-providers-40ff` (PR #71), not `master`. The diff against #71 is what this PR adds; the diff against `master` is the union. Once #71 lands, GitHub will auto-rebase this against `master`.

<div><a href="https://cursor.com/agents/bc-4a6ac94b-fdef-4dfd-819a-edee318440ff"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-4a6ac94b-fdef-4dfd-819a-edee318440ff"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/coordinator/cmd/coordinator/main.go
+++ b/coordinator/cmd/coordinator/main.go
@@ -433,6 +433,16 @@ func seedModelCatalog(st store.Store, logger *slog.Logger) {
 		{ID: "mlx-community/mxbai-embed-large-v1", S3Name: "mxbai-embed-large-v1", DisplayName: "mxbai Large Embeddings", ModelType: "embedding", SizeGB: 0.7, Architecture: "335M BERT-large", Description: "English embeddings, top of MTEB", MinRAMGB: 8, Active: true},
 		{ID: "mlx-community/bge-reranker-v2-m3", S3Name: "bge-reranker-v2-m3", DisplayName: "BGE Reranker v2-m3", ModelType: "rerank", SizeGB: 1.2, Architecture: "568M cross-encoder", Description: "Multilingual cross-encoder reranker", MinRAMGB: 8, Active: true},
 		{ID: "mlx-community/Qwen3-Reranker-0.6B", S3Name: "Qwen3-Reranker-0.6B", DisplayName: "Qwen3 0.6B Reranker", ModelType: "rerank", SizeGB: 0.7, Architecture: "0.6B cross-encoder", Description: "Tiny multilingual reranker", MinRAMGB: 8, Active: true},
+
+		// --- Smart-prefill compressor models (attention-based prompt compression) ---
+		// A small draft LLM hosted on a tiny-tier provider compresses the
+		// consumer's prompt to ~1/4 of its original length using its own
+		// attention scores. The big-tier provider then runs normal prefill on
+		// the shorter prompt — typically 2-3x lower TTFT at >90% task quality.
+		// Cross-family drafts work (arXiv 2603.02631), so a single Qwen3-0.6B
+		// can serve as compressor for every model in the catalog above.
+		{ID: "mlx-community/Qwen3-0.6B", S3Name: "Qwen3-0.6B", DisplayName: "Qwen3 0.6B (Compressor)", ModelType: "compressor", SizeGB: 0.7, Architecture: "0.6B Qwen3 dense", Description: "Smart-prefill draft compressor", MinRAMGB: 8, Active: true},
+		{ID: "mlx-community/Qwen3-1.7B", S3Name: "Qwen3-1.7B", DisplayName: "Qwen3 1.7B (Compressor)", ModelType: "compressor", SizeGB: 1.8, Architecture: "1.7B Qwen3 dense", Description: "Smart-prefill draft compressor (higher quality)", MinRAMGB: 8, Active: true},
 	}
 
 	added := 0
--- a/coordinator/internal/api/compress.go
+++ b/coordinator/internal/api/compress.go
@@ -1,0 +1,376 @@
+package api
+
+// Smart-prefill prompt compression — POST /v1/compress.
+//
+// A small-tier provider runs a tiny draft LLM (e.g. Qwen3-0.6B), captures
+// per-token attention scores over the consumer's prompt, and returns the
+// top-K% of tokens in order. The coordinator either:
+//
+//   * returns the compressed prompt directly (this endpoint), or
+//   * swaps it into a chat-completion request before dispatching to the
+//     big-tier model (smart_prefill middleware in consumer.go).
+//
+// Either way the consumer's big-tier prefill cost drops by 2-5x at
+// >90% task quality. See docs/smart-prefill.md for the full design.
+//
+// Architecture, billing, refund-on-error, retry, E2E encryption, and
+// tier preference all mirror /v1/embeddings — see embeddings.go for the
+// canonical pattern. This file is intentionally a thin reuse.
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/eigeninference/coordinator/internal/e2e"
+	"github.com/eigeninference/coordinator/internal/payments"
+	"github.com/eigeninference/coordinator/internal/protocol"
+	"github.com/eigeninference/coordinator/internal/registry"
+	"github.com/eigeninference/coordinator/internal/store"
+	"github.com/google/uuid"
+	"nhooyr.io/websocket"
+)
+
+const (
+	// compressionTimeout — compression on a 0.6B draft is single-pass and
+	// fast (<5s for 32k tokens on M2 Air per vllm-mlx#179 numbers); 60s
+	// covers cold-start of the draft model.
+	compressionTimeout = 60 * time.Second
+
+	// defaultCompressorModel is what the smart-prefill middleware uses
+	// when the consumer doesn't specify one. Qwen3-0.6B is small enough
+	// for a 8 GB Air and works as a cross-family draft for every model
+	// in the big-tier catalog (cross-family draft validity from
+	// arXiv 2603.02631).
+	defaultCompressorModel = "mlx-community/Qwen3-0.6B"
+
+	// defaultCompressionRatio is 0.25 = 4x compression. The cited
+	// research consistently shows >90% LongBench quality at 4x; below
+	// that the big-tier model loses too much context.
+	defaultCompressionRatio = 0.25
+)
+
+// compressionMaxAttempts mirrors embeddingMaxAttempts (3) so a single
+// flaky compressor doesn't 502 the consumer.
+const compressionMaxAttempts = 3
+
+// handleCompress handles POST /v1/compress.
+//
+// Request body:
+//
+//	{ "prompt": "...", "compressor_model": "...", "target_ratio": 0.25, "min_keep_tokens": 64 }
+//
+// Returns:
+//
+//	{ "compressed_prompt": "...", "model": "...",
+//	  "usage": { "original_tokens": 4096, "compressed_tokens": 1024 } }
+func (s *Server) handleCompress(w http.ResponseWriter, r *http.Request) {
+	rawBody, err := io.ReadAll(r.Body)
+	if err != nil {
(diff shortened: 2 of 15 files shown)


hankbobtheresearchoor:
## Hermes Agent Review

This is a well-designed feature. The smart-prefill design is architecturally sound — shipping text-level compression now with a path to sparse prefill later is the right call given KV-cache bandwidth constraints. The review-round-2 fixes for the field-leak and refund bugs were real catches.

### ⚠️ Warning

1. **Compression is lossy with no quality fallback signal to the consumer** — When smart_prefill middleware falls through (short prompt, no compressor available), the consumer gets a normal response — good. But when compression *succeeds*, the consumer has no way to know the quality impact on their specific request. Consider adding an opt-in X-SmartPrefill-Quality-Estimate header or a quality_warning field so consumers in loss-sensitive domains (legal, code) can log/alert when compression was applied.

2. **defaultCompressionRatio = 0.25 is hardcoded** — The 4x compression ratio is based on LongBench/RULER benchmarks, but quality varies by domain. If an operator needs a different default (e.g., 0.5 for medical/legal customers), they currently need a code change. Consider making this configurable via environment variable or coordinator config.

3. **dispatchCompression decrypt-failure path does not call cleanup()** — When the CompressionCh returns a result with EncryptedData and decryption fails, the code calls RecordJobFailure and returns, but RemovePending was already called by handlePromptCompressionComplete before the channel send. The idle-setting path is correct per the PR description, but this asymmetry with the cleanup() closure is fragile — a future refactor could easily miss that RemovePending is handled elsewhere. Consider a comment or assertion.

### 💡 Suggestion

4. **compressionTimeout = 60s may be too generous** — A 0.6B model on M2 Air compresses 32k tokens in under 5s. The 60s timeout covers cold-start, but a cold-started compressor that takes 55s will block the consumer's request for nearly a minute before retry. Consider a shorter timeout (e.g., 15s) for the first attempt and a longer timeout for subsequent attempts, or a separate cold-start detection path.

5. **approxTokens uses 4-chars-per-token for compression prompts** — This is the same heuristic as embeddings, but compression prompts can be much longer (32k+ tokens). A 4-chars-per-token estimate can be off by 2x for CJK text or code-heavy prompts, affecting billing accuracy. Consider using the tokenizer count from the compressor response for final billing (which is already done), but document that the reservation may be significantly over/under for non-English text.

6. **Smart-prefill middleware preserve_boundaries is accepted but not enforced** — The PromptCompressionRequestBody has a preserve_boundaries field that flows through the protocol, but the provider-side do_compression just forwards the body verbatim to the sidecar. If the sidecar does not implement boundary preservation, the flag is silently ignored. Consider either (a) documenting that this is sidecar-dependent, or (b) adding a response header indicating whether boundaries were preserved.

7. **No rate limiting on /v1/compress** — The standalone endpoint allows unlimited compression requests. Since each request dispatches to a tiny-tier provider, a malicious consumer could exhaust the compressor fleet. The chat path has implicit rate limiting via billing, but the compress endpoint should have explicit per-consumer rate limits.

### ✅ Looks Good

8. The two-surface design (standalone /v1/compress + middleware smart_prefill field) is clean — same runCompression helper ensures billing/retry/E2E parity.

9. Review-round-2 fixes are solid: the applySmartPrefill re-marshal fix prevents field leakage, and the refundCompression closure covers all error paths.

10. E2E encryption is mandatory on both request and response legs — the coordinator never sees plaintext prompts or compressed results. This is consistent with the chat path trust boundary.

11. The tier routing design (compressor to tiny/small first, fallback to standard) is correct for fleet utilization.

PR #77, opened 2026-04-20
Arya/privacy audit handoff 2026 04 20

aryabhuptani:
This draft PR contains the Darkbloom privacy audit handoff packet plus proof-backed tests documenting the main coordinator/provider privacy gaps found during review.

Included:
- audit report and POC appendix under `audits/2026-04-20-privacy-review/`
- coordinator-side Go proof tests
- provider-side Rust proof tests
- the small provider-side support fixes needed to keep those proofs portable

This branch is intended as review material first, not as an assumption that every included proof should necessarily be merged unchanged. The goal is to give the team a concrete, test-backed handoff they can evaluate and triage.


hankbobtheresearchoor:
## Hermes Agent Review
**Verdict:** Needs changes

This is a thorough and valuable audit handoff. The proof-backed approach is excellent. Below are findings on the code changes themselves, not the audit content.

### 🔴 Critical

1. **Plaintext request fallback still accepted (findings C1-C3)** — The three new `handle_*_request_message` functions all fall back to plaintext `body` when `encrypted_body` is `None`. The PR correctly documents this as a privacy gap but does not enforce encrypted-only delivery. Given that this PR explicitly surfaces these issues, it should either (a) reject plaintext requests outright with a clear error, or (b) add a config flag / feature flag to enforce encrypted-only mode, so operators can close this gap without a further code change. The current state preserves the vulnerability by design.

2. **Temp file leak in transcription path (finding E)** — The POC test proves `/tmp/eigeninference-stt-<request_id>.wav` is left behind on failure. This is a concrete sensitive-data escape to disk. The fix (cleanup in the error path) is straightforward and should be included in this PR rather than left as a follow-up, since the PR already touches `proxy.rs` to add the proof test.

### ⚠️ Warning

3. **`Cargo.toml` dependency reordering changes build caching** — Moving the macOS-specific `security-framework` / `security-framework-sys` / `core-foundation` dependencies from before `futures-util` to after `pyo3` is a pure reordering. On Linux CI this has no effect (those deps are `cfg(target_os = "macos")`), but on macOS it changes the Cargo feature unification order. Worth a clean rebuild verification on macOS.

4. **`main.rs` `#[cfg(feature = "python")]` type annotation** — Adding an explicit type annotation `Option<std::sync::Arc<inference::SharedEngine>>` behind the `#[cfg(feature = "python")]` gate is correct, but the non-python path no longer declares `inprocess_engine` at all. If there are any unconditional references to `inprocess_engine` elsewhere in `cmd_serve`, the non-python build will fail. The PR description says `cargo check --tests` passes, but a full `cargo check` without the `python` feature should be verified.

5. **Proof tests use `time.Sleep` for synchronization** — Multiple tests use `time.Sleep(100ms)` / `time.Sleep(200ms)` to wait for provider registration to propagate. This is inherently flaky under load. Consider using a channel or callback-based synchronization instead.

### 💡 Suggestion

6. **Audit documents should note which findings are merge-blocking vs. advisory** — The report and appendix are well-structured, but it would help the team to have an explicit "merge-blocking" vs. "tracking" classification for each verified finding, so PRs that depend on this branch know what to gate on.

7. **Consider adding a `//go:build privacypoc` build tag** — The proof tests are valuable but may not need to run in every CI cycle. A build tag would let them be run on-demand while keeping the standard test suite fast.

### ✅ Looks Good

8. The three-pass methodology (theory → semantic → proof-backed) is rigorous and well-documented.

9. The remediation priority ordering (P0: coordinator plaintext exposure, P1: encrypted-only enforcement, P2: trust tightening, P3: temp-file cleanup, P4: documentation) is sensible.

10. Extracting the duplicated inference/transcription/image decryption logic into separate `handle_*_request_message` functions is a genuine code quality improvement beyond just the audit.

PR #78, opened 2026-04-21
fix: Remove custom codeql in-favor of repo default workflow

ethenotethan:
### Problem
CodeQL only allows one active configuration per repository, but we had two running simultaneously:
- **Default setup** — enabled via GitHub repository settings (covers all languages)
- **Custom action** — defined in `.github/workflows/codeql.yml` (scans Go and JS only)

This caused the following error when the custom action tried to upload its SARIF results:
```
CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled
```

### Investigation
Attempted to validate the SARIF output locally using `codeql database bundle --validate` to rule out a malformed file, but IT system policy blocked the CodeQL binary from running:
```
libjli.dylib not valid for use in process: library load disallowed by system policy
```
Given the specificity of the error message, the root cause is the configuration conflict, not the SARIF file itself.

### Fix
Remove the custom workflow in favour of the default GitHub-managed setup, which has broader language coverage across the repo.

### Impact
- Resolves the CodeQL SARIF upload failure
- Retains security scanning via the default setup (all languages)
- No loss of Go or JS coverage


Gajesh2007 (approved):
LGTM

PR #94, opened 2026-04-28
Add provider discounts and cost routing

Gajesh2007:
## Summary
- add provider discount storage/resolution with account-global, account/model, machine-global, and machine/model precedence
- expose authenticated provider pricing endpoints and wire discounted billing/payout math through coordinator paths
- add consumer routing preference parsing for body/header/query, preserve it through queues, and support lowest-cost routing
- add darkbloom pricing CLI commands plus provider pricing and chat routing UI updates

## Validation
- PASS: `cd coordinator && GOCACHE=/tmp/go-build go test ./internal/pricing ./internal/store ./internal/registry ./internal/api`
- PASS: `cd console-ui && npm test`
- PASS: `cd console-ui && npx eslint src/` (97 existing warnings, 0 errors)
- PASS: `cd console-ui && npm run build`
- PASS: `cd provider && cargo fmt --check`
- PASS: `cd provider && PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 cargo test pricing_cli -- --nocapture`
- PASS: `cd provider && PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 cargo test test_pricing_provider_key_validation -- --nocapture`
- PASS: focused rerun of `coordinator::tests::test_coordinator_connect_register_and_receive`
- FAIL: full `cd provider && PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 cargo test -- --nocapture` hit `coordinator::tests::test_coordinator_connect_register_and_receive` in the full suite; first run failed with `expected at least register and error messages, got 1`, second run failed with `expected an inference_error message` and then hung in `secure_enclave_key::tests::test_ephemeral_handle_attestation_without_binary_hash`, so I killed the stuck process
- BLOCKED/EXISTING: `cd console-ui && npx tsc --noEmit` still fails on pre-existing unrelated type errors in `EarningsContent.tsx`, `DatadogRUM.tsx`, `telemetry.ts`, and `vitest.config.ts`

## Notes
- Discounts are clamped to 0-90% and are stored internally as basis points while exposing `discount_percent` externally.
- Provider pricing writes accept provider auth surfaces but reject consumer API keys; machine-scoped writes verify ownership.
- Cost routing optimizes estimated consumer charge and leaves the default performance path unchanged.

### DIFF
--- a/console-ui/__tests__/api-routes.test.ts
+++ b/console-ui/__tests__/api-routes.test.ts
@@ -108,6 +108,40 @@ describe("GET /api/me/summary", () => {
 });
 
 // =========================================================================
+// POST /api/chat
+// =========================================================================
+
+describe("POST /api/chat", () => {
+  it("forwards routing preference header and query param to coordinator", async () => {
+    upstreamFetch.mockResolvedValueOnce(
+      new Response("ok", {
+        status: 200,
+        headers: { "Content-Type": "text/plain" },
+      })
+    );
+
+    const { POST } = await import("@/app/api/chat/route");
+    const req = makeRequest("/api/chat?routing_preference=cost", {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        "x-api-key": "key123",
+        "x-darkbloom-routing-preference": "cost",
+      },
+      body: JSON.stringify({ model: "model-a", messages: [] }),
+    });
+    const res = await POST(req);
+
+    expect(res.status).toBe(200);
+    const [upstreamUrl, upstreamOpts] = upstreamFetch.mock.calls[0];
+    expect(String(upstreamUrl)).toBe(`${DEFAULT_COORD}/v1/chat/completions?routing_preference=cost`);
+    expect(upstreamOpts.headers.Authorization).toBe("Bearer key123");
+    expect(upstreamOpts.headers["X-Darkbloom-Routing-Preference"]).toBe("cost");
+    expect(JSON.parse(upstreamOpts.body)).toEqual({ model: "model-a", messages: [] });
+  });
+});
+
+// =========================================================================
 // GET /api/payments/balance
 // =========================================================================
 
--- a/console-ui/__tests__/api.test.ts
+++ b/console-ui/__tests__/api.test.ts
@@ -7,6 +7,7 @@ import {
   fetchModels,
   fetchPricing,
   healthCheck,
+  streamChat,
 } from "@/lib/api";
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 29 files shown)


hankbobtheresearchoor:
## Hermes Agent Review

This PR adds a provider discount system with cost-based routing preference, spanning the store layer, scheduler, provider CLI, and console UI. Substantial feature with good test coverage.

### Findings

Warning: Discount BPS validation is split across layers
- The provider CLI validates 0-90% range, but the store layer (SetProviderDiscount) accepts any BPS value including negatives or values exceeding 10000 (100%). The admin API handler should also enforce bounds server-side. A malicious or buggy client could set discount_bps to 99999, resulting in negative prices.
- Add server-side validation in SetProviderDiscount (or the API handler) to reject out-of-range BPS values.

Warning: CalculateCost during routing holds the registry lock
- The scheduler calls providerpricing.CalculateCost(r.store, ...) inside buildCandidateWithReason, which is called under the registry mutex (selectBestCandidateLockedFull). If the store query is slow (e.g., Postgres under load), this blocks all routing decisions for all models. This is a latency and availability risk.
- Consider pre-computing/caching effective prices when providers register or when discounts change, rather than querying the store on every routing decision.

Warning: Routing preference header is trusted without validation
- The chat route forwards x-darkbloom-routing-preference header and routing_preference query param directly to the coordinator. While the coordinator normalizes it (normalizeRoutingPreference), there is no input validation on the console-ui side. A crafted value could be logged or stored as-is in observability tools.
- Consider validating the value in the console proxy before forwarding.

Suggestion: nearTie logic for cost mode may be too strict
- When routing_preference is "cost", nearTie returns false if prices differ at all. This means two providers with prices 1 cent apart are never considered near-ties, even if their latency difference is massive. A small price tolerance (e.g., within 5%) might produce better user outcomes.

Suggestion: providerpricing.CalculateCost error handling
- If the store lookup fails or returns unexpected data, CalculateCost could return 0, making a provider appear free. Verify that CalculateCost returns a meaningful baseline price as fallback.

Looks Good
- The ProviderDiscount.Scope() method provides a clean abstraction for the four discount levels (account_global, account_model, machine_global, machine_model).
- The scheduler test TestReserveProviderCostRoutingUsesDiscountedPrice is thorough and verifies both performance and cost preference paths.
- The ListProviderDiscounts test correctly verifies that the returned slice is a copy (mutation isolation).
- The CLI pricing command with --this-machine flag is a nice UX touch for providers.
- Routing preference is persisted in localStorage and transmitted via both header and body — good wire compatibility.
- The Postgres migration uses ON CONFLICT DO UPDATE for upsert — correct idempotent pattern.

PR #97, opened 2026-04-28
Remove stale Python integration test

hankbobtheresearchoor:
Removes `tests/integration_test.py` which is superseded by the coordinator's Go-based integration tests in `coordinator/internal/api/`.

### Why remove

- **Coverage duplicated**: `integration_test.go` covers provider protocol flow (register, challenge, heartbeat, inference) with Go simulated providers — same assertions, no binary build required
- **Full-stack GPU tests**: `fullstack_integration_test.go` tests real vllm-mlx backends (gated behind `LIVE_FULLSTACK_TEST=1`)
- **Stale**: References old binary names (`eigeninference-provider`), old flags (`--backend mlx-lm`), and predates attestation challenges, E2E encryption, and the vllm-mlx migration — most test paths would skip or fail against current code
- **No unique coverage**: Coordinator `main.go` wiring (env vars, Postgres, billing) is not tested here either — no `EIGENINFERENCE_DATABASE_URL` is set, so it hits the same `store.NewMemory()` fallback as every other test
- **SDK tests belong in SDK repo**: Tests 4.5.x exercise `from eigeninference import EigenInference` — that's a dependency boundary the SDK's own test suite should own

No test coverage is lost that isn't already covered elsewhere in the coordinator's Go test suite.


hankbobtheresearchoor:
## Hermes Agent Review
**Verdict:** LGTM

This PR removes the stale Python integration test suite (`tests/integration_test.py`, 684 lines). The test referenced legacy binary paths, the old EigenInference branding, and hardcoded infrastructure assumptions that are no longer valid.

✅ **Looks Good**
- Clean deletion with no orphaned references left behind — no other files import or depend on `tests/integration_test.py`.
- The test relied on local binaries (`coordinator/bin/coordinator`, `provider/target/release/eigeninference-provider`), a specific model download (`mlx-community/Qwen3.5-4B-MLX-4bit`), and hardcoded ports — all of which make it brittle and unsuitable for CI.
- Removing dead test code reduces maintenance burden and avoids false confidence from a suite that was likely never run in automation.
- The existing Go and TypeScript test suites in `coordinator/` and `console-ui/` provide coverage for the same functionality.

💡 **Suggestion**
- If integration-level E2E testing is still desired, consider adding a CI-compatible smoke test (perhaps as a GitHub Actions workflow with service containers) rather than relying on local-only Python scripts. The old test structure (4.1–4.7 groups) could serve as a specification for a future rewrite.

Gajesh2007 (approved):

PR #98, opened 2026-04-28
fix: Rename all EigenInference references to Darkbloom

ethenotethan:
## Summary
Agentic workflows will need to carve meaning in this codebase. simple things like conflicting class names can cause them to draw incorrect relationships when forming higher order knowledge. 

Any variables to `EigenInference` used for persistent resource definitions (e.g, postgres keys, table names) or env have been kept in-tact to not break existing usability. 

## Linked issue

Closes #

## Test plan

- [ ]
- [ ]

## Components touched

- [ ] coordinator (Go)
- [ ] provider (Rust)
- [ ] console-ui (Next.js)
- [ ] image-bridge (Python)
- [ ] app (macOS Swift)
- [ ] enclave (Swift)
- [ ] infra / CI / release
- [ ] docs

- [ ] No protocol/interface changes


coordinator/internal/e2e/coordinator_key.go
```diff
@@ -28,7 +28,7 @@ import (
 // CoordinatorKeyHKDFInfo is the HKDF info string used to separate the
 // X25519 encryption key from any other key derived from the same mnemonic.
 // Bumping the version here rotates the coordinator key for all senders.
-const CoordinatorKeyHKDFInfo = "eigeninference-coordinator-e2e-v1"
+const CoordinatorKeyHKDFInfo = "darkbloom-coordinator-e2e-v1"
```
ethenotethan: could break backwards compat

deploy/datadog/apply-dev-dashboard.sh
```diff
@@ -22,15 +22,15 @@ APP_KEY="${DD_APPLICATION_KEY:-${DD_APP_KEY:-${DD_WRITE_KEY:-}}}"
...
 fi
 
 if [[ -z "$APP_KEY" ]]; then
-  APP_KEY="$(fetch_secret eigeninference-dd-app-key || true)"
+  APP_KEY="$(fetch_secret darkbloom-dd-app-key || true)"
 fi
 
 if [[ -z "$SITE" ]]; then
-  SITE="$(fetch_secret eigeninference-dd-site || true)"
+  SITE="$(fetch_secret darkbloom-dd-site || true)"
```
ethenotethan, line 40: breaks backwards compat

hankbobtheresearchoor:
## PR #98 Review: EigenInference → Darkbloom Rename

Reviewed the full diff (167 files, ~870 additions / ~857 deletions). The bulk of the rename is correct and consistent, but there are several **incorrect renames** that break migration paths, one **build-breaking** issue, and a couple of **wire-protocol changes** that need attention.

### Critical — Broken Migration Paths (data loss for existing users)

1. **`console-ui/src/lib/migrate-storage.ts`** — The `KEY_MAP` was changed from `["eigeninference_*", "darkbloom_*"]` to `["darkbloom_*", "darkbloom_*"]`. This is now a no-op migration: existing users with `eigeninference_*` localStorage keys will never be migrated. The old source keys must be kept.

2. **`scripts/install.sh`** — The directory migration chain was changed from `~/.dginf → ~/.eigeninference → ~/.darkbloom` to `~/.dginf → ~/.darkbloom → ~/.darkbloom`. Existing `~/.eigeninference` directories will no longer be migrated.

3. **`provider/src/secure_enclave_key.rs`** `LEGACY_KEY_FILES` — `.eigeninference/` entries were replaced with `.darkbloom/`, creating duplicate entries in the array. The `.eigeninference/` paths must be kept so old key files are cleaned up.

4. **`provider/src/crypto.rs`** `legacy_secret_paths` — Same issue: `[".darkbloom", ".dginf", ".eigeninference"]` → `[".darkbloom", ".dginf", ".darkbloom"]`. The `.eigeninference` cleanup path is lost.

### Critical — Build-Breaking

5. **Go module path** — `github.com/eigeninference/coordinator` → `github.com/darkbloom/coordinator` (and similarly for analytics). The `github.com/darkbloom` org/repo does not exist on GitHub, so `go mod tidy` and `go build` will fail. Either the org needs to be created/mirrored, or the module path should stay at the actual repo location.

### Wire-Protocol Changes (provider↔coordinator compatibility)

6. **`SealedContentType`** — `application/eigeninference-sealed+json` → `application/darkbloom-sealed+json`. Both the coordinator and all senders (console-ui, SDK) must update simultaneously, or sealed requests will fail. Flagged for awareness — OK if this is a coordinated deploy.

7. **API key prefix** — New keys will have `darkbloom-` prefix instead of `eigeninference-`. Existing keys still validate (hash-based lookup), but any client code that checks the prefix string will break.

8. **Device token prefix** — `eigeninference-pt-` → `darkbloom-pt-`. Same concern.

### Incorrect Over-Rename

9. **`coordinator/cmd/coordinator/main.go`** line 1 — Comment changed from `"Darkbloom (EigenInference)"` to `"Darkbloom (Darkbloom)"`. The parenthetical was the old name and should have been kept or removed, not double-renamed.

### Minor

10. **`scripts/install.sh`** — Temp file path `/tmp/eigen-site-packages.tar.gz` was not renamed while the download URL changed to `darkbloom-site-packages.tar.gz`. Cosmetic but confusing.

11. **R2 CDN URL** `https://pub-3d1cb668259340eeb2276e1d375c846d.r2.dev` and **IP** `34.197.17.112` remain hardcoded in scripts — pre-existing, not introduced by this PR.

Inline comments follow for the file-specific issues.

app/Darkbloom/Sources/Darkbloom/ConfigManager.swift
```diff
@@ -53,7 +53,7 @@ enum ConfigManager {
         ).first!
 
         let newPath = appSupport.appendingPathComponent("darkbloom").appendingPathComponent("provider.toml")
-        let legacyPath = appSupport.appendingPathComponent("eigeninference").appendingPathComponent("provider.toml")
+        let legacyPath = appSupport.appendingPathComponent("darkbloom").appendingPathComponent("provider.toml")
```
ethenotethan: rm

analytics/README.md
```diff
@@ -1,6 +1,6 @@
 # Analytics
 
-Standalone read-only analytics service for Darkbloom / EigenInference.
+Standalone read-only analytics service for Darkbloom / Darkbloom.
```
ethenotethan: update

console-ui/src/hooks/useAuth.ts
```diff
@@ -5,7 +5,7 @@ import { useAuthContext } from "@/components/providers/PrivyClientProvider";
 import { trackEvent } from "@/lib/google-analytics";
 
 const API_KEY_STORAGE = "darkbloom_api_key";
-const OLD_API_KEY_STORAGE = "eigeninference_api_key";
+const OLD_API_KEY_STORAGE = "darkbloom_api_key";
```
ethenotethan: rm

coordinator/deploy/start.sh
```diff
@@ -61,9 +77,9 @@ if [ -n "$MICROMDM_API_KEY" ]; then
     if [ -n "$MDM_PUSH_P12_B64" ] && [ ! -f /data/micromdm/push.crt ]; then
         echo "Decoding MDM push certificate from PKCS#12..."
         printf '%s' "$MDM_PUSH_P12_B64" | tr '_-' '/+' | base64 -d > /tmp/push.p12
-        openssl pkcs12 -in /tmp/push.p12 -clcerts -nokeys -passin pass:eigeninference \
+        openssl pkcs12 -in /tmp/push.p12 -clcerts -nokeys -passin pass:darkbloom \
```
hankbobtheresearchoor, line 100: 🔴 **PKCS#12 password regression**: This changes the push cert decryption password from `eigeninference` to `darkbloom`. If the existing push cert in GCP Secret Manager was encrypted with the old password, this will fail to decrypt it.

**Fix**: Try the new password first, then fall back to the old one:
```bash
openssl pkcs12 -in /tmp/push.p12 -clcerts -nokeys -passin pass:darkbloom \
    -out /data/micromdm/push.crt 2>/dev/null || \
openssl pkcs12 -in /tmp/push.p12 -clcerts -nokeys -passin pass:eigeninference \
    -out /data/micromdm/push.crt 2>/dev/null
```

coordinator/cmd/coordinator/main.go
```diff
@@ -413,7 +419,7 @@ func main() {
 	if mdmURL := os.Getenv("EIGENINFERENCE_MDM_URL"); mdmURL != "" {
 		mdmKey := os.Getenv("EIGENINFERENCE_MDM_API_KEY")
 		if mdmKey == "" {
-			mdmKey = "eigeninference-micromdm-api" // default
+			mdmKey = "darkbloom-micromdm-api" // default
```
hankbobtheresearchoor: 🔴 **MDM API key default regression**: Fallback changed from `eigeninference-micromdm-api` to `darkbloom-micromdm-api`. If `EIGENINFERENCE_MDM_API_KEY` env var is not set, the coordinator will send the wrong API key to an existing MicroMDM instance.

**Fix**: Check the old default as fallback, similar to how GCP secrets work:
```go
mdmKey = "darkbloom-micromdm-api" // default
// Legacy: if new default fails, try old default
```

coordinator/deploy/start.sh
```diff
@@ -12,7 +12,7 @@ ln -sfn "$PERSIST" /data
 if [ ! -d "/data/step-ca/config" ]; then
     echo "Initializing step-ca (first boot)..."
     mkdir -p /data/step-ca/secrets
-    echo "eigeninference-step-ca" > /data/step-ca/secrets/password
+    echo "darkbloom-step-ca" > /data/step-ca/secrets/password
```
hankbobtheresearchoor, line 23: 🔴 **Step-CA password regression**: Password file changed from `eigeninference-step-ca` to `darkbloom-step-ca`. On an existing coordinator VM with persistent `/data/step-ca/`, if the CA was initialized with the old password, step-ca will fail to unlock.

Note: The first-boot guard (`if [ ! -d "/data/step-ca/config" ]`) prevents this on existing VMs — but on a **fresh deploy or re-initialization**, the new password is used and any manually-copied CA data would be inaccessible.

Consider: Try new password first, fall back to old password on unlock failure.

coordinator/deploy/start.sh
```diff
@@ -23,18 +23,30 @@ if [ ! -d "/data/step-ca/config" ]; then
         --name "Darkbloom CA" \
         --dns "${DOMAIN:-localhost}" \
         --address ":9000" \
-        --provisioner "eigeninference-admin" \
+        --provisioner "darkbloom-admin" \
```
hankbobtheresearchoor, line 34: 🔴 **Step-CA provisioner name regression**: Provisioner changed from `eigeninference-admin` to `darkbloom-admin`. Existing client cert signing requests that reference `eigeninference-admin` as the provisioner will fail.

The ACME provisioner correctly has an alias (`eigeninference-acme` is preserved), but the JWK provisioner (`eigeninference-admin` → `darkbloom-admin`) has no such alias.

**Fix**: Add `eigeninference-admin` as a second JWK provisioner in the step-CA config, similar to the ACME alias pattern already used.

app/Darkbloom/Sources/Darkbloom/ConfigManager.swift
```diff
@@ -52,16 +52,7 @@ enum ConfigManager {
...
-        let legacyPath = appSupport.appendingPathComponent("eigeninference").appendingPathComponent("provider.toml")
-
-        if FileManager.default.fileExists(atPath: newPath.path) {
-            return newPath
-        }
-        if FileManager.default.fileExists(atPath: legacyPath.path) {
-            return legacyPath
-        }
-        return newPath
+        return appSupport.appendingPathComponent("darkbloom").appendingPathComponent("provider.toml")
```
hankbobtheresearchoor: 🔴 **Legacy config path migration removed**: The old code checked both `darkbloom/` and `eigeninference/` paths and returned whichever existed. The new code only checks `darkbloom/`. Users upgrading from `EigenInference.app` who have config at `~/Library/Application Support/eigeninference/provider.toml` will lose their existing config.

**Fix**: Restore the fallback:
```swift
let newPath = appSupport.appendingPathComponent("darkbloom").appendingPathComponent("provider.toml")
let legacyPath = appSupport.appendingPathComponent("eigeninference").appendingPathComponent("provider.toml")
if FileManager.default.fileExists(atPath: newPath.path) {
    return newPath
}
if FileManager.default.fileExists(atPath: legacyPath.path) {
    return legacyPath
}
return newPath
```

provider/src/main.rs
```diff
@@ -5854,34 +5851,31 @@ async fn cmd_update(coordinator: String, force: bool) -> Result<()> {
...
     // Move binaries to bin dir
+    let _ = std::fs::rename(darkbloom_dir.join("darkbloom"), bin_dir.join("darkbloom"));
     let _ = std::fs::rename(
-        eigeninference_dir.join("darkbloom"),
-        bin_dir.join("darkbloom"),
-    );
-    let _ = std::fs::rename(
-        eigeninference_dir.join("eigeninference-enclave"),
-        bin_dir.join("eigeninference-enclave"),
+        darkbloom_dir.join("darkbloom-enclave"),
```
hankbobtheresearchoor, line 5870: 🟠 **Enclave binary name**: `darkbloom-enclave` is now the expected binary name, but existing installations have `eigeninference-enclave` in `~/.darkbloom/bin/`. Consider checking both names during `cmd_doctor` and `cmd_update`, similar to how `default_config_path()` falls back to the legacy path.

provider/src/main.rs
```diff
@@ -931,8 +931,8 @@ fn ensure_python_verified(python_cmd: &str, coordinator_base: &str) -> bool {
             match serde_json::from_slice::<serde_json::Value>(&output.stdout) {
                 Ok(release) => release.get("url").and_then(|v| v.as_str()).map(|url| {
                     url.replace(
-                        "eigeninference-bundle-macos-arm64.tar.gz",
-                        "eigeninference-python-macos-arm64.tar.gz",
+                        "darkbloom-bundle-macos-arm64.tar.gz",
```
hankbobtheresearchoor, line 934: 🟠 **R2 artifact path**: `darkbloom-bundle-macos-arm64.tar.gz` won't exist for previously published releases (which use `eigeninference-bundle-*`). Self-update on an older provider version will 404.

Consider: If the download URL returns 404, fall back to the `eigeninference-bundle-*` path. Or version-gate the URL construction.

hankbobtheresearchoor (changes requested):
## Backwards Compatibility Review — EigenInference → Darkbloom Rename

### ✅ Well-Handled (backwards compat present)
- **GCP Secret Manager**: `fetch`/`fetch_secret` try `darkbloom-*` first, fall back to `eigeninference-*` — good
- **ACME provisioner**: Adds `eigeninference-acme` as alias alongside `darkbloom-acme` — good
- **Rust config dir**: Checks `~/.config/darkbloom/` first, falls back to `~/.config/eigeninference/`, auto-copies — good
- **Install.sh auth_token**: Checks both paths — good
- **Sealed Content-Type**: Coordinator accepts both `application/darkbloom-sealed+json` and legacy `application/eigeninference-sealed+json` — good
- **ENV vars**: `EIGENINFERENCE_*` env vars intentionally preserved — good
- **Console UI localStorage**: `eigeninference_* → darkbloom_*` migration map preserved — good
- **Legacy key cleanup**: `LEGACY_KEY_FILES` and `legacy_secret_paths` still include `.eigeninference` paths — good

### 🔴 Still Breaking — Must Fix Before Merge

1. **PKCS#12 push cert password** — Hardcoded `pass:eigeninference` → `pass:darkbloom`. Existing push certs encrypted with old password can't be decrypted. Must try old password first.
2. **MDM API key default** — Fallback changed from `eigeninference-micromdm-api` → `darkbloom-micromdm-api`. Existing MicroMDM deployments using old API key break if env var not set.
3. **Step-CA password file** — `eigeninference-step-ca` → `darkbloom-step-ca`. Existing CA with old password can't be unlocked on restart.
4. **Step-CA provisioner name** — `eigeninference-admin` → `darkbloom-admin`. Existing client cert signing requests using old provisioner name will fail.
5. **Swift ConfigManager legacy path removed** — Migration bridge from `eigeninference/` to `darkbloom/` path deleted. Upgrading macOS app users lose existing config.

### 🟠 Should Fix — Will Break Some Deployments

6. **R2 release artifact paths** — `eigeninloom-bundle-*` → `darkbloom-bundle-*`. Self-update will 404 on previously published releases. Needs fallback URL or version-aware path construction.
7. **Enclave binary name** — `eigeninference-enclave` → `darkbloom-enclave`. Existing installations with `eigeninference-enclave` in `~/.darkbloom/bin/` have a stale binary. Should check both names.
8. **Go module path** — `github.com/eigeninference/coordinator` → `github.com/darkbloom/coordinator`. All downstream Go imports break. Acceptable for internal monorepo, but any external consumers will need a major version bump.

### 🟡 Low Risk / Acceptable

9. **Python SDK import** — `from eigeninference import EigenInference` → `from darkbloom import Darkbloom`. Internal test file only, no published PyPI package to break.
10. **Install.sh RC cleanup** — Removes `.eigeninference/bin` from PATH cleanup patterns. Existing providers with `eigeninference` in PATH won't get cleaned up on reinstall.

**Verdict: Changes Requested** — 5 items in the 🔴 category will brick existing deployments and must be addressed.

coordinator/deploy/start.sh
```diff
@@ -45,8 +65,24 @@ if [ ! -d "/data/step-ca/config" ]; then
...
+        '(.authority.provisioners[] | select(.type == "ACME")) |= $acme |
+         .authority.provisioners += [$legacy]' \
+        "$CA_JSON" > /tmp/ca.json && mv /tmp/ca.json "$CA_JSON"
+    echo "ACME provisioner configured (darkbloom-acme + eigeninference-acme alias)."
+
+    # Add eigeninference-admin as a JWK provisioner alias sharing the same key
+    # as darkbloom-admin, so existing clients using the old provisioner name
+    # can still sign certificate requests.
+    echo "Adding eigeninference-admin JWK provisioner alias..."
+    ADMIN_KEY=$(jq -r '.authority.provisioners[] | select(.type=="JWK" and .name=="darkbloom-admin") | .key' "$CA_JSON")
```
hankbobtheresearchoor, line 79: 💡 **Nit**: The JWK alias provisioner is only added during first-boot initialization (`if [ ! -d "/data/step-ca/config" ]`). For existing VMs that already have step-ca initialized, the `eigeninference-admin` alias won't be added. Consider adding a migration step (similar to the password migration above) that appends the alias to an existing `ca.json` on subsequent boots.

Non-blocking — existing VMs with the old provisioner will continue to work since the old provisioner name is already in their `ca.json`. This only matters if step-ca is re-initialized.

coordinator/cmd/coordinator/main.go
```diff
@@ -413,7 +419,7 @@ func main() {
 	if mdmURL := os.Getenv("EIGENINFERENCE_MDM_URL"); mdmURL != "" {
 		mdmKey := os.Getenv("EIGENINFERENCE_MDM_API_KEY")
 		if mdmKey == "" {
-			mdmKey = "eigeninference-micromdm-api" // default
+			mdmKey = "eigeninference-micromdm-api" // legacy default
```
hankbobtheresearchoor, line 422: 💡 **Note**: Keeping `eigeninference-micromdm-api` as the Go-side default is a pragmatic choice — it means existing deployments keep working without env var changes. The new `darkbloom-micromdm-api` default only kicks in via the deploy script's nested fallback. This works but means the two codepaths have different defaults. Consider a code comment explaining this split for future maintainers.

hankbobtheresearchoor (approved):
## Re-Review After Fixes — EigenInference → Darkbloom Rename

### ✅ All 5 Critical Items Resolved

1. **PKCS#12 push cert password** — ✅ Now tries `pass:darkbloom` first, falls back to `pass:eigeninference` via `||`. Both cert and key extraction covered.
2. **MDM API key default** — ✅ Kept as `eigeninference-micromdm-api` (legacy default) in `main.go`. Deploy scripts use `${MICROMDM_API_KEY:-${MICROMDM_API_KEY_FALLBACK:-darkbloom-micromdm-api}}` for new installs.
3. **Step-CA password** — ✅ Adds migration block that detects `eigeninference-step-ca` in password file and rewrites to `darkbloom-step-ca`. New installs get `darkbloom-step-ca` directly.
4. **Step-CA JWK provisioner alias** — ✅ Adds `eigeninference-admin` as a JWK provisioner alias sharing the same key as `darkbloom-admin`. Matches the ACME alias pattern.
5. **Swift ConfigManager legacy path** — ✅ Restored the `eigeninference/` fallback check. Checks `darkbloom/` first, then `eigeninference/`, returns new path if neither exists.

Great fixes — all follow the established pattern of new-name-first, old-name-fallback.

### 🟠 Remaining Items (non-blocking but should be tracked)

6. **R2 artifact path** — Still `darkbloom-bundle-*` only. Self-update will 404 on previously published releases using `eigeninference-bundle-*`. Low urgency since old releases can be re-uploaded or redirected at the CDN level.
7. **Enclave binary name** — `cmd_doctor` only checks `darkbloom-enclave`, doesn't fall back to `eigeninference-enclave`. If a user has the old binary but not the new one, doctor reports it missing. Low risk since `cmd_update` installs the new binary.
8. **Go module path** — `github.com/darkbloom/coordinator` is the new path. Acceptable for internal monorepo; just needs awareness that any external consumers break.

These are minor and can be addressed in follow-up PRs.

**Verdict: Approve** — All critical backwards compatibility regressions are resolved.

PR #104, opened 2026-04-30
Clarify provider trust diagnostics

Gajesh2007:
## Summary
- move the local-only provider trust diagnostics changes onto a clean branch from current master
- improve provider trust/diagnostics behavior in provider/src/main.rs

## Verification
- cd provider && cargo fmt --check
- pre-push hook passed


hankbobtheresearchoor (changes requested):
## Hermes Agent Review

Good structural improvement — distinguishing local MDM profile presence from coordinator-side hardware trust is a real correctness fix. The `doctor --support` flag is practical. A few issues to address:

### 🔴 Critical

**1. `/v1/providers/attestation` leaks all providers' serial numbers and trust states to every caller**

The provider CLI queries this public, unauthenticated endpoint and filters client-side by `serial_number`. This means *any* caller gets every provider's `serial_number`, `provider_id`, `trust_level`, `status`, `sip_enabled`, etc. Serial numbers are PII-equivalent — they uniquely identify hardware. The coordinator should filter server-side: accept `?serial_number=X` as a query param and only return records for that serial.

**2. `cmd_status` uses `DEFAULT_COORDINATOR_HTTP_URL` instead of respecting user's configured coordinator**

Line ~4910: `fetch_coordinator_provider_trust(DEFAULT_COORDINATOR_HTTP_URL, &serial)` — the `status` subcommand doesn't accept a `--coordinator` flag, so if a provider is running against staging or a local coordinator, `darkbloom status` will query prod and show wrong trust info. Either add `--coordinator` to `Status`, or read it from the existing config file.

### ⚠️ Warnings

**3. `build.rs` still emits `DARKBLOOM_ENROLL_PROFILE_URL` but it's no longer consumed**

`build.rs:39` still does `println!("cargo:rustc-env=DARKBLOOM_ENROLL_PROFILE_URL={enroll_url}")` but `main.rs` no longer has the `option_env!("DARKBLOOM_ENROLL_PROFILE_URL")` that reads it. This is dead code in the build script — should be cleaned up or it'll confuse future maintainers.

**4. `CoordinatorProviderTrust` uses `serde(default)` on bool fields — silent wrong defaults**

All the `bool` fields (`mdm_verified`, `acme_verified`, `mda_verified`, `sip_enabled`, etc.) use `#[serde(default)]` which defaults to `false`. If the coordinator response is missing or restructures any field, the provider will silently interpret "missing" as "not verified" — which is the *safe* direction, but could mask API breakage. Consider using `Option<bool>` instead so a deserialization mismatch surfaces as `None` rather than a silent `false`.

**5. `already_enrolled` branch in `cmd_install` skips serial-based profile entirely**

When MDM is already present locally, the code just prints "Local MDM profile present" and moves on. But the old profile might be the legacy static one (pre-ACME). The user should be nudged to re-enroll to get the ACME `device-attest-01` payload. A one-liner like `"Consider re-enrolling with darkbloom enroll to update your attestation profile."` would help.

### 💡 Suggestions

**6. `prefer_provider_record` sorts by `provider_id` as tertiary key — but provider IDs are UUIDs**

Sorting by UUID string comparison gives no meaningful ordering. Consider sorting by most-recent registration time or just taking the first hardware-verified online record without sorting the rest.

**7. `support_command_output` truncates to 12 lines silently**

For `profiles status -type enrollment`, 12 lines is probably fine, but for future commands added to `--support`, this could silently drop important output. Consider adding an indication like `(truncated, X more lines)` when output is cut.

**8. `total_checks` is hardcoded to 9 but computed from a mutable counter**

If someone adds/removes a check, `total_checks` won't auto-update. Consider deriving it from the number of `passed + issues.len()` at the end, or at least adding a comment.

### ✅ Looks Good

- Dynamic enrollment via `POST /v1/enroll` with serial — much better than static `.mobileconfig` download
- `profile_url` → `Option<String>` with `#[arg(hide = true)]` — clean backwards compat path
- `CoordinatorProviderTrust` struct with `is_hardware_verified()` / `is_online()` helpers — good encapsulation
- Doctor check #9 (coordinator hardware trust) — fills a real gap
- Test for `prefer_provider_record` sort ordering
- Correct phrasing change: "MDM enrolled → hardware trust" → "Local MDM profile present"

provider/src/main.rs
```diff
@@ -405,6 +447,47 @@ fn coordinator_http_base(coordinator_url: &str) -> String {
...
+async fn fetch_coordinator_provider_trust(
+    coordinator_url: &str,
+    serial_number: &str,
+) -> Result<Vec<CoordinatorProviderTrust>> {
+    let base_url = coordinator_http_base(coordinator_url);
+    let client = reqwest::Client::builder()
+        .timeout(std::time::Duration::from_secs(5))
+        .build()?;
+    let resp = client
+        .get(format!("{base_url}/v1/providers/attestation"))
```
hankbobtheresearchoor, line 469: 🔴 **Critical: This endpoint leaks all providers' attestation data to every caller**

The client fetches the *entire* provider list from `/v1/providers/attestation` and filters client-side by `serial_number`. This means any unauthenticated caller gets every provider's `serial_number`, `provider_id`, `trust_level`, `status`, `sip_enabled`, etc.

Serial numbers are PII-equivalent — they uniquely identify hardware and can't be rotated.

**Fix:** Add a query param `?serial_number=X` on the server side and only return records matching that serial. The endpoint should not expose other providers' data.

provider/src/main.rs
```diff
@@ -1845,42 +1935,65 @@ async fn cmd_install(
...
-    // Step 3: MDM enrollment (skip if already enrolled)
+    // Step 3: MDM enrollment profile. Local profile presence is not the same
+    // as coordinator hardware trust; doctor checks the network-side state.
     println!("Step 3/6: MDM enrollment...");
 
     let already_enrolled = security::check_mdm_enrolled();
 
     if already_enrolled {
-        println!("  ✓ Already enrolled in MDM — skipping");
+        println!("  ✓ Local MDM profile present");
```
hankbobtheresearchoor, line 1945: ⚠️ **Warning: `already_enrolled` branch skips re-enrollment nudge**

When MDM is already present locally, the code just prints "Local MDM profile present" and moves on. But the old profile might be the legacy static one (pre-ACME). The user should be nudged to re-enroll to get the ACME `device-attest-01` payload.

**Suggestion:** Add a one-liner like:
```rust
println!("  Consider re-enrolling with `darkbloom enroll` to update your attestation profile.");
```

provider/src/main.rs
```diff
@@ -405,6 +447,47 @@ fn coordinator_http_base(coordinator_url: &str) -> String {
...
 }
 
+fn prefer_provider_record(
+    a: &CoordinatorProviderTrust,
+    b: &CoordinatorProviderTrust,
+) -> std::cmp::Ordering {
+    b.is_hardware_verified()
+        .cmp(&a.is_hardware_verified())
+        .then_with(|| b.is_online().cmp(&a.is_online()))
+        .then_with(|| a.provider_id.cmp(&b.provider_id))
```
hankbobtheresearchoor, line 457: 💡 **Suggestion: Sorting by UUID string comparison gives no meaningful ordering**

`provider_id` values are UUIDs, so `a.provider_id.cmp(&b.provider_id)` is essentially random. Consider sorting by most-recent registration time, or just taking the first hardware-verified online record without sorting the rest.

provider/src/main.rs
```diff
@@ -5121,12 +5290,90 @@ async fn cmd_earnings(coordinator_url: String) -> Result<()> {
...
+            let combined = format!(
+                "{}{}",
+                String::from_utf8_lossy(&output.stdout),
+                String::from_utf8_lossy(&output.stderr)
+            );
+            let trimmed = combined.trim();
+            if trimmed.is_empty() {
+                "<no output>".to_string()
+            } else {
+                trimmed.lines().take(12).collect::<Vec<_>>().join("\n    ")
```
hankbobtheresearchoor, line 5306: 💡 **Suggestion: `take(12)` truncates silently**

For `profiles status -type enrollment`, 12 lines is probably fine, but for future commands added to `--support`, this could silently drop important output. Consider adding an indication like `(truncated, X more lines)` when output is cut.

provider/src/main.rs
```diff
@@ -5121,12 +5290,90 @@ async fn cmd_earnings(coordinator_url: String) -> Result<()> {
...
+    }
+}
+
+async fn cmd_doctor(coordinator_url: String, support: bool) -> Result<()> {
     println!("Darkbloom Doctor — System Diagnostics");
     println!();
 
     let mut issues: Vec<String> = Vec::new();
     let mut passed = 0;
+    let total_checks = 9;
```
hankbobtheresearchoor, line 5373: 💡 **Suggestion: `total_checks` is hardcoded to 9**

If someone adds/removes a check, `total_checks` won't auto-update. Consider deriving it from `passed + issues.len()` at the end, or at least adding a comment like `// Update this when adding/removing checks`.

hankbobtheresearchoor (changes requested):

provider/src/main.rs
```diff
@@ -405,6 +447,47 @@ fn coordinator_http_base(coordinator_url: &str) -> String {
...
+    coordinator_url: &str,
+    serial_number: &str,
+) -> Result<Vec<CoordinatorProviderTrust>> {
+    let base_url = coordinator_http_base(coordinator_url);
+    let client = reqwest::Client::builder()
+        .timeout(std::time::Duration::from_secs(5))
+        .build()?;
+    let resp = client
+        .get(format!("{base_url}/v1/providers/attestation"))
+        .send()
```
hankbobtheresearchoor, line 470: 🔴 **Critical: This endpoint leaks all providers' attestation data to every caller**

The client fetches the *entire* provider list from `/v1/providers/attestation` and filters client-side by `serial_number`. This means any unauthenticated caller gets every provider's `serial_number`, `provider_id`, `trust_level`, `status`, `sip_enabled`, etc.

Serial numbers are PII-equivalent — they uniquely identify hardware and can't be rotated.

**Fix:** Add a query param `?serial_number=X` on the server side and only return records matching that serial.

provider/src/main.rs
```diff
@@ -4789,6 +4902,62 @@ async fn cmd_status() -> Result<()> {
     );
     println!();
 
+    // Coordinator trust is the network-side source of truth for routing.
+    println!("  Coordinator trust:");
+    match get_serial_number() {
+        Ok(serial) => {
+            println!("    Serial:    {serial}");
+            match fetch_coordinator_provider_trust(DEFAULT_COORDINATOR_HTTP_URL, &serial).await {
+                Ok(records) if records.is_empty() => {
```
hankbobtheresearchoor, line 4911: 🔴 **Critical: `cmd_status` hardcodes `DEFAULT_COORDINATOR_HTTP_URL`**

If a provider is running against staging or a local coordinator, `darkbloom status` will query prod and show wrong trust info. The `Status` subcommand doesn't accept a `--coordinator` flag.

**Fix:** Add `--coordinator` to `Status` (like `Doctor` has), or read it from the existing config file.

provider/src/main.rs
```diff
@@ -96,6 +92,52 @@ struct CatalogModel {
...
+#[derive(Debug, Clone, serde::Deserialize)]
+struct CoordinatorProviderTrust {
+    #[serde(default)]
+    provider_id: String,
+    #[serde(default)]
+    serial_number: String,
+    #[serde(default)]
+    trust_level: String,
+    #[serde(default)]
+    status: String,
```
hankbobtheresearchoor, line 110: ⚠️ **Warning: `serde(default)` on `bool` fields gives silent wrong defaults**

`#[serde(default)]` defaults bools to `false`. If the coordinator response is missing a field, the provider silently interprets "missing" as "not verified" — safe direction, but masks API breakage.

**Consider:** `Option<bool>` so a deserialization mismatch surfaces as `None` rather than a silent `false`.

provider/src/main.rs
```diff
@@ -1845,42 +1935,65 @@ async fn cmd_install(
...
+    // Step 3: MDM enrollment profile. Local profile presence is not the same
+    // as coordinator hardware trust; doctor checks the network-side state.
     println!("Step 3/6: MDM enrollment...");
 
     let already_enrolled = security::check_mdm_enrolled();
 
     if already_enrolled {
-        println!("  ✓ Already enrolled in MDM — skipping");
+        println!("  ✓ Local MDM profile present");
+        println!("    Coordinator hardware trust will be verified after provider registration.");
```
hankbobtheresearchoor, line 1946: ⚠️ **Warning: `already_enrolled` branch skips re-enrollment nudge**

When MDM is already present locally, the code just prints "Local MDM profile present". But the old profile might be the legacy static one (pre-ACME). The user should be nudged to re-enroll to get the ACME `device-attest-01` payload.

```rust
println!("  Consider re-enrolling with `darkbloom enroll` to update your attestation profile.");
```

provider/src/main.rs
```diff
@@ -405,6 +447,47 @@ fn coordinator_http_base(coordinator_url: &str) -> String {
...
 
+fn prefer_provider_record(
+    a: &CoordinatorProviderTrust,
+    b: &CoordinatorProviderTrust,
+) -> std::cmp::Ordering {
+    b.is_hardware_verified()
+        .cmp(&a.is_hardware_verified())
+        .then_with(|| b.is_online().cmp(&a.is_online()))
+        .then_with(|| a.provider_id.cmp(&b.provider_id))
+}
```
hankbobtheresearchoor, line 458: 💡 **Suggestion: Sorting by UUID string comparison gives no meaningful ordering**

`provider_id` values are UUIDs, so `a.provider_id.cmp(&b.provider_id)` is essentially random. Consider sorting by most-recent registration time, or just taking the first hardware-verified online record.

hankbobtheresearchoor (changes requested):

hankbobtheresearchoor:
⚠️ **Warning (out-of-diff): `build.rs` still emits `DARKBLOOM_ENROLL_PROFILE_URL` but it's no longer consumed**

`provider/build.rs:39` still does `println!("cargo:rustc-env=DARKBLOOM_ENROLL_PROFILE_URL={enroll_url}")` but `main.rs` no longer has the `option_env!("DARKBLOOM_ENROLL_PROFILE_URL")` that reads it. This is dead code in the build script — should be cleaned up or it'll confuse future maintainers.

provider/src/main.rs
ethenotethan: eh this isn't really leaking since the endpoint is publicly accessible. the riskiest field is the serial number which is embedded as an OIN in the x509 cert signed by apple attesting to the device's identity and the p256 key that's generated within the enclave. chatted with @anupsv who also agreed its a bit risky to expose publicly.

provider/src/main.rs
ethenotethan: @claude how frequent are queries being made to the coordinator server by the inference provider on `/v1/providers/attestation` ?

provider/src/main.rs
ethenotethan: this is more or less dynamic resolution and doesn't ensure adequate compatibility if this data type were to get updated by the coordinator in the future.

provider/src/main.rs
ethenotethan: got it, So: 
```
Only when a user manually runs status or doctor
```

In that case this is a good callout but is better suited as a follow up issue (not blocking merge).

ethenotethan: hmm still curious about what the request lifecycle looks like for the `/attestation` endpoint and if there could be DoS... not an issue for this PR though

provider/src/main.rs
ethenotethan: @hankbobtheresearchoor I agree with your deduction but disagree with `--coordinator` being the proper user entrypoint, perhaps something like `dev` and `prod` distinctions would make the most sense for user comprehensibility.

ethenotethan: actually second guessing myself, for spinning up the network in a local env being able to pass endpoint is critical. probably makes sense to:
1. default to prod coordinator if no coordinator URL is provided to binary env

PR #107, opened 2026-05-01
feat(provider): wire BatchedEngine for continuous batching in-process

hankbobtheresearchoor:
## Summary

Replace `AdaptiveEngine(SimpleEngine)` with `BatchedEngine(EngineCore+Scheduler)` so concurrent inference requests share forward passes via iteration-level continuous batching.

### Problem

`SharedEngine` currently serializes every request through a single blocking Python thread. Each `engine.chat()` / `engine.stream_chat()` call blocks to completion before the next request is dequeued from `cmd_rx`. `AdaptiveEngine` was chosen because its docstring says batching overhead outweighs gains for 27B+ models at low concurrency — but this assumes single-user workloads, not a distributed inference network where multiple consumers hit the same provider concurrently.

### Solution

Wire `BatchedEngine` into the in-process path. `BatchedEngine` wraps `AsyncEngineCore` which runs a `_engine_loop()` that continuously steps the `Scheduler`. When multiple `generate()`/`stream_generate()` calls arrive concurrently, they all submit to the same scheduler and get interleaved in the same forward pass.

### Changes

| Path | Change |
|------|--------|
| `load_vllm_mlx()` | Create `BatchedEngine` with `SchedulerConfig(max_num_seqs=16, prefill_batch=4, completion_batch=16)`. Start persistent asyncio event loop on daemon thread. Schedule `_engine.start()` on it (blocks until model loaded). Store both engine + loop in builtins. |
| `generate_via_server_handler()` | `asyncio.run(engine.chat(...))` → `asyncio.run_coroutine_threadsafe(engine.chat(...), _loop).result()`. Schedules on the persistent loop where `_engine_loop()` runs. |
| `stream_generate()` | `SyncStreamIterator` receives persistent loop instead of creating `new_event_loop()`. `__next__()` uses `run_coroutine_threadsafe(__anext__(), loop).result()` since the loop is already running `run_forever()`. |
| `unload_vllm_mlx()` | `engine.stop()` via `run_coroutine_threadsafe`, then `loop.call_soon_threadsafe(loop.stop)` to shut down daemon thread. |

### Architecture

```
Before (serial):
  Consumer A ──→ cmd_rx ──→ engine.chat() ──► blocks to completion ──► reply
  Consumer B ──→ cmd_rx ──┘ (waiting)                ↑ next dequeue

After (continuous batching):
  Consumer A ──→ cmd_rx ──→ run_coroutine_threadsafe ──→ _loop ─┐
  Consumer B ──→ cmd_rx ──→ run_coroutine_threadsafe ──→ _loop ─┤
                                                                 ▼
                                                    EngineCore._engine_loop()
                                                    Scheduler.step() interleaves
                                                    A.prefill → B.prefill → A.decode → ...
```

### Why not AdaptiveEngine?

`AdaptiveEngine` wraps `SimpleEngine` with a FIFO queue + `asyncio.Lock`. It still processes one request at a time — the queue just means the caller doesn't block on the GIL. `BatchedEngine` uses `EngineCore` which has true iteration-level scheduling: prefill and decode phases from different requests are interleaved in the same step.

### Tuning

`max_num_seqs=16` is conservative for 24-48GB Apple Silicon machines. On the 256GB Mac Studio this can safely go to 32-64. `prefill_batch_size=4` and `completion_batch_size=16` match the vllm-mlx defaults for continuous batching.

### DIFF
--- a/provider/src/inference.rs
+++ b/provider/src/inference.rs
@@ -323,11 +323,17 @@ for _name in (
         Ok(())
     }
 
-    /// Initialize the vllm-mlx engine via its server module's `load_model()`.
+    /// Initialize the vllm-mlx engine with continuous batching.
     ///
-    /// This creates an AdaptiveEngine (SimpleEngine + request queuing) and
-    /// stores it in a Python builtins dict keyed by cache_key. The engine
-    /// supports `engine.chat()` and `engine.stream_chat()` with full
+    /// This creates a BatchedEngine (wrapping EngineCore + Scheduler) which
+    /// supports true continuous batching: multiple concurrent requests share
+    /// the same forward pass via iteration-level scheduling.  An asyncio
+    /// event loop is started on a background daemon thread so the engine's
+    /// `_engine_loop()` can keep processing steps while the Rust caller
+    /// thread submits requests.
+    ///
+    /// The engine is stored in a Python builtins dict keyed by cache_key.
+    /// It supports `engine.chat()` and `engine.stream_chat()` with full
     /// OpenAI-compatible features (chat templates, tool calling, structured
     /// output) without starting an HTTP server.
     fn load_vllm_mlx(&self, py: Python<'_>) -> Result<()> {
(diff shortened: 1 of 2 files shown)


ethenotethan:
closing since we're deprioritizing for swift-llm integration

PR #123, opened 2026-05-04
security: call block_dangerous_modules during engine load

antojoseph:
## Summary

- `block_dangerous_modules()` was defined in `provider/src/inference.rs` and covered by tests, but was never called in the production `load()` path
- `socket`, `subprocess`, `ctypes`, `multiprocessing`, and `faulthandler` were importable by provider-controlled `vllm_mlx` code at inference time despite `dangerous_modules_blocked: true` being reported to the coordinator
- Fix: call both `lock_python_path` and `block_dangerous_modules` at the top of `load_vllm_mlx`, in the same GIL scope that runs the model load

## Impact

A malicious provider serving private text could import `socket` or `subprocess` to exfiltrate prompts or establish outbound connections from inside the hardened process.

## Test plan

- [ ] Existing `test_block_dangerous_modules_blocks_imports` still passes
- [ ] `cargo test --features python` passes
- [ ] Verify a provider can load a model successfully (no regression in happy path)

### DIFF
--- a/provider/src/inference.rs
+++ b/provider/src/inference.rs
@@ -331,6 +331,13 @@ for _name in (
     /// OpenAI-compatible features (chat templates, tool calling, structured
     /// output) without starting an HTTP server.
     fn load_vllm_mlx(&self, py: Python<'_>) -> Result<()> {
+        // Enforce both security layers in the same GIL scope that runs vllm_mlx.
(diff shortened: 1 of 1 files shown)


ethenotethan:
rust backend is actually getting nuked really soon in-favor of a native swift provider

PR #124, opened 2026-05-04
fix: deduplicate Stripe webhook credits via ledger reference index

antojoseph:
## Summary
- Adds a unique partial index `idx_ledger_reference` on `ledger_entries(entry_type, reference) WHERE reference <> ''` so duplicate Stripe webhook deliveries cannot double-credit an account
- Updates `creditTx` to pre-check for an existing entry and use `ON CONFLICT DO NOTHING` on insert, returning `ErrAlreadyCredited` on duplicate

## Security impact
Without this fix a Stripe `checkout.session.completed` webhook replayed by Stripe (or by an attacker replaying the raw POST) would credit the same account multiple times. The index makes the operation idempotent at the database level.

## Test plan
- [ ] Run coordinator unit tests: `go test ./internal/store/...`
- [ ] Manually trigger duplicate webhook delivery in Stripe test dashboard and verify balance credited only once

### DIFF
--- a/coordinator/internal/store/postgres.go
+++ b/coordinator/internal/store/postgres.go
@@ -195,6 +195,7 @@ func (s *PostgresStore) migrate(ctx context.Context) error {
 			created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
 		)`,
 		`CREATE INDEX IF NOT EXISTS idx_ledger_account ON ledger_entries(account_id, created_at DESC)`,
+		`CREATE UNIQUE INDEX IF NOT EXISTS idx_ledger_reference ON ledger_entries(entry_type, reference) WHERE reference <> ''`,
 
 		// Referral system tables
 		`CREATE TABLE IF NOT EXISTS referrers (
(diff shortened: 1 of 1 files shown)


coordinator/internal/store/postgres.go
```diff
@@ -916,6 +943,9 @@ func (s *PostgresStore) Credit(accountID string, amountMicroUSD int64, entryType
 	defer tx.Rollback(ctx)
 
 	if err := creditTx(ctx, tx, accountID, amountMicroUSD, entryType, reference, time.Time{}); err != nil {
+		if errors.Is(err, ErrAlreadyCredited) {
+			return nil
```
hankbobtheresearchoor, line 947: 🔴 This makes *all* repeated `Credit` calls with the same `(account_id, entry_type, reference)` silently no-op, not just Stripe checkout deposits. Existing callers reuse references inside one account: reservation refunds use `"reservation_refund"`, and admin credits default to `"admin_credit"`. After this change, a user with two early-failed reserved requests only gets the first refund; the second `LedgerRefund` is swallowed here. Either scope idempotency to external deposit refs / `LedgerStripeDeposit`, or make those other credit callers pass truly unique references before enforcing this globally.

coordinator/internal/store/postgres.go
```diff
@@ -195,6 +195,7 @@ func (s *PostgresStore) migrate(ctx context.Context) error {
 			created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
 		)`,
 		`CREATE INDEX IF NOT EXISTS idx_ledger_account ON ledger_entries(account_id, created_at DESC)`,
+		`CREATE UNIQUE INDEX IF NOT EXISTS idx_ledger_reference ON ledger_entries(account_id, entry_type, reference) WHERE reference <> ''`,
```
hankbobtheresearchoor, line 199: 🟡 This startup migration will fail if production already has duplicate non-empty `(account_id, entry_type, reference)` rows. Given existing reused references like `reservation_refund` / default `admin_credit`, that seems plausible. Add a preflight/dedupe migration or narrow the unique index predicate to the Stripe deposit reference shape so deploys do not fail closed on existing ledger history.

coordinator/internal/store/postgres.go
```diff
@@ -837,7 +838,29 @@ func nullableCreatedAt(ts time.Time) any {
 	return ts
 }
 
+// ErrAlreadyCredited is returned by creditTx when the (account_id, entry_type,
+// reference) tuple already exists, indicating an idempotent duplicate.
+var ErrAlreadyCredited = errors.New("store: credit already applied for this reference")
+
```
hankbobtheresearchoor, line 844: 🟡 The Store contract now differs between implementations: `PostgresStore.Credit` is idempotent for duplicate references, but `MemoryStore.Credit` still applies every duplicate. Most billing/API tests use memory, so the Stripe replay invariant and the legitimate repeated-refund/admin-credit cases are not covered. Please add focused tests for duplicate `LedgerStripeDeposit` and for non-Stripe repeated credit refs, ideally exercising Postgres semantics.

hankbobtheresearchoor (changes requested):

PR #129, opened 2026-05-04
fix: reject stale and future-dated attestation blobs at provider registration

antojoseph:
## Summary
- Calls `attestation.CheckTimestamp` with a 5-minute max-age window immediately after `VerifyJSON` succeeds at provider registration
- A replayed or future-dated attestation blob causes `MarkUntrusted` and an early return — no trust level is granted

## Security impact
Without this fix, an attacker who obtained a valid attestation blob (even legitimately, from a machine they control) could replay it indefinitely to register fake provider identities with `self_signed` or higher trust. The 5-minute window matches the existing challenge cycle interval, so legitimate providers re-attesting normally are unaffected.

## Test plan
- [ ] `go test ./internal/api/... ./internal/attestation/...`
- [ ] Send a registration with an attestation blob timestamped > 5 min ago and verify the provider is marked untrusted
- [ ] Send a registration with a future-dated timestamp and verify same rejection

### DIFF
--- a/coordinator/internal/api/provider.go
+++ b/coordinator/internal/api/provider.go
@@ -1208,6 +1208,22 @@ func (s *Server) verifyProviderAttestation(providerID string, provider *registry
 		return
 	}
 
+	// Reject stale attestation blobs. A replayed blob from a previously-valid
+	// provider could otherwise grant trust to a new connection without fresh
+	// hardware verification. 5 minutes matches the challenge cycle interval.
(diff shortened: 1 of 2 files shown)


hankbobtheresearchoor (approved):
LGTM. This adds the missing freshness gate before any provider trust is granted, and CheckTimestamp now explicitly rejects future-dated attestations instead of treating negative age as valid.

PR #130, opened 2026-05-04
fix: verify all adjacent certificate chain links in browser MDA validation

antojoseph:
## Summary
- `console-ui/src/lib/cert-verify.ts`: step 3 now iterates all adjacent pairs `certs[i].verify(certs[i+1])` instead of only checking `leaf → certs[1]`
- Step label updated to "Verifying certificate chain links"

## Security impact
The previous code only verified that the leaf was signed by `certs[1]`, and separately that the topmost cert was signed by Apple's root. A crafted chain like `[forged_leaf, attacker_CA, Apple_root]` — where `forged_leaf` is signed by `attacker_CA` and `attacker_CA` is self-signed — would pass both checks without `attacker_CA` being validated against `Apple_root`. The fix closes this gap by verifying every link in the chain before accepting it.

This is a client-side verification displayed in the UI's "Verify Device" panel, so exploitation requires serving a forged cert chain to a user who clicks "Verify."

## Test plan
- [ ] `npx tsc --noEmit` — no new type errors in cert-verify.ts
- [ ] Manual: paste a valid 3-cert Apple MDA chain and verify all steps pass
- [ ] Manual: insert a self-signed intermediate and verify step 3 fails


hankbobtheresearchoor (approved):
LGTM. This closes the browser-side MDA chain gap by checking each adjacent certificate signature before the Apple root check.

PR #110, opened 2026-05-01
Add Swift provider runtime

Gajesh2007:
## Summary

Adds the Swift-native provider package under `provider-swift/`, wired to local MLX submodules and Hummingbird for the standalone OpenAI-compatible endpoint surface.

This also updates the coordinator trust path for the Swift provider cutover:

- accepts the canonical `mlx-swift` backend for private text routing
- makes Swift runtime verification backend-aware instead of requiring Python/vllm runtime hashes
- makes binary hash policy fail closed when configured
- requires challenge binary hashes to match the registration attestation hash
- preserves RDMA status reporting without treating RDMA enabled as an automatic hypervisor failure

The PR includes new focused tests for Swift runtime eligibility and binary-hash registration/challenge enforcement.

## Submodule note

This PR adds:

- `libs/mlx-swift` at `49d73ab`
- `libs/mlx-swift-lm` at `3ec4b8a`

The `mlx-swift-lm` commit is on `Gajesh2007/mlx-swift-lm:codex/local-mlx-swift-dependency` and changes its `Package.swift` to use the sibling local `../mlx-swift` submodule. That is needed so the parent repo can resolve both MLX packages locally during SwiftPM builds.

## Validation

- `swift build --package-path provider-swift`
- `swift test --package-path provider-swift`
- `go test ./internal/api`
- `go test ./internal/registry`
- `git diff --cached --check`

PR #131, opened 2026-05-04
fix: fatal codesign check and hash verification for Python runtime tarballs

antojoseph:
## Summary
- `scripts/install.sh` and `coordinator/internal/api/install.sh` (its embedded copy):
  1. `codesign --verify` failure is now fatal (`exit 1` + cleanup) — an unsigned binary is never installed or added to PATH
  2. `python_hash` and `runtime_hash` are extracted from the release JSON alongside `bundle_hash`
  3. All three download paths that fetch Python runtime / site-packages tarballs from R2 now verify `shasum -a 256` against the manifest hash before extraction; mismatch aborts with `exit 1`

## Security impact
- **Codesign**: a MITM or compromised CDN could serve a tampered bundle that passes the existing SHA-256 hash check (if they also control the coordinator response) but would not be Apple-signed. Failing closed prevents silent installation of unsigned code.
- **Runtime tarballs**: the Python runtime and site-packages were downloaded from R2 and extracted with no integrity check. A compromised R2 bucket or MITM could replace them with arbitrary Python code that runs inside the provider process. Hash verification closes this gap.

When `python_hash`/`runtime_hash` fields are absent from the release JSON (older releases), the hash checks are skipped gracefully — no regression for existing releases.

## Test plan
- [ ] Run `bash -n scripts/install.sh` to syntax-check
- [ ] Verify `diff scripts/install.sh coordinator/internal/api/install.sh` only shows expected template-placeholder differences
- [ ] `go build ./...` in coordinator passes (install.sh is embedded as a static file)
- [ ] Simulate a hash mismatch by corrupting a tarball and verify `exit 1`

### DIFF
--- a/coordinator/internal/api/install.sh
+++ b/coordinator/internal/api/install.sh
@@ -72,6 +72,8 @@ json_val() { echo "$1" | sed -n "s/.*\"$2\":\"\([^\"]*\)\".*/\1/p"; }
 BUNDLE_URL=$(json_val "$RELEASE_JSON" url)
 BUNDLE_HASH=$(json_val "$RELEASE_JSON" bundle_hash)
 BINARY_HASH=$(json_val "$RELEASE_JSON" binary_hash)
+PYTHON_HASH=$(json_val "$RELEASE_JSON" python_hash)
+RUNTIME_HASH=$(json_val "$RELEASE_JSON" runtime_hash)
 VERSION=$(json_val "$RELEASE_JSON" version)
 
 echo "  Version: $VERSION"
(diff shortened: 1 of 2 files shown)


hankbobtheresearchoor:
@antojoseph heads up: the provider Python backend is currently being migrated to the native Swift backend and is expected to be removed entirely within the next week.

I looked through this PR for anything we should carry forward into the Swift LLM work. The Python runtime / site-packages tarball verification paths are tied to the backend being deprecated, so I don’t think those pieces are worth merging now. The only broadly reusable idea here is making installer code-signature verification fail closed for `darkbloom`; if we still need that, it should be carried into the Swift release/installer path directly rather than hardening the soon-to-be-removed Python runtime install path.

Gajesh2007:
we're migrating off the python runtime #110

PR #133, opened 2026-05-04
Remove in-process engine; Swift LLM backend only

ethenotethan:
## Summary

- Strip all PyO3/vllm-mlx/inprocess-mlx code across provider, coordinator, Swift provider, console UI, and docs
- The only supported private-text backend is now `mlx-swift`
- Old providers registering with `backend == "inprocess-mlx"` are rejected with an upgrade-required error

### Provider (Rust)
- Delete `provider/src/inference.rs` (in-process PyO3 engine, 1251 lines)
- Delete `provider/src/backend/vllm_mlx.rs` (vllm-mlx subprocess backend, 322 lines)
- Remove `python` feature flag and `pyo3` dependency from Cargo.toml
- Remove all `#[cfg(feature = "python")]` gated blocks from main.rs
- Remove `TextBackendMode`, `SharedInprocessEngineMap`, `handle_inprocess_request`, `validate_private_text_runtime`, etc.
- Remove `RuntimeHashes`, `compute_runtime_hashes`, `isolate_python_preinit`, `purge_pycache` from security.rs
- Remove `python_hash`/`runtime_hash` from protocol messages
- Remove `text_backend_inprocess`, `python_runtime_locked`, `dangerous_modules_blocked` from `PrivacyCapabilities`
- Remove backend subprocess auto-restart (no longer spawning vllm-mlx)

### Coordinator (Go)
- Remove `BackendInprocessMLX` constant; `privateTextBackendSupported()` now only accepts `mlx-swift`
- Reject providers with `backend == "inprocess-mlx"` at registration with upgrade error
- Remove `PythonHash`/`RuntimeHash` from `RegisterMessage`, `AttestationResponseMessage`, `Provider` struct
- Remove `TextBackendInprocess`, `PythonRuntimeLocked`, `DangerousModulesBlocked` from `PrivacyCapabilities`
- Remove Python/runtime hash verification from runtime manifest and attestation flows
- Update all test fixtures to use `"mlx-swift"` backend

### Swift provider
- Remove `textBackendInprocess`, `pythonRuntimeLocked`, `dangerousModulesBlocked` from `PrivacyCapabilities`

### Console UI & Docs
- Update TrustExplainerModal, setup FAQ, AGENTS.md, CLAUDE.md, README.md, CONTRIBUTING.md, ARCHITECTURE.md

## Validation

- `cargo check` passes in `provider/`
- `go build ./...` passes in `coordinator/`
- Pre-commit hooks (gofmt, cargo fmt, eslint) pass


hankbobtheresearchoor:
Reviewed PR #133 from the Swift-only cutover/security perspective.

Findings:

🔴 Must fix — legacy Rust `darkbloom serve` now registers/advertises private text as `backend = "mlx-swift"`, but the backend spawn path was removed. Every `BackendSlot` keeps `pid == None`/`healthy == false`; when an inference request arrives the event loop returns `503 "backend ... is not running"`. If the cutover is Swift-only, this Rust `serve` path should not keep advertising/routing private text, or it should be removed/turned into an explicit handoff to `provider-swift` rather than registering models it cannot serve. (`provider/src/main.rs` around the request health check / proxy path)

🔴 Must fix — `cargo test` no longer compiles. I ran `cd provider && PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 cargo test --no-run`; the test target still references deleted `runtime_smoke_test` and `spawn_inference_backend`, and `provider/src/coordinator.rs` has a duplicated/broken `test_handle_attestation_challenge_without_public_key` with undefined `nonce`/`signature`. Binary `cargo check` is not enough for this PR because the Rust test target is red.

🟡 Should fix — public CLI/update diagnostics still depend on the removed Python/vllm runtime: `hash-runtime` remains in help but always errors, `benchmark` imports `vllm_mlx.engine.SimpleEngine`, `doctor` reports inference runtime missing unless `vllm_mlx`/`mlx_lm` imports, and update verification requires `~/.darkbloom/python/bin/python3.12`. Fresh Swift-only installs will see false failures and obsolete remediation. These should be removed or rewired to validate the Swift provider bundle, `darkbloom-enclave`, `mlx.metallib`, and model cache.

🟡 Should fix — coordinator runtime revalidation can deroute providers but cannot restore them. In `coordinator/internal/api/server.go` around `revalidateConnectedProvidersAgainstRuntimePolicy`, if a live provider is marked `RuntimeManifestChecked=false` after a bad/stale manifest and the manifest is later fixed, the `runtimeOK == true` case does not set the fields back to true, so the provider remains unroutable until reconnect/challenge. Registration/challenge set both fields from `runtimeOK`; revalidation should do the same.

🔴 Must fix — for the Swift-only cutover, the challenge path still keeps the old status-signature rollout downgrade window open. Missing `status_signature` only logs, while positive `sip_enabled`/template-hash status can still feed `ChallengeVerifiedSIP`/runtime routing gates. With an attested SE key but compromised userspace, a provider can sign only nonce+timestamp and lie in unsigned status fields. Since this PR bumps the provider version and rejects the retired backend, missing `status_signature` should be a hard challenge failure before granting private-text routing trust.

Validation run:
- `go test ./internal/protocol ./internal/registry -count=1` ✅
- targeted `go test ./internal/api -run 'RuntimeManifest|ProviderIdentity|KnownBinaryHash|BinaryHashPolicy|ChallengeResponseRejects|SwiftProviderRequiresRuntimeManifestCheck|ProviderWithoutManifestCheck' -count=1 -timeout=90s` ✅
- `cd provider && PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 cargo test --no-run` ❌ test target does not compile
- `cd provider-swift && swift test` could not run until submodules are initialized (`libs/mlx-swift*/Package.swift` missing in this worktree)

hankbobtheresearchoor (commented):
## Review: Remove in-process engine; Swift LLM backend only

**Verdict: 🟡 Should fix before merge** — 2 must-fix items, several should-fix.

This is a clean, well-scoped cutover PR. The mechanical removal of `inprocess-mlx`, `python_hash`, `runtime_hash`, and Python-specific privacy caps is thorough and consistent across coordinator, Rust provider, Swift provider, console UI, docs, and Postgres. The registration rejection path for legacy backends is the right design.

---

### 🔴 Must Fix

**1. Duplicate banned substring in `install_sh_test.go`**
`"PBS_PYTHON_VERSION"` appears twice in `bannedSubstrings`. The original list had `"vllm-mlx"` and `"vllm_mlx"` as separate entries. This diff replaces `"vllm_mlx"` → `"PBS_PYTHON_VERSION"` but creates a duplicate since `"PBS_PYTHON_VERSION"` is already in the list one line below. Remove the duplicate.

**2. `fullstack_integration_test.go` references non-existent `mlx-swift` binary**
`exec.Command("mlx-swift", args...)` — there is no `mlx-swift` CLI binary. The Swift provider is `darkbloom`, and it doesn't accept `--host`, `--port`, `--model`, `--continuous-batching` flags the way `vllm-mlx serve` did. This test was mechanically find-replaced from `vllm-mlx` → `mlx-swift` but won't work at all. Either: (1) rewrite to use `darkbloom start --local --port N --model M`, or (2) skip with a `// TODO: rewrite for Swift backend` comment. As-is, `go test -run TestFullStack` will instant-fail.

---

### 🟡 Should Fix

**3. `runtime_hash` survives in `console-ui/src/app/providers/types.ts` but `python_hash` was removed**
The coordinator no longer populates `runtime_hash` on the provider — it was removed from both the `Provider` struct and `myProvider`. This TS field will always be `undefined`. Either remove it or document it's retained for backward compat with older coordinator responses.

**4. Hardcoded `"inprocess-mlx"` rejection in two places in `registry.go`**
`ErrBackendRejected` and the `if msg.Backend == "inprocess-mlx"` check both hardcode the string. Consider using `!privateTextBackendSupported(msg.Backend)` as the condition and including the actual backend value in the error message — that makes `privateTextBackendSupported` the single source of truth.

**5. No DB migration for dropped `python_hash`/`runtime_hash` columns**
The `releases` and `providers` tables still have these columns in production Postgres. The queries no longer reference them, which works if columns are nullable, but they should be dropped in a migration. Otherwise future devs will think they're still populated.

**6. Dead `python_cmd` plumbing in Rust `main.rs`**
The PR removes `with_runtime_hashes`/`with_runtime_hash_command` builder methods, but `main.rs` still contains `python_cmd` derivation and Python runtime validation/smoke-test code. Since the stated goal is to strip all PyO3/vllm-mlx code, the remaining `python_cmd` plumbing should be removed too — it's dead code that makes it look like the Rust provider still has a Python path.

---

### 🔵 Observations (no action needed)

**7. Registration rejection pattern is good** — old providers get a clear JSON error + `StatusPolicyViolation` close instead of a silent disconnect.

**8. `providerSupportsPrivateTextLocked` correctly drops from 8→5 caps** — `text_backend_inprocess`, `python_runtime_locked`, `dangerous_modules_blocked` are no longer meaningful. The remaining 5 (`text_proxy_disabled`, `anti_debug_enabled`, `core_dumps_disabled`, `env_scrubbed`, + `SIPEnabled` via challenge) still apply to the Swift provider.

**9. Removing `python_hash`/`runtime_hash` from signed canonical payload** slightly reduces the attack surface for a compromised provider, but `binary_hash`, `template_hashes`, `model_hashes` are still unsigned, so the core concern (status_signature only covers nonce+timestamp) remains unchanged.

coordinator/internal/api/install_sh_test.go
```diff
@@ -82,10 +82,10 @@ func TestInstallScriptTemplating(t *testing.T) {
 		body := fetchInstallScript(t, srv.URL)
 
 		// Swift cutover invariants -- install.sh must not download Python
-		// or the vllm-mlx site-packages tarball.
+		// or the legacy site-packages tarball.
 		bannedSubstrings := []string{
 			"vllm-mlx",
-			"vllm_mlx",
+			"PBS_PYTHON_VERSION",
 			"PBS_PYTHON_VERSION",
```
hankbobtheresearchoor, line 89: 🔴 **Duplicate entry** — `"PBS_PYTHON_VERSION"` appears on both line 88 and line 89.

hankbobtheresearchoor (commented):
Inline review test

hankbobtheresearchoor (commented):
## Review: Remove in-process engine; Swift LLM backend only

**Verdict: 🟡 Should fix before merge** — 2 must-fix, 4 should-fix.

Overall this is a clean, well-scoped cutover PR. The mechanical removal of `inprocess-mlx`, `python_hash`, `runtime_hash`, and Python-specific privacy caps is thorough and consistent across coordinator, Rust provider, Swift provider, console UI, docs, and Postgres. The registration rejection path for legacy backends is the right design.

See inline comments for details.

coordinator/internal/api/install_sh_test.go
hankbobtheresearchoor, line 89: 🔴 **Duplicate entry** — `"PBS_PYTHON_VERSION"` appears on both line 88 and line 89. The original list had `"vllm-mlx"` and `"vllm_mlx"` as two separate entries; this diff replaced `"vllm_mlx"` → `"PBS_PYTHON_VERSION"` but creates a duplicate since `"PBS_PYTHON_VERSION"` already appears one line above. Remove the duplicate.

coordinator/internal/api/fullstack_integration_test.go
```diff
@@ -77,19 +77,19 @@ func startBackendWithOptions(t *testing.T, model string, port int, continuousBat
 	if continuousBatching {
 		args = append(args, "--continuous-batching")
 	}
-	cmd := exec.Command("vllm-mlx", args...)
+	cmd := exec.Command("mlx-swift", args...)
```
hankbobtheresearchoor, line 80: 🔴 **Non-existent binary** — `exec.Command("mlx-swift", args...)` references a CLI that doesn't exist. The Swift provider is `darkbloom`, and it doesn't accept `serve`, `--host`, `--port`, `--model`, `--continuous-batching` flags the way `vllm-mlx serve` did. This test was mechanically find-replaced from `vllm-mlx` → `mlx-swift` but will instant-fail.

Either: (1) rewrite to use `darkbloom start --local --port N --model M`, or (2) skip the fullstack tests with a `// TODO: rewrite for Swift backend` comment.

console-ui/src/app/providers/types.ts
```diff
@@ -87,7 +87,6 @@ export interface MyProvider {
   mda_sepos_version?: string;
 
   runtime_verified: boolean;
-  python_hash?: string;
   runtime_hash?: string;
```
hankbobtheresearchoor, line 90: 🟡 **Dead field** — `runtime_hash` survives here but `python_hash` was removed. The coordinator no longer populates `runtime_hash` on the provider (removed from both the `Provider` struct and `myProvider`). This TS field will always be `undefined`. Either remove it or add a comment that it's retained for backward compat with pre-cutover coordinator responses.

coordinator/internal/registry/registry.go
```diff
@@ -54,10 +55,13 @@ const (
...
 
 const (
-	BackendInprocessMLX = "inprocess-mlx"
-	BackendMLXSwift     = "mlx-swift"
+	BackendMLXSwift = "mlx-swift"
 )
 
+// ErrBackendRejected is returned when a provider attempts to register with
+// an unsupported backend type.
+var ErrBackendRejected = fmt.Errorf("provider backend 'inprocess-mlx' is no longer supported; please upgrade to the latest provider version")
```
hankbobtheresearchoor, line 63: 🟡 **Duplicated backend string** — `"inprocess-mlx"` is hardcoded here in `ErrBackendRejected` and again on line 738 (`if msg.Backend == "inprocess-mlx"`). If another backend becomes deprecated later, this needs editing in two places. Consider defining the rejected backend as a constant, or deriving the error message from the actual value.

coordinator/internal/api/provider.go
```diff
@@ -163,7 +163,22 @@ func (s *Server) providerReadLoop(ctx context.Context, conn *websocket.Conn, pro
...
+					"error", regErr,
+				)
+				errMsg, _ := json.Marshal(map[string]string{
+					"type":  "error",
+					"error": regErr.Error(),
+				})
+				writeCtx, writeCancel := context.WithTimeout(loopCtx, 5*time.Second)
+				_ = conn.Write(writeCtx, websocket.MessageText, errMsg)
+				writeCancel()
+				conn.Close(websocket.StatusPolicyViolation, regErr.Error())
```
hankbobtheresearchoor, line 179: 🔵 Good — old providers registering with `inprocess-mlx` get a clear JSON error + `StatusPolicyViolation` close instead of a silent disconnect. The error message from `ErrBackendRejected` tells them to upgrade.

coordinator/internal/store/postgres.go
```diff
@@ -2582,9 +2582,9 @@ func (s *PostgresStore) UpdateProviderRuntime(ctx context.Context, id string, ve
 	defer cancel()
 
 	_, err := s.pool.Exec(ctx,
-		`UPDATE providers SET runtime_verified = $2, python_hash = $3, runtime_hash = $4
+		`UPDATE providers SET runtime_verified = $2
```
hankbobtheresearchoor, line 2585: 🟡 **Dead parameters + missing migration** — `pythonHash` and `runtimeHash` params (line 2580) are accepted but never used. The SQL only sets `runtime_verified = $2`. Two things:

1. Remove the dead params from this func, `MemoryStore.UpdateProviderRuntime`, and the `Store` interface (line 333).
2. Add a DB migration to drop `python_hash` and `runtime_hash` columns from both `providers` and `releases` tables. These columns still exist in production Postgres but are no longer written.

coordinator/internal/registry/registry.go
```diff
@@ -182,22 +184,18 @@ func providerSupportsPrivateTextLocked(p *Provider) bool {
 	if caps == nil {
```
hankbobtheresearchoor, line 184: 🔵 **On `providerSupportsPrivateTextLocked` (line 171):** Correctly drops from 8→5 privacy caps. `text_backend_inprocess`, `python_runtime_locked`, and `dangerous_modules_blocked` are no longer meaningful without a Python runtime. The remaining 5 are all still meaningful for the Swift provider.

hankbobtheresearchoor:
**Additional comments on lines not in the diff:**

🟡 **`coordinator/internal/store/interface.go:333`** — `UpdateProviderRuntime` still accepts `pythonHash, runtimeHash string` parameters, but neither implementation uses them. The SQL writes only `runtime_verified = $2`. These dead parameters should be removed from the interface and both implementations.

🟡 **`provider/src/main.rs:3107`** — Dead Python/vllm-mlx code. The PR's stated goal is to strip all PyO3/vllm-mlx code, and `with_runtime_hashes`/`with_runtime_hash_command` were correctly removed. But the `python_cmd` derivation + vllm-mlx availability check (lines 3106–3122) is dead code — the Rust provider no longer uses Python for inference. Same applies to `darkbloom doctor` (around line 3951) which still checks for `vllm_mlx`/`mlx_lm` imports. Leaving this is confusing: it makes the Rust provider look like it still has a Python path.

PR #136, opened 2026-05-04
E2E testbed: integration tests, profiling, and benchmarking infrastructure

ethenotethan:
## Summary
- Flatten `coordinator/internal/` to `coordinator/`, add E2E integration test suite with real Swift provider + coordinator
- Add multi-model provider specs, user pool with round-robin API key rotation
- Add `X-Timing` JSON response header with per-phase microsecond latency decomposition (parse/reserve/encrypt/dispatch/total)
- Add `SegmentStatsMap()` on `LoadResult` with mean/median/p95/p99/max per segment
- Add 12 integration tests, 3 profiling tests, 8 benchmark tests with coordinator overhead assertions
- Add `PromptBytes` field on `RequestConfig` for controlled prompt sizing
- Add CI workflow for macOS-15 with Postgres, model download, and serial test execution
- `ForceTrustProvider`, `SetChallengeInterval`, `Billing()`, `SetChallengeVerifiedSIP` test helpers on coordinator


Gajesh2007:
@codex can you take a look at this too?

e2e/integration_test.go
```diff
@@ -0,0 +1,616 @@
...
+
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/eigeninference/d-inference/coordinator/payments"
+	"github.com/eigeninference/d-inference/e2e/testbed"
+	tbassert "github.com/eigeninference/d-inference/e2e/testbed/assert"
+	tbprofile "github.com/eigeninference/d-inference/e2e/testbed/profile"
+	"github.com/eigeninference/d-inference/e2e/testbed/profile"
```
Gajesh2007, line 23: **Will Break Build:** Duplicate import — both `tbprofile` (line 22) and `profile` (line 23) alias the same package. `tbprofile` is used on lines 194/226/539, `profile` is used on line 263. Go compiler will reject this.

Fix: remove one alias and use the other consistently.

e2e/testbed/assert/assert_test.go
```diff
@@ -0,0 +1,92 @@
...
+	assert.False(t, report.Passed, "expected fail for missing segment")
+}
+
+func TestDefaultThresholds(t *testing.T) {
+	thresholds := DefaultThresholds()
+	require.NotEmpty(t, thresholds)
+
+	found := false
+	for _, th := range thresholds {
+		if th.Segment == testbed.SegmentTotalE2E {
```
Gajesh2007, line 74: **Will Fail:** `DefaultThresholds()` only contains Parse, Reserve, Encrypt, and Dispatch — there is no `SegmentTotalE2E` entry. This loop won't find it, `found` stays `false`, and the assert on line 77 fails.

Fix: either add `SegmentTotalE2E` to `DefaultThresholds()` or change this test to check for a segment that actually exists.

.github/workflows/deploy-dev-coordinator.yml
```diff
@@ -16,11 +16,10 @@ jobs:
 
       - uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 # v5.3.0
         with:
-          go-version-file: coordinator/go.mod
+          go-version-file: go.mod
 
       - name: Cross-compile coordinator
         run: |
-          cd coordinator
           CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o /tmp/coordinator ./cmd/coordinator
```
Gajesh2007: **Will Break Dev Deploy:** The `cd coordinator` was removed but the build path wasn't updated. This builds `./cmd/coordinator` from the repo root, but the package lives at `./coordinator/cmd/coordinator`.

The Dockerfile got this right (`./coordinator/cmd/coordinator`), but this workflow didn't.

Fix: `go build -a -o /tmp/coordinator ./coordinator/cmd/coordinator`

coordinator/api/consumer.go
```diff
@@ -1068,6 +1079,44 @@ func (s *Server) handleChatCompletions(w http.ResponseWriter, r *http.Request) {
...
+			tj.ParseUs = timing.ParsedAt.Sub(timing.ReceivedAt).Microseconds()
+		}
+		if !timing.ReservedAt.IsZero() && !timing.ParsedAt.IsZero() {
+			tj.ReserveUs = timing.ReservedAt.Sub(timing.ParsedAt).Microseconds()
+		}
+		if !timing.RoutedAt.IsZero() && !timing.ReservedAt.IsZero() {
+			tj.RouteUs = timing.RoutedAt.Sub(timing.ReservedAt).Microseconds()
+		}
+		if !timing.QueuedAt.IsZero() && !timing.DispatchedAt.IsZero() {
+			tj.QueueUs = timing.DispatchedAt.Sub(timing.QueuedAt).Microseconds()
```
Gajesh2007, line 1104: **Wrong Results:** `queue_us` is computed as `DispatchedAt - QueuedAt`, but `DispatchedAt` is set *after* routing, encryption, marshal, and WebSocket write (line 785). Meanwhile `route_us`, `encrypt_us`, and `dispatch_us` already cover those intervals individually.

Timeline: `QueuedAt` → queue wait → `RoutedAt` → encryption → `EncryptedAt` → marshal+write → `DispatchedAt`

So `queue_us` double-counts time already in `route_us`, `encrypt_us`, and `dispatch_us`. The phases are not disjoint — benchmark output will overstate queue latency.

Fix: `queue_us` should be `RoutedAt - QueuedAt` (just the actual queue wait time).

e2e/testbed/profile/profile.go
```diff
@@ -0,0 +1,246 @@
...
+	for _, seg := range []testbed.Segment{
+		testbed.SegmentTotalE2E,
+		testbed.SegmentQueueWait,
+		testbed.SegmentE2EEncrypt,
+		testbed.SegmentCoordinatorToProvider,
+		testbed.SegmentProviderToBackend,
+		testbed.SegmentTTFT,
+		testbed.SegmentDecodeTPS,
+		testbed.SegmentProviderToCoordinator,
+		testbed.SegmentTotalE2E,
```
Gajesh2007, line 233: **Wrong Results:** `SegmentTotalE2E` appears as both the first (line 225) and last (line 233) element in this list. The `total_e2e` row will print twice in the summary table.

This was likely intended to be `testbed.SegmentProviderToClient`.

e2e/testbed/load.go
```diff
@@ -0,0 +1,485 @@
...
+	copy(sorted, durations)
+	sort.Slice(sorted, func(i, j int) bool { return sorted[i] < sorted[j] })
+
+	var total time.Duration
+	for _, d := range sorted {
+		total += d
+	}
+	mean := total / time.Duration(len(sorted))
+	median := sorted[len(sorted)/2]
+	p95Idx := len(sorted) * 95 / 100
```
Gajesh2007, line 473: **Inconsistent Results:** P95 index is computed as `len*95/100` (integer truncation), but `e2e/testbed/profile/profile.go` uses `ceil(n*0.95)-1` (nearest-rank). For 20 samples: this picks index 19 (the max), profile.go picks index 18. The two profilers disagree on P95 values.

Fix: use the same formula in both places. `ceil(n*0.95)-1` is the standard nearest-rank method.

e2e/testbed/load.go
```diff
@@ -0,0 +1,485 @@
...
+	result.TotalDuration = time.Since(start)
+	result.SuccessCount = int(successCount.Load())
+	result.ErrorCount = int(errorCount.Load())
+	result.RequestResults = requestResults
+	result.ProfileRun = &ProfileRun{SegmentTimings: segmentTimings}
+
+	return result
+}
+
+func (lg *LoadGenerator) extractTTFT(body []byte) time.Duration {
```
Gajesh2007, line 289: **Dead Code:** `extractTTFT` unconditionally returns `0` on both branches. The TTFT segment will never be populated from the load generator.

Either implement it (measure wall-clock time to first SSE chunk during the streaming `resp.Body` read, not after) or remove it.

e2e/testbed/suite.go
```diff
@@ -0,0 +1,392 @@
...
+		return env
+	}
+	return "mlx-community/Qwen3.5-0.8B-MLX-4bit"
+}
+
+func (s *Suite) PrimaryModelID() string {
+	return s.Config.PrimaryModelID()
+}
+
+func (s *Suite) Start(ctx context.Context) error {
```
Gajesh2007, line 128: **Resource Leak on Partial Failure:** If a later step fails (e.g. `startProviders` or `startCoordinator`), earlier resources (Postgres, coordinator, partial providers) are never cleaned up. The caller (`startSuite` in tests) registers `t.Cleanup(s.Stop)` only *after* `Start` returns nil, so partial failures leak processes and ports.

Fix: either (a) call `s.Stop()` on error before returning, or (b) register cleanup at the start of the calling test unconditionally — `Stop()` already handles nil fields gracefully.

e2e/testbed/suite.go
```diff
@@ -0,0 +1,392 @@
...
+		state, _ := cmd.Process.Wait()
+		if state != nil && state.ExitCode() >= 0 {
+			p.Logger.Warn("provider process exited", "exit_code", state.ExitCode())
+		}
+	}()
+
+	return nil
+}
+
+func (p *Provider) Stop() {
```
Gajesh2007, line 372: **Double Wait:** `Provider.Start()` (line 366) launches a goroutine that calls `cmd.Process.Wait()`. This `Stop()` method also calls `p.cmd.Wait()` in another goroutine (line 379). Double-waiting on an `os.Process` is a race and returns `"wait: no child processes"` on some platforms.

Fix: coordinate via a `done` channel set by the Start goroutine, and have Stop just wait on that channel + send signals.

coordinator/api/provider.go
```diff
@@ -397,6 +397,10 @@ func providerAttestationMatchesACMEKey(provider *registry.Provider, acmeResult *
 
 // challengeLoop periodically sends attestation challenges to a provider.
 func (s *Server) challengeLoop(ctx context.Context, conn *websocket.Conn, providerID string, provider *registry.Provider, tracker *challengeTracker) {
+	if s.skipChallenge {
```
Gajesh2007, line 400: **Data Race:** `s.skipChallenge` is read here in a provider goroutine without synchronization. The exported setter `SetSkipChallenge` (server.go:390) writes the same field with no mutex or atomic. Same issue for `s.challengeInterval` read on line 404 vs `SetChallengeInterval`.

This will trip `-race` if setters are called while providers are connecting (which the testbed does).

Fix: protect with `s.mu` or use `atomic.Bool` / `atomic.Int64`.

e2e/testbed/deps/postgres.go
```diff
@@ -0,0 +1,230 @@
...
+func (p *PostgresLifecycle) stopNative() {
+	if p.ContainerID == "" {
+		return
+	}
+	var pid int
+	fmt.Sscanf(p.ContainerID, "native:%d", &pid)
+	if pid > 0 {
+		proc, err := os.FindProcess(pid)
+		if err == nil {
+			proc.Signal(os.Interrupt)
```
Gajesh2007, line 215: **Zombie Process:** `stopNative` signals and kills the Postgres process via `os.FindProcess` but never calls `Wait()` on it. This leaves a zombie process on Unix.

Fix: add `proc.Wait()` after the kill, or store the `*exec.Cmd` (which has `Wait`) instead of just the PID string.

e2e/testbed/suite.go
```diff
@@ -0,0 +1,392 @@
...
+	cmd    *os.Process
+	cancel context.CancelFunc
+}
+
+func NewSuite(cfg SuiteConfig) *Suite {
+	logger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelInfo}))
+
+	if os.Getenv("DARKBLOOM_REPO_ROOT") == "" {
+		if cwd, err := os.Getwd(); err == nil {
+			os.Setenv("DARKBLOOM_REPO_ROOT", cwd+"/../..")
```
Gajesh2007, line 82: **Global State Mutation + Wrong Path:** `os.Setenv` is process-global and unsafe with `t.Parallel()`.

Also, the fallback `cwd+"/../.."` is wrong for `go test ./e2e/` — Go test runs with the package dir as cwd, so for the `e2e` package: `e2e/../..` resolves to the *parent* of the repo root, not the repo root. It should be `cwd+"/.."` for the `e2e` package.

Fix: pass repo root through the config struct, or use `t.Setenv()` from the test, and fix the path arithmetic.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,112 @@
...
+      - name: Download model
+        run: |
+          /tmp/mlxvenv/bin/pip install huggingface_hub
+          /tmp/mlxvenv/bin/python -c "from huggingface_hub import snapshot_download; snapshot_download('mlx-community/Qwen3.5-0.8B-MLX-4bit')"
+          /tmp/mlxvenv/bin/python -c "from huggingface_hub import snapshot_download; snapshot_download('mlx-community/gemma-3-270m-4bit')"
+
+      - name: Run integration + profile tests
+        run: |
+          export PATH="$(brew --prefix postgresql@16)/bin:$PATH"
+          go test ./e2e/ -count=1 -v -timeout 25m -p=1 \
```
Gajesh2007, line 49: **CI Gap — Testbed Unit Tests Never Run:** `go test ./e2e/` only runs tests in the `e2e` package itself. Tests in `e2e/testbed/`, `e2e/testbed/assert/`, and `e2e/testbed/profile/` are skipped. Main CI (ci.yml) excludes all `/e2e` paths via `grep -v /e2e`.

This means the testbed framework's own unit tests (`assert_test.go`, `instrument_test.go`, `testbed_test.go`, `profile_test.go`) have zero CI coverage.

Fix: add a separate step with `go test ./e2e/testbed/...` (doesn't need Postgres/provider), or use `go test ./e2e/...` with build tags to separate unit from integration.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,112 @@
...
+      - name: Post benchmark results
+        if: always()
+        run: |
+          if [ ! -f /tmp/benchmark-results.md ]; then
+            echo "No benchmark results file found"
+            exit 0
+          fi
+          BODY=$(cat /tmp/benchmark-results.md)
+          gh pr comment ${{ github.event.pull_request.number }} \
+            --body "$BODY" \
```
Gajesh2007, line 107: **Shell Expansion Risk + Missing Permissions:**

1. `--body "$BODY"` is vulnerable to shell expansion if benchmark markdown contains backticks, `$()`, or other shell metacharacters. Use `--body-file /tmp/benchmark-results.md` instead.

2. This workflow has no `permissions:` block. `gh pr comment` requires `pull-requests: write`, which depends on repo-level defaults and will fail in restricted-token settings. Add `permissions: pull-requests: write` to the benchmark job.

e2e/testbed/assert/assert.go
```diff
@@ -0,0 +1,156 @@
...
+
+type Threshold struct {
+	Segment   testbed.Segment
+	MaxMean   time.Duration
+	MaxP95    time.Duration
+	MaxP99    time.Duration
+	MaxMedian time.Duration
+}
+
+type AssertionResult struct {
```
Gajesh2007, line 18: **Typo:** `AssertionResult` / `AssertionReport` — should be **Assertion** (with an "er"), not "Asserti*on*".

This appears ~30+ times across `assert.go` and `accounting.go`. Rename now before it's entrenched in the API.

e2e/testbed/events.go
```diff
@@ -0,0 +1,114 @@
...
+	events []Event
+}
+
+func NewEventBuffer() *EventBuffer {
+	return &EventBuffer{
+		events: make([]Event, 0),
+	}
+}
+
+func (b *EventBuffer) Consume(event Event) {
```
Gajesh2007, line 72: **Thread Safety:** `EventBuffer.Consume()` appends to a bare slice with no mutex. It implements the public `EventConsumer` interface, so external callers could use it concurrently.

Currently safe only because callers go through `Instrument` (which has its own mutex), but this is a latent bug waiting to happen. Consider adding a mutex, or documenting the single-threaded-only constraint.

AGENTS.md
```diff
@@ -1,41 +1,55 @@
...
+├── auth/             Privy JWT integration
+├── billing/          Stripe, Solana USDC deposits, referrals
+├── e2e/              X25519 request-encryption helpers
+├── mdm/              MicroMDM client + webhook handling
+├── payments/         ledger + pricing
+├── protocol/         WebSocket message types shared with provider
+├── registry/         provider registry, queueing, routing, reputation
+└── store/            in-memory or Postgres persistence
+
+testbed/              System-level testing framework (shared Go module with coordinator)
```
Gajesh2007, line 33: **Wrong Path:** This shows `testbed/` as a top-level directory, but the actual code lives at `e2e/testbed/`.

e2e/testbed/suite.go
```diff
@@ -0,0 +1,392 @@
...
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
+	if s.Coordinator.Registry.ProviderCount() < expectedCount {
+		return fmt.Errorf("only %d/%d providers registered after %v", s.Coordinator.Registry.ProviderCount(), expectedCount, timeout)
+	}
+	s.Logger.Info("providers registered", "count", s.Coordinator.Registry.ProviderCount())
+
+	time.Sleep(3 * time.Second)
```
Gajesh2007, line 269: **Magic Sleep:** No comment explaining what this 3-second delay is for. The original rationale (from commit messages) was "let the initial attestation challenge fire before overriding," but `SetSkipChallenge(true)` is set in `startCoordinator`, so no challenges fire.

Either remove this or add a comment if there's still a reason for it.

Gajesh2007 (commented):
## Code Review

The testbed architecture is excellent — real E2E tests with actual Swift provider + in-process coordinator + ephemeral Postgres, exercising the full HTTP path including billing, encryption, streaming, and attestation. The X-Timing latency decomposition is a good observability addition.

That said, there are issues that need addressing before merge. Inline comments organized by severity:

**Will Break Build/Tests (3):** duplicate import won't compile, `TestDefaultThresholds` will fail, dev deploy workflow broken

**Will Produce Wrong Results (4):** `queue_us` overlaps other phases, duplicate `SegmentTotalE2E` row, P95 formula inconsistency between profilers, `extractTTFT` always returns 0

**Resource Leaks / Race Conditions (5):** `Suite.Start` partial failure cleanup, double `Wait` on provider process, `skipChallenge`/`challengeInterval` unsynchronized reads, native Postgres never reaped, `os.Setenv` global mutation

**CI Gaps (4):** testbed subpackage tests never run, missing permissions for PR comment, duplicated expensive setup across jobs, shell expansion risk in benchmark comment

**Code Quality (5):** "Asserti*on*" typo (should be "Assertion"), `EventBuffer` not thread-safe, `AGENTS.md` wrong path, magic 3s sleep, broken repo-root fallback

**Additionally:** `ForceTrustProvider` masks production routing behavior by setting Python-specific privacy capabilities to true for the Swift provider.

coordinator/registry/registry.go
```diff
@@ -1032,6 +1052,40 @@ func (r *Registry) MarkUntrusted(providerID string) {
...
+	p.LastChallengeVerified = time.Now()
+	p.FailedChallenges = 0
+	p.RuntimeVerified = true
+	p.RuntimeManifestChecked = true
+	if p.PrivacyCapabilities == nil {
+		p.PrivacyCapabilities = &protocol.PrivacyCapabilities{}
+	}
+	p.PrivacyCapabilities.TextBackendInprocess = true
+	p.PrivacyCapabilities.TextProxyDisabled = true
+	p.PrivacyCapabilities.PythonRuntimeLocked = true
```
Gajesh2007, line 1076: **Masks Production Behavior:** `ForceTrustProvider` sets `PythonRuntimeLocked=true` and `DangerousModulesBlocked=true`, which are always false for the Swift provider (no Python runtime). The privacy routing gate (line 206) checks these fields — tests pass through a code path that never exists in production for Swift providers. This can mask real routing failures.

Suggestion: accept a `PrivacyCapabilities` parameter, or add a `ForceTrustSwiftProvider` variant that only sets the capabilities a Swift provider actually reports.

.github/workflows/deploy-dev-coordinator.yml
Gajesh2007: ^ above was codex + claude review. can you delete the dev coordinator?

ethenotethan:
Thanks for the thorough review @Gajesh2007. Pushed commit `6c251c5a` addressing all feedback:

**Will Break Build/Tests:**
- `DefaultThresholds()` now includes `SegmentTotalE2E` — `TestDefaultThresholds` passes
- `challengeInterval`/`skipChallenge` changed to `atomic.Int64`/`atomic.Bool` — no data race. All 30 test assignments updated to use `SetChallengeInterval()`

**Will Produce Wrong Results:**
- Removed duplicate `SegmentTotalE2E` in `profile.go` SummaryTable, added all missing segments
- Unified P95 formula to nearest-rank `ceil(n*p)-1` across `load.go` and `profile.go`
- `extractTTFT` simplified to honest stub — TTFT can't be derived from a full response body (needs streaming instrumentation, which is a separate piece of work)

**Resource Leaks / Race Conditions:**
- `deps/postgres.go`: Store `*exec.Cmd`, call `Wait()` after `Kill()` in `stopNative` to reap zombies
- `suite.go`: Store `*exec.Cmd` instead of `*os.Process` on `Provider`, use `sync.Once` to prevent double-`Wait()`
- Removed `os.Setenv` global mutation — added `RepoRoot` to `SuiteConfig` with correct path (`cwd+"/.."` for the `e2e` package)
- Added `s.Stop()` cleanup on `Suite.Start` partial failures
- Added `sync.Mutex` to `EventBuffer`

**Magic Sleep / ForceTrustProvider:**
- Removed 3s sleep in `waitForProviderRegistration` (was for attestation timing, but `skipChallenge=true` makes it unnecessary)
- `ForceTrustProvider` now accepts `*protocol.PrivacyCapabilities` parameter instead of hardcoding Python-specific values. Callers pass `nil` or provider-appropriate capabilities.

**CI Gaps:**
- Added `go test ./e2e/testbed/...` step (covers assert, profile, instrument unit tests)
- Changed `--body "$BODY"` to `--body-file` to prevent shell expansion
- Added `permissions: pull-requests: write` to benchmark job

**AGENTS.md:**
- Fixed `testbed/` → `e2e/testbed/`, removed non-existent `integration/` dir, added missing files (`suite.go`, `load.go`)

.github/workflows/integration.yml
```diff
@@ -0,0 +1,112 @@
...
+
+on:
+  push:
+    branches: [master, main]
+  pull_request:
+
+jobs:
+  integration-tests:
+    name: E2E Integration Tests
+    runs-on: macos-15
```
Gajesh2007: change the runner to macos-26-xlarge

Gajesh2007 (approved):

PR #146, opened 2026-05-10
Persistent Secure Enclave key with keychain access group enforcement

Gajesh2007:
## Summary

- Adds `PersistentEnclaveKey` — a Secure Enclave P-256 signing key stored in the macOS data protection keychain with team-scoped access group (`SLDQ2GJ6TL.io.darkbloom.provider`)
- Only binaries signed by team `SLDQ2GJ6TL` can access the key — enforced by `securityd` at the kernel level. A patched binary re-signed with `codesign -s -` gets `errSecMissingEntitlement`
- Introduces `AttestationSigner` protocol so `ProviderLoop` tries persistent key first, falls back to ephemeral CryptoKit key gracefully

## Security model

The key is created once via `SecKeyCreateRandomKey` with `kSecAttrTokenIDSecureEnclave` + `kSecAttrIsPermanent: true` and persists in the data protection keychain. The private key never leaves the Secure Enclave hardware — `SecKeyCopyExternalRepresentation` fails by design.

Access enforcement chain:
```
Secure Boot → kernel → AMFI (code signing) → securityd (keychain ACL)
  → access group check: caller must be signed by team SLDQ2GJ6TL
  → Secure Enclave: signs only if securityd approves
```

Verified via POC — 16 attack vectors tested:
- Ad-hoc signed binary: **BLOCKED**
- Different team ID: **BLOCKED**
- Root + `security` CLI: **BLOCKED**
- Raw keychain DB extraction: **encrypted blob only**
- Private key export: **hardware enforced**

## Requires for production

A Developer ID provisioning profile from Apple Developer Portal authorizing the `keychain-access-groups` entitlement for the provider's App ID. Without it, the code gracefully falls back to the existing ephemeral CryptoKit key.

## Test plan

- [x] `swift build` passes
- [x] `swift test` — 67/67 tests pass (8 new persistent enclave key tests)
- [x] POC verified: key creation, persistence, signing, access denial for ad-hoc/cross-team binaries
- [x] Production test with Eigen Labs Developer ID provisioning profile


provider-swift/entitlements.plist
```diff
@@ -0,0 +1,16 @@
...
+<plist version="1.0">
+<dict>
+    <key>com.apple.security.hypervisor</key>
+    <true/>
+    <key>com.apple.security.network.client</key>
+    <true/>
+    <key>com.apple.security.network.server</key>
+    <true/>
+    <key>keychain-access-groups</key>
+    <array>
```
hankbobtheresearchoor, line 14: 🔴 **Wrong entitlement key name.** The key should be `com.apple.security.keychain-access-groups`, not `keychain-access-groups`. Apple's hardened runtime entitlements all live under the `com.apple.security.*` namespace.

The existing `scripts/entitlements.plist` already has this right:
```xml
<key>com.apple.security.keychain-access-groups</key>
```

With the bare `keychain-access-groups`, `codesign --entitlements` will embed an entitlement that securityd doesn't recognize, and keychain access will fail with `-34018` (errSecMissingEntitlement).

Additionally, this file is a duplicate of `scripts/entitlements.plist` (which CI already references). Either fix the key name and update `release-swift.yml` to use this file instead, or remove this file and keep the single source of truth in `scripts/`.

provider-swift/Sources/ProviderCore/Security/PersistentEnclaveKey.swift
```diff
@@ -0,0 +1,299 @@
...
+    /// Load an existing persistent key from the keychain, or create one if not found.
+    public static func loadOrCreate(
+        accessGroup: String? = nil,
+        label: String? = nil
+    ) throws -> PersistentEnclaveKey {
+        let group = resolveAccessGroup(accessGroup)
+        let keyLabel = label ?? defaultLabel
+
+        if let existing = try? findExisting(accessGroup: group, label: keyLabel) {
+            logger.info("Loaded existing persistent Secure Enclave key")
```
hankbobtheresearchoor, line 122: 🟡 **Silent fallthrough on transient keychain errors.** `try?` swallows all errors from `findExisting`, including transient ones like `errSecAuthFailed` (keychain locked) or `errSecDeviceUnavailable`. These should probably be surfaced rather than masked by a fallthrough to `createNew`, which then races with the existing key.

Suggest:
```swift
if let existing = try? findExisting(accessGroup: group, label: keyLabel) {
    return existing
}
// Only fall through on not-found, propagate entitlement/auth errors
```

Or catch `PersistentEnclaveKeyError.keyLookupFailed(status: errSecItemNotFound)` explicitly and re-throw everything else.

provider-swift/Sources/ProviderCore/Security/PersistentEnclaveKey.swift
```diff
@@ -0,0 +1,299 @@
...
+public final class PersistentEnclaveKey: @unchecked Sendable {
+    private let privateKey: SecKey
+    private let _publicKeyRaw: Data
+
+    /// Default access group. The team ID prefix is hardcoded because codesign
+    /// does NOT expand $(AppIdentifierPrefix) -- that's Xcode-only.
+    public static let defaultAccessGroup = "SLDQ2GJ6TL.io.darkbloom.provider"
+
+    public static let defaultLabel = "io.darkbloom.provider.attestation-signing.v1"
+
```
hankbobtheresearchoor, line 77: 🔵 **No rotation path.** The `v1` suffix in `defaultLabel` implies future versions, but there's no mechanism to discover or migrate to `v2`. If the key is compromised or needs rotation:
1. Old key stays orphaned in keychain (no cleanup)
2. Coordinator still has the old SE pubkey stored
3. New provider instances create `v1` keys with different pubkeys, but the label is the same so they'd overwrite

Consider adding a `rotateKey()` static method or a `darkbloom verify --rotate-key` CLI command, and document the coordinator-side SE pubkey update flow.

provider-swift/Sources/ProviderCore/Security/PersistentEnclaveKey.swift
```diff
@@ -0,0 +1,299 @@
...
+        default:
+            throw PersistentEnclaveKeyError.deletionFailed(status: status)
+        }
+    }
+
+    // MARK: - Availability
+
+    /// Whether the Secure Enclave is available on this device.
+    public static var isAvailable: Bool {
+        #if targetEnvironment(simulator)
```
hankbobtheresearchoor: 🔵 **`isAvailable` doesn't check entitlement.** Returns `true` on any macOS 13+ Apple Silicon machine, but persistent SE key also requires the `keychain-access-groups` entitlement. Callers who check `isAvailable` before calling `loadOrCreate()` will get a misleading `true` on unsigned debug builds, then hit `-34018`.

This is consistent with `SecureEnclaveIdentity.isAvailable` (which also doesn't check entitlements), so not blocking — but worth a doc comment noting the entitlement dependency:
```swift
/// - Note: Also requires `com.apple.security.keychain-access-groups` entitlement
///   in the binary's code signature. Without it, `loadOrCreate()` will throw
///   `PersistentEnclaveKeyError.missingEntitlement`.
```

hankbobtheresearchoor (commented):
## Review Summary

Clean abstraction and good graceful fallback logic. Two real issues: the entitlements plist has the wrong key name (🔴), and the `loadOrCreate` error handling should be tighter (🟡). Several observations about the security model implications of persistent identity.

**1 🔴 Must Fix · 1 🟡 Should Fix · 4 🔵 Observations**

hankbobtheresearchoor:
## Security model observations (not line-anchored)

**Persistent identity tradeoff**: This PR changes the SE key from ephemeral (session-scoped) to persistent (survives restarts). Previously, if userspace was compromised between attestation checkpoints, the damage was limited to one session — the key was gone on restart. With persistent keys, the same compromised key can sign attestation challenges across restarts until explicitly deleted. The PR body should acknowledge this tradeoff explicitly. The access-group enforcement mitigates this (only signed binaries can use the key), but it doesn't eliminate the window.

**Coordinator-side implications**: The coordinator already stores `SEPublicKey` in Postgres and builds a lookup map `"sekey:" + SEPublicKey`. With persistent keys, the same SE public key appears across reconnects, enabling identity correlation that wasn't possible with ephemeral keys. The coordinator doesn't currently USE SE keys for identity binding (it uses serial number for `DisconnectDuplicatesBySerial`), but this PR creates the prerequisites. The PR should explicitly state whether SE-key-based identity tracking is intended, to avoid implicit feature drift.

**`DARKBLOOM_KEYCHAIN_ACCESS_GROUP` env var**: Fine for dev/testing, but worth a code comment marking it as non-production. In production, the access group should always come from the entitlement — env var overrides could allow pointing at an attacker-controlled keychain group (though the entitlement still gates access).

Gajesh2007:
@codex can you review this?

PR #153, opened 2026-05-11
E2E tests for R2/S3 bucket flows and Rust→Swift fleet upgrade

ethenotethan:
## E2E Bucket Tests + Fleet Upgrade Validation

Adds MinIO-based integration tests for R2/S3 release registration, self-update, and model weight download flows. Validates the 3-hop fleet upgrade path (v0.4.7 → bridge → Swift).

### New Tests
- `TestIntegration_ReleaseRegistration` — register a release bundle, verify artifact download via R2 CDN URL
- `TestIntegration_SelfUpdateCheck` — provider self-update check returns latest release metadata
- `TestIntegration_ModelWeightDownload` — model weight tarball download from R2 bucket
- `TestIntegration_FleetUpgradeToSwift` — 3-hop fleet upgrade: v0.4.7→bridge (under old coordinator), bridge stability under load, then bridge→Swift (under new coordinator)

### Key Design Decisions
- **MinIO** replaces Docker/LocalStack — GitHub macOS runners lack nested VM support; MinIO runs as a native binary
- **No production `provider/` changes** — E2E tests patch `verify_installed_update_runtime` with `return Ok(());` at build time only. Production provider source is untouched (reset to master).
- **Bridge bundle keeps Python** — it needs Python to serve requests during the migration window
- **Cargo.toml version at 0.4.7** — intentional: this branch freezes `provider/` at 0.4.7. The bridge binary is built from this same source with test-side patching. Future Swift provider lives in `provider-swift/` with its own versioning.
- **`image_bridge_hash`/`grpc_binary_hash` always `None`** — `build_status_canonical` accepts these for forward-compatibility with the Swift provider, which will wire them up. The text-only Rust provider has no image bridge, so `None` is correct here.
- **Swift-detection for `verify_installed_update_runtime`** — not implemented in this PR since we can't modify `provider/`. The E2E tests patch it out. The production fix (check for `python/bin/` presence and skip verification for Swift bundles) belongs in the `swift-provider` branch as a separate PR.

### Cache Key Consistency
The `latest_release:v1:` + platform cache key is used consistently on both the write side (`handleRegisterRelease` invalidation) and read side (`handleLatestRelease`). Verified no mismatch.


hankbobtheresearchoor:
Good test coverage for hops 2 and 3 of the migration, but noticed a gap — the test starts mid-migration with the bridge binary already installed:

```go
// Phase 0: Old coordinator + bridge update to Swift
//
// Simulates the state after v0.4.7→bridge auto-update. The bridge binary
// (current Rust with Swift detection) is already installed.
```

The actual fleet path is three hops:

```
Old Rust (v0.4.7) → Bridge Rust (current + Swift detection) → Swift darkbloom (v0.5.0) → Swift + per-file hash verify
        hop 1 ❌ untested              hop 2 ✅ Phase 0                      hop 3 ✅ Phase 1
```

The test manually `cp`s the bridge binary into `~/.darkbloom/bin/darkbloom` and skips hop 1 entirely. That's the riskiest hop too — it's the biggest code delta (old Rust → current Rust with Swift detection), and if there's a breaking change in the update protocol, config format, or CLI flags between v0.4.7 and the bridge, you'd only hit it in production.

To cover it: build the old Rust provider from `v0.4.7` source (same pattern as `BuildOldCoordinator` — `git archive v0.4.7 provider/` → `cargo build --release`), install it as the starting binary, register the bridge release on the old coordinator, and exercise the old binary's `update` command. Main unknown is whether v0.4.7's update mechanism supports the same flags the test assumes.

hankbobtheresearchoor:
Another gap — the test doesn't validate database migration between v0.4.7 and current.

Phase 0 writes a release row via the old coordinator (v0.4.7 schema — no `binary_hash`/`metallib_hash` columns). Phase 1 starts the new coordinator against the **same** Postgres, but instead of reading back the row the old coordinator wrote, it re-registers a fresh release:

```go
regBody2, _ := json.Marshal(store.Release{
    Version:      "0.5.0",
    ...
    BinaryHash:   bundle.binaryHash,
    BundleHash:   bundle.bundleHash,
    MetallibHash: bundle.metallibHash,
    ...
})
req2, _ := http.NewRequestWithContext(s.Ctx, http.MethodPost,
    coordinatorHTTP+"/v1/releases", strings.NewReader(string(regBody2)))
```

So three things go untested:

1. **Schema migration** — Does the new coordinator's DDL run cleanly against the v0.4.7 schema? If `releases` gained `binary_hash` and `metallib_hash` columns, is it `ALTER TABLE ADD COLUMN` with safe defaults, or does it blow up?

2. **Data readability** — Can the new coordinator read the row v0.4.7 wrote? If the old row lacks the new columns, does a `SELECT` fail or return zero-value strings?

3. **Idempotent re-registration** — The test `POST /v1/releases` for v0.5.0 twice (once on old coord, once on new). Is that an upsert? Overwrite? Conflict? Nothing is asserted about the old row.

To cover it: after stopping the old coordinator and starting the new one, `GET /v1/releases/latest` and assert the old v0.5.0 release is still readable (with empty/null `binary_hash`/`metallib_hash` since v0.4.7 didn't populate those). Then the re-registration tests the upsert/overwrite path explicitly.

hankbobtheresearchoor:
One more thing — the test validates upgrade correctness (binary swaps, hash verification, schema migration) but doesn't validate **network stability during the transition**. The suite starts with `StartWithConfig(ctx, StartConfig{})` which is `{Coordinator: false, Providers: false}` — no providers ever connect, no consumer requests are ever sent.

The only things exercised are:
- `darkbloom update --force` (self-update binary swap)
- `darkbloom update --check-only` (version check)
- `GET /api/version`, `POST /v1/releases`, `GET /v1/releases/latest` (coord CRUD)

So the test proves binaries swap correctly and hashes verify, but doesn't prove a live fleet stays functional mid-migration. Specifically:

1. **Are old providers still connected when old coord stops?** — never tested
2. **Can the new coordinator accept provider heartbeats mid-migration?** — never tested
3. **Can consumer requests route through during the coord swap?** — never tested
4. **Do providers reconnect after coord restart without manual intervention?** — never tested
5. **Is there request loss or latency spike during the swap?** — never tested

A stability test would start a background goroutine sending `POST /v1/chat/completions` throughout all three hops and assert that error rate stays below a threshold and latency doesn't spike beyond some bound. Could also keep a provider connected via WebSocket across the coord restart and assert it reconnects/re-registers automatically.

This is probably worth a follow-up issue rather than blocking this PR — the upgrade mechanism coverage is solid on its own — but the stability gap is real for production confidence.

ethenotethan:
Agreed — the upgrade mechanism coverage is solid but the stability gap is real. Created #154 to track the follow-up: background traffic + live provider WebSocket across coord restart, with error rate and latency assertions.

hankbobtheresearchoor:
Minor point on the new `TestIntegration_FleetUpgradeStability` — it only hits `POST /v1/chat/completions` for background traffic, but the coordinator has ~48 distinct endpoints across 9 domains. Some notable gaps during a cutover:

- **`GET /v1/models`** — if this breaks, consumers can't discover models at all. This is the first call any client makes.
- **`GET /ws/provider`** — the test checks `Registry.ProviderCount() >= 2` on the server side, but never actually validates that a provider WebSocket upgrade succeeds against the new coordinator. The count could come from stored state, not a live connection.
- **`GET /v1/payments/balance`** — billing reads should survive the coord swap. Breakage here means providers can't check earnings mid-migration.
- **`POST /v1/device/code`** — new provider onboarding during the migration window. If this breaks, new nodes can't join the fleet while the cutover is happening.

Not blocking — chat/completions is the highest-value single endpoint to hit — but worth a follow-up to add at least a `GET /v1/models` and a WebSocket reconnection check to the stability loop.

hankbobtheresearchoor:
Looking at the test taxonomy more carefully — the two tests cover opposite corners of the risk matrix but neither touches the high-risk quadrant:

| | Same version | Version upgrade |
|---|---|---|
| **Idle** | trivial | ✅ `FleetUpgradeToSwift` |
| **Under traffic** | ✅ `FleetUpgradeStability` | ❌ untested |

`FleetUpgradeToSwift` proves binaries swap correctly when nobody's watching. `FleetUpgradeStability` proves providers reconnect after a same-version restart. But the actual production risk — old coordinator stops, new coordinator starts on the same port against the same Postgres, providers reconnect under load — falls in the gap between them.

The names also create false confidence. "FleetUpgradeStability" sounds like it proves the fleet upgrade is stable, but it's actually just WebSocket reconnection resilience with the same binary. "FleetUpgradeToSwift" sounds like it proves the fleet upgrade works, but only when idle.

A cleaner taxonomy along the actual risk axes would be:

1. **Upgrade correctness** — binaries swap, hashes verify, schema migrates (no traffic) → *already covered by FleetUpgradeToSwift*
2. **Restart resilience** — same coordinator, providers reconnect, traffic resumes → *already covered by FleetUpgradeStability*
3. **Cutover stability** — old coordinator stops, new coordinator starts on same port, providers reconnect, traffic survives → *not covered*
4. **Provider migration under load** — providers self-update while serving requests → *not covered*
5. **State continuity** — new coordinator inherits trust levels, reputation, queued requests, billing state from old coordinator across the version boundary → *not covered* (the schema migration test only checks release row readability)

The fix isn't more tests — it's one test that combines what already exists: start old coordinator on a fixed port, connect providers, send background traffic, stop old coordinator, start new coordinator on same port against same Postgres, assert providers reconnect and traffic resumes. That covers #3 and is worth more than both existing tests combined for production confidence.

#4 and #5 are follow-up territory.

hankbobtheresearchoor (commented):
## Review Summary

**Verdict: Request Changes**

### 🔴 Must Fix

**1. Production bridge gap: Swift update handling missing from provider/src/main.rs**

The PR deletes `is_swift_release`, `install_swift_update_bundle`, and the Swift branch in both `cmd_update` and `auto_update_check`. The tests patch around this by injecting `return Ok(());` into `verify_installed_update_runtime` at build time, but the actual v0.4.8 bridge release will fail when self-updating to a Swift bundle (no `python/` dir → `verify_installed_update_runtime` fails).

The PR body says the bridge "skips verify_installed_update_runtime when python/ absent (Swift)" — but this logic is not in the code.

**2. `BucketClient` error handling in `NewBucketClient` is broken**

The `CreateBucket` idempotency check declares pointer variables and compares them to `nil`, never using `errors.As`. This silently breaks on every pre-existing bucket. Use `errors.As()`.

### 🟡 Should Fix

**3. Cache invalidation key mismatch**
`release_handlers.go` changed cache invalidation to `latest_release:v1:PLATFORM`. Verify the read side (e.g. `api/server.go` or `handleLatestRelease`) also uses this key format. If not, invalidated cache entries never get read by the fetch path.

**4. `image_bridge_hash` added to canonical but not wired up**
`coordinator.rs:676` passes `None` for `image_bridge_hash` in `handle_attestation_challenge`. Either wire it up to an actual source or drop the parameter to avoid dead attestation surface.

**5. Fleet upgrade test relies on fragile source patching**
`BuildOldProvider` uses `bytes.Replace` on `main.rs` which breaks if formatting shifts. Failing hard when patches don't apply is safer than a warning-and-continue.

**6. Fleet test hardcoded `time.Sleep` durations (15s + 5s + 2s + 10s)**
These make the test slow and flaky. Prefer readiness checks (provider count, HTTP health) over sleeps.

**7. `TestIntegration_FleetUpgradeToSwift` is ~380 lines**
Consider extracting each hop into helpers for debuggability.

**8. `provider/Cargo.toml` version reverted from 0.4.8 to 0.4.7**
If this branch is the bridge release, version should be 0.4.8. If `provider/` is frozen at 0.4.7 and dev moved to `provider-swift/`, state that explicitly.

**9. `FindRepoRoot()` mutates `os.Setenv`**
Hidden global state can confuse parallel tests.

### 🔵 Observations (non-blocking)

**10.** `startSuiteWithBucket` and `startInfrastructure` are nearly identical — could unify.
**11.** Model download test seeds 5 files but asserts only 3 (`chat_template.jinja` and `tokenizer_config.json` never verified).
**12.** `createSwiftReleaseBundle` creates a fake site-packages tarball — clarify if this is backward-compat or leftover.
**13.** Good: MinIO replaces LocalStack cleanly; no Docker dependency.
**14.** Good: `ReleaseRegistration` test is well-structured vertical slice.

e2e/testbed/deps/bucket.go
```diff
@@ -0,0 +1,109 @@
...
+		o.UsePathStyle = true
+	})
+
+	_, err = client.CreateBucket(ctx, &s3.CreateBucketInput{
+		Bucket: aws.String(bucket),
+	})
+	if err != nil {
+		var alreadyOwned *types.BucketAlreadyExists
+		var alreadyOwnedByYou *types.BucketAlreadyOwnedByYou
+		if alreadyOwned == nil && alreadyOwnedByYou == nil {
```
hankbobtheresearchoor: 🔴 **Must Fix:** Broken error type assertion. These declared pointers are always `nil`; the check `alreadyOwned == nil && alreadyOwnedByYou == nil` is always true.

Use `errors.As(err, &alreadyOwned)` / `errors.As(err, &alreadyOwnedByYou)` instead, or the fallback path fires on every pre-existing bucket.

provider/src/coordinator.rs
```diff
@@ -673,6 +673,7 @@ pub fn handle_attestation_challenge(
         rt_hash.as_deref(),
         &template_hashes,
         None, // grpc_binary_hash removed (text-only)
+        None, // image_bridge_hash removed (text-only)
```
hankbobtheresearchoor, line 676: 🟡 **Should Fix:** `image_bridge_hash` is added to `build_status_canonical` but always passed as `None` here. Either wire it up to an actual hash source (if the provider has image-gen capability) or drop the parameter to avoid dead attestation surface.

coordinator/api/release_handlers.go
```diff
@@ -85,7 +85,7 @@ func (s *Server) handleRegisterRelease(w http.ResponseWriter, r *http.Request) {
 	// out the TTL.
 	s.readCache.Invalidate("api_version:v1")
 	s.readCache.Invalidate("runtime_manifest:v1")
-	s.readCache.Invalidate("latest_release:v1")
+	s.readCache.Invalidate("latest_release:v1:" + release.Platform)
```
hankbobtheresearchoor, line 88: 🟡 **Should Fix:** Cache invalidation key changed from `latest_release:v1` to `latest_release:v1:` + `platform`. Verify the **read side** (e.g. `handleLatestRelease` or `api/server.go`) also uses this exact key format. A mismatch means invalidations never clear what the reader fetches — stale releases persist indefinitely.

provider/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "darkbloom"
-version = "0.4.8"
+version = "0.4.7"
```
hankbobtheresearchoor, line 3: 🟡 **Should Fix:** Version reverted from `0.4.8` → `0.4.7`. If this branch is the bridge / final Rust release, it should be `0.4.8`. If the intent is to freeze `provider/` at 0.4.7 and move forward in `provider-swift/`, document that explicitly in the PR description so reviewers don't flag it as a mistake.

provider/src/main.rs
```diff
@@ -6184,43 +5867,38 @@ async fn cmd_update(coordinator: String, force: bool) -> Result<()> {
...
-        verify_installed_update_runtime(&eigeninference_dir, &coordinator_http, true)?;
     }
 
     std::fs::remove_file(tmp_path).ok();
 
+    let coordinator_http = base_url
+        .replace("wss://", "https://")
+        .replace("ws://", "http://")
+        .replace("/ws/provider", "");
+    verify_installed_update_runtime(&eigeninference_dir, &coordinator_http, true)?;
```
hankbobtheresearchoor, line 5900: 🔴 **Must Fix:** Production bridge gap. This `verify_installed_update_runtime` call is unconditional — it checks `python/bin/python3.12` and calls `verify_python_core_signature_match`. When updating to a Swift bundle, there's no `python/` directory, so this will fail at runtime.

The PR body says the bridge "skips verify_installed_update_runtime when python/ absent," but that logic is not implemented. Add Swift-detection (e.g., check for `python/bin/` presence) to skip verification for Swift bundles.

hankbobtheresearchoor:
## Review: E2E bucket tests + fleet upgrade

**Verdict: Request Changes**

The bucket test infrastructure and the three focused tests are solid, but there are two blockers and several cleanup items.

---

### 🔴 Blocker 1: Production gap — Swift update handling removed from Rust provider

`provider/src/main.rs` deletes `is_swift_release`, `install_swift_update_bundle`, and the Swift branches in `cmd_update` / `auto_update_check`. The PR description says the bridge "skips `verify_installed_update_runtime` when `python/` is absent (Swift)" — but that check is **not in the code**.

The E2E test makes this work by **patching the v0.4.7 source at build time** (`return Ok(());` injected via `bytes.Replace` in `BuildOldProvider` / `buildRustProvider`), so the test passes — but the real production bridge binary won't have that patch.

So the test is green, but the actual v0.4.8 bridge release will hit a missing `python/` directory after extracting a Swift bundle, `verify_installed_update_runtime` will fail, and the auto-update aborts. The bridge release needs to either:
- Keep the Swift branch in the update flow, or
- Make `verify_installed_update_runtime` return `Ok(())` when `python/` is missing

---

### 🔴 Blocker 2: `BucketClient` error handling is broken

`e2e/testbed/deps/bucket.go` lines 43–47:
```go
var alreadyOwned *types.BucketAlreadyExists
var alreadyOwnedByYou *types.BucketAlreadyOwnedByYou
if alreadyOwned == nil && alreadyOwnedByYou == nil {
    return nil, fmt.Errorf("testbed/bucket: create bucket: %w", err)
}
```

This never catches the named errors — it checks the declared `nil` pointers, not the `err` value. Use `errors.As()`:

```go
if !errors.As(err, &alreadyOwned) && !errors.As(err, &alreadyOwnedByYou) {
    return nil, fmt.Errorf("testbed/bucket: create bucket: %w", err)
}
```

---

### 🟡 Should Fix

3. **Cache invalidation key mismatch risk** — `release_handlers.go` changes cache invalidation from `latest_release:v1` to `latest_release:v1:` + `release.Platform`. Verify the **read-side** cache key also includes the platform suffix. If not, the release handler invalidates a key the reader never checks.

4. **`image_bridge_hash` added to canonical but not wired up** — `coordinator.rs` adds `image_bridge_hash` to `build_status_canonical`, but `handle_attestation_challenge` always passes `None`. Either wire it up or drop the parameter to avoid dead surface in the attestation protocol.

5. **Fleet upgrade test relies on fragile source patching** — `BuildOldProvider` and `buildRustProvider` use `bytes.Replace` on `main.rs`. These patterns break if formatting changes. Consider failing hard if the patch doesn't apply (currently logs a warning and continues), or using a compile-time feature flag instead of source mutation.

6. **Fleet test has excessive hardcoded `time.Sleep` durations** — `15s` provider start, `5s` traffic warmup, `2s` coordinator gap, `10s` post-cutover. Prefer readiness checks (provider count, HTTP health) over sleeps.

7. **`TestIntegration_FleetUpgradeToSwift` is ~380 lines in one function** — Consider extracting each hop into helpers or sub-tests. The current monolith is hard to debug when it fails.

8. **`provider/Cargo.toml` version reverted from 0.4.8 to 0.4.7** — If this branch is the bridge release, the version should be 0.4.8. If the intent is to freeze `provider/` at 0.4.7 and move development to `provider-swift/`, document that explicitly.

9. **`FindRepoRoot()` mutates `os.Setenv` as a side-effect** — Hidden global state can confuse parallel tests or subprocesses.

---

### 🔵 Observations (non-blocking)

10. `startSuiteWithBucket` and `startInfrastructure` are nearly identical — only difference is `StartWithConfig(ctx, StartConfig{Coordinator:true,Providers:true})` vs `StartWithConfig(ctx, StartConfig{})`. Could unify.

11. Model download test seeds 5 files but asserts only 3 — `chat_template.jinja` and `tokenizer_config.json` are put in MinIO but never verified after download.

12. `createSwiftReleaseBundle` creates a fake site-packages tarball — Swift shouldn't need Python site-packages. Clarify in a comment if this is for backward compat or leftover from the bridge path.

13. Good: MinIO replaces LocalStack cleanly — native binary, no Docker dependency, works on macOS runners.

14. Good: ReleaseRegistration test is well-structured — registers, downloads, hashes, extracts, asserts files. Clean vertical slice.

PR #161, opened 2026-05-14
fix(payments): use len(prompt) as universal token upper bound, remove post-inference cap

### LINKED ISSUE #115: Pre-flight reservation heuristic can undercharge, leaving provider unpaid for work
The coordinator estimates prompt tokens using a `len(string)/4` heuristic (`approximateTokenCount` in `consumer.go`) before debiting the consumer's balance. The reservation is then `estimatePromptTokens × inputRate + maxTokens × outputRate`. After inference, the actual cost is computed from the provider-reported token counts, but capped at `ReservedMicroUSD` — meaning the consumer never pays more than the reservation.

The problem: if the heuristic underestimates prompt tokens (which is likely for code, non-English text, or chat template expansions), the reservation can be lower than the actual cost. The cap silently shorts the provider — they did the work but only get paid 95% of the capped amount, not 95% of actual usage. The consumer gets a discount they didn't ask for, and the provider absorbs the loss.

This is a fairness issue, not a correctness one. Possible fixes include adding a safety margin to the heuristic (e.g., multiply by 1.2×), tracking cumulative undercharge per provider for reconciliation, or accepting the current trade-off with documentation.

hankbobtheresearchoor:
## What

Two changes to eliminate the silent provider shortfall from Issue #115.

## Why

**Issue #115**: The `/4` heuristic underestimates token count for code, non-English text, and chat template expansions. When the estimate is low, the pre-flight reservation is lower than the actual cost, and the post-inference clamp at `ReservedMicroUSD` silently shorts the provider — they do the work but get paid less than 95% of actual.

A second cause: `reservationCost` uses platform pricing, but `handleComplete` uses provider pricing. A provider with a custom price above the platform rate can trigger the clamp even with a perfect token estimate.

### 1. `approximateTokenCount`: `len(x)/4` → `len(x)`

Every BPE tokenizer starts with one token per byte and can only merge, so `len(prompt)` is mathematically ≥ tokens for any model family, any language, forever. The reservation now always covers the actual token workload.

### 2. Try overage charge before clamping

When `totalCost > ReservedMicroUSD` (e.g. provider custom pricing), the coordinator now attempts to charge the consumer the difference first. The consumer already proved they had enough balance for the reservation — the overage is typically small. Only clamps as a last resort if the extra charge fails (balance drained mid-flight).

The clamp is retained as a circuit breaker against provider over-reporting but should never fire on honest work now.

| File | Change |
|------|--------|
| `coordinator/api/consumer.go` | `len(x)/4` → `len(x)` in `approximateTokenCount`; updated comments |
| `coordinator/api/provider.go` | Overage charge before clamp; updated comments |

Closes #115


hankbobtheresearchoor:
This PR targets a stale base and has 16 merge conflicts from rebased Swift-provider commits. I've opened **#186** which cherry-picks only the provider.go billing change (overage charge before clamping) onto a clean origin/master: https://github.com/Layr-Labs/d-inference/pull/186

The TODO comments for the consumer.go / approximation heuristic upgrade are included. #186 has 1 clean commit, 2 files, tests pass.

PR #167, opened 2026-05-15
feat(provider-swift): multi-model concurrent serving

Gajesh2007:
# Summary

- Replaces the Swift provider's single-model-at-a-time runtime with per-model inference slots so one provider can keep multiple models warm and serve them concurrently.
- Adds lazy on-demand loading, per-model schedulers/workers, per-model idle eviction, aggregate capacity heartbeats, and standalone `--local` multi-model serving.
- Hardens the multi-model edge cases found during follow-up review: preload coalescing, shutdown cancellation, global KV budgeting, coordinator capacity accounting, standalone streaming errors, token/auth compatibility, stale capacity handling, live cancellation capacity reporting, live duplicate `load_model` coverage, and standalone HTTP disconnect cleanup.

# Core Changes

## `ProviderLoop.swift`

- Replaces the single scheduler with `modelSlots: [String: ModelSlot]`.
- Tracks `requestToModel` so cancellation, completion, idle eviction, and attestation state stay model-aware.
- Loads models lazily and prevents duplicate same-model loads with continuation-based guards.
- Coalesces duplicate coordinator `load_model` requests and broadcasts one terminal `load_model_status` to all waiters.
- Adds bounded preload shutdown and cancellation-safe drain behavior.
- Evicts idle models in LRU order when a new model needs memory.
- Reports aggregate backend capacity across all loaded slots.
- Adds test-only hooks so live tests can deterministically pause a real MLX load without touching legacy provider key files or creating attestation keys.

## `BatchScheduler` / `GlobalKVCacheBudget`

- Adds global KV-cache reservation accounting across model schedulers.
- Caps total KV reservations at the configured safety factor instead of applying the factor only to remaining memory.
- Rejects duplicate reservation IDs and overflow-sized reservations.
- Bounds model `config.json` reads before parsing.
- Refreshes planner policy when live token budget changes.
- Stops reporting already-cancelled rows as active scheduler capacity while the generator is still applying cancellation internally.

## `StandaloneServer.swift`

- Adds multi-model scheduler cache for `darkbloom start --local`.
- Loads local models on demand from the HuggingFace cache.
- Adds standalone memory-headroom checks and LRU idle eviction before loading additional models.
- Keeps scheduler reservations for the full streaming response lifetime.
- Emits SSE error events for scheduler failures instead of returning a normal stop chunk and `[DONE]`.
- Classifies non-streaming admission/capacity errors as 400/429/503 instead of blanket 500.
- Watches real Hummingbird channel close futures so non-streaming standalone disconnects cancel admitted scheduler work and release reservations.

## Coordinator / Testbed

- Decodes and logs provider `load_model_status` messages.
- Clears stale `backend_capacity` when newer heartbeats omit it.
- Counts coordinator-pending prompt + max tokens consistently for token-budget admission.
- Avoids double-counting provider-reported active/queued token budget.
- Cleans up provider pending reservations on pre-dispatch errors.
- Moves route timing to provider selection instead of including provider billing/eligibility latency.
- Adds testbed support for one provider process advertising multiple `--model` values.
- Adds an opt-in full-network coordinator plus real Swift provider multi-model smoke that runs model A, model B, then model A through one provider.
- Adds memory-store testbed startup for full-network smokes that should not depend on local Docker/Postgres.

## Compatibility / Config

- `start --local --all` now honors `--all` for standalone model selection.
- Swift auth token loading falls back to legacy Rust provider paths and migrates to the canonical Swift path.
- Logout removes migrated legacy auth tokens as well.
- `PrivacyCapabilities` decodes older payloads without `hypervisor_active`.
- Swift backend capacity encoding omits zero-valued additive fields to match Go/Rust wire semantics.
- Adds config migration coverage for `backend.max_model_slots`.

# Before vs After

## Before: one loaded model per Swift provider

```mermaid
flowchart TB
    ConsumerA["Consumer request: model A"] --> Coord["Coordinator"]
    ConsumerB["Consumer request: model B"] --> Coord
    Coord -->|routes request| Loop["Swift ProviderLoop"]

    subgraph BeforeProvider [Swift provider before]
        Loop --> Current["Single currentModel"]
        Current --> Scheduler["Single BatchScheduler"]
        Scheduler --> Worker["Single inference worker"]
        Current -. model switch .-> Reload["Unload current model, then load requested model"]
        Current --> Warm["warm_models reports one model"]
        Scheduler --> Capacity["backend_capacity reports one slot"]
    end

    Reload --> ColdStart["Cold start penalty when model changes"]
```

## After: per-model slots with aggregate capacity

```mermaid
flowchart TB
    ConsumerA["Consumer request: model A"] --> Coord["Coordinator"]
    ConsumerB["Consumer request: model B"] --> Coord
    Coord -->|load_model / inference| Loop["Swift ProviderLoop"]

    subgraph AfterProvider [Swift provider after]
        Loop --> Ensure["ensureModelLoaded(modelID)"]
        Ensure --> Slots{"modelSlots[modelID]"}

        Slots --> SlotA["ModelSlot A: container + BatchScheduler + worker"]
        Slots --> SlotB["ModelSlot B: container + BatchScheduler + worker"]
        Slots --> SlotN["ModelSlot N: container + BatchScheduler + worker"]

        Loop --> RequestMap["requestToModel tracks request ownership"]
        Loop --> KV["GlobalKVCacheBudget caps total KV reservations"]
        Loop --> Eviction["LRU idle eviction frees memory for new models"]
        Loop --> Preload["Duplicate load_model requests coalesce"]

        SlotA --> Heartbeat["heartbeat: warm_models + backend_capacity slots"]
        SlotB --> Heartbeat
        SlotN --> Heartbeat
    end

    Heartbeat --> Coord
    Coord -->|routes by warm model, capacity, and token budget| Loop
```

# Review Notes

- Ran parallel subagent verification for all 24 reported follow-up findings.
- Confirmed and fixed the high-severity issues around `load_model_status`, duplicate preloads, shutdown hangs, KV-budget math, coordinator token accounting, pending cleanup, and standalone streaming errors.
- One reported partial-insert planner leak was not reachable with the current `BatchGenerator.insert` implementation, but the defensive cleanup was added anyway.
- Ran final parallel subagent reviews over coordinator, ProviderLoop/KV, standalone scheduler, compatibility/config, and the integrated coordinator-provider diff.
- Ran local live MLX tests on Apple Silicon after fetching the matching `mlx.metallib`; this exposed and fixed a cancelled-request capacity accounting bug.
- Added deterministic live ProviderLoop + MockCoordinator coverage that pauses a real MLX load, sends a duplicate coordinator `load_model`, and verifies one preload task fans out two `started` and two `succeeded` statuses.
- Added real socket-level standalone disconnect coverage for streaming and non-streaming requests; this exposed and fixed non-streaming reservation cleanup.
- Added and verified an opt-in full-network coordinator plus real Swift provider smoke across two model IDs on the same provider process.
- Final focused subagent review after the standalone disconnect race fix: no actionable correctness findings.

# Test Plan

- [x] `cd coordinator && go test ./...`
- [x] `cd provider-swift && swift test`
- [x] `cd provider-swift && swift test --filter 'ProtocolTests|ConfigTests|DeviceAuthTests|StandaloneServerTests|BatchingTests|GlobalKVCacheBudgetTests'`
- [x] `cd provider-swift && DARKBLOOM_LIVE_MLX_TESTS=1 swift test --filter InferenceLiveTests`
- [x] `cd provider-swift && DARKBLOOM_LIVE_MLX_TESTS=1 DARKBLOOM_LIVE_MLX_MULTI_MODEL=1 swift test --filter InferenceLiveTests`
- [x] `go test ./e2e/testbed/...`
- [x] `go test ./e2e -run 'TestIntegration_FullNetworkSingleSwiftProviderMultiModelRouting|TestDefaultConfigs' -count=1`
- [x] `DARKBLOOM_FULL_NETWORK_SMOKE=1 TESTBED_PROVIDER_CONFIG=debug DARKBLOOM_REPO_ROOT="/Users/gaj/Documents/Builds/d-inference/.claude/worktrees/swift-multi-model" go test ./e2e -run TestIntegration_FullNetworkSingleSwiftProviderMultiModelRouting -count=1 -v`
- [x] `git diff --check`
- [x] Final subagent review over the latest Swift/Go diff
- [ ] Optional 27 GB Gemma live test: `DARKBLOOM_LIVE_MLX_TESTS=1 DARKBLOOM_LIVE_MLX_GEMMA=1 swift test --filter InferenceLiveTests`

# Residual Coverage

- The previous standalone HTTP disconnect and full-network coordinator routing gaps are now covered by opt-in live tests.
- Remaining optional coverage is the large 27 GB Gemma path, gated separately because of model size and hardware requirements.


Gajesh2007:
@codex review it again boy

hankbobtheresearchoor (commented):
## Review Summary

Reviewed the multi-model concurrent serving PR across 4 files (+282/−70). Core architecture is sound — per-model slots with LRU eviction is the right pattern. Swift build passes, ProviderCore tests all pass.

**3 inline findings below.** Overall: 🟡 request changes for the memory estimation bug; the rest are non-blocking.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -682,29 +725,111 @@ public actor ProviderLoop {
...
+    }
 
-        let container = try await loadModelContainer(from: modelPath)
-        await scheduler.loadModel(container: container, modelId: modelId)
+    private func syncWarmModelState() {
+        let loaded = Array(modelSlots.keys).sorted()
+        state.warmModels = loaded
+        if let mostRecent = modelSlots.max(by: { $0.value.lastInferenceAt < $1.value.lastInferenceAt }) {
+            state.currentModel = mostRecent.key
+            state.currentModelHash = loopConfig.modelHashes[mostRecent.key]
```
hankbobtheresearchoor, line 1160: 🔴 **Must Fix: `availableMemoryGb()` overestimates available GPU memory**

`ProcessInfo.processInfo.physicalMemory - MLX.GPU.activeMemory` reports all free physical RAM as available, ignoring macOS kernel/wired memory, other processes, and MLX's own cache overhead. On a 64GB Mac with 10GB of other-process memory, this reports ~54GB available when MLX can realistically allocate far less.

This is especially dangerous combined with the `evictUntilAvailable` loop — if a model needs 20GB and `availableMemoryGb()` reports 24GB when real availability is 18GB, the provider skips eviction and hits OOM during `loadModelContainer`.

At minimum, apply a 0.7× safety factor and subtract `MLX.GPU.cacheMemory`. Better: use `host_statistics` for actual free pages, or track MLX allocations directly.

```swift
// Suggested minimum fix:
let total = ProcessInfo.processInfo.physicalMemory
let active = UInt64(MLX.GPU.activeMemory)
let cache = UInt64(MLX.GPU.cacheMemory)
let safetyFactor = 0.7
let usable = total > (active + cache) ? total - active - cache : 0
return Double(usable) * safetyFactor / (1024.0 * 1024.0 * 1024.0)
```

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -682,29 +725,111 @@ public actor ProviderLoop {
...
+        do {
+            let requiredGb = modelInfo.estimatedMemoryGb * 1.2
+            await evictUntilAvailable(requiredGb: requiredGb)
+
+            logger.info("Loading model: \(modelId) from \(modelPath.path)")
+
+            let container = try await loadModelContainer(from: modelPath)
+            let scheduler = BatchScheduler(
+                maxConcurrentRequests: 4,
+                pendingTimeout: .seconds(120),
```
hankbobtheresearchoor: 🟡 **Missing `updateAggregateCapacity()` after model load succeeds**

`ensureModelLoaded` calls `syncWarmModelState()` on success (line 761) but not `updateAggregateCapacity()`. Compare with `unloadModel` (line 781) which calls both.

This means a newly loaded model won't appear in heartbeat capacity reports until the next inference request triggers capacity update (line 566). If the coordinator preloads a model via `load_model` and then routes requests to it before any inference completes, the capacity snapshot will be stale (showing 0 capacity for that model).

Add `await updateAggregateCapacity()` after `syncWarmModelState()` in the success path.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -577,7 +598,7 @@ public actor ProviderLoop {
     /// `succeeded` -- the coordinator can use this as an idempotent
     /// "ensure warm" call.
     private func handleLoadModelRequest(modelId: String, send: SendHandle) {
```
hankbobtheresearchoor, line 756: 🟡 **Unstructured `Task` at L618 survives actor shutdown**

On line 618 inside this function, `Task { try await me.ensureModelLoaded(...) }` is unstructured — no cancellation token or `isShuttingDown` guard. If the actor begins shutdown (event stream ends, lines 269–275) while a preload is in progress, that Task can insert a model into `modelSlots` *after* shutdown starts, miss the final `unloadModel` sweep (which snapshots keys once at line 273), and leave GPU memory leaked.

Consider: (a) adding a `isShuttingDown` flag checked in `ensureModelLoaded`, (b) storing the Task reference and cancelling it in shutdown, or (c) looping `unloadModel` until `modelSlots` is truly empty at shutdown.

Gajesh2007:
@hankbobtheresearchoor review again now

provider-swift/Sources/ProviderCore/Server/StandaloneServer.swift
```diff
@@ -34,20 +46,70 @@ private let standaloneLogger = Logger(
...
         self.config = config
-        self.scheduler = scheduler
         self.models = models
     }
 
+    public func loadModel(_ modelId: String, container: MLXLMCommon.ModelContainer) async {
+        let scheduler = BatchScheduler(
+            maxConcurrentRequests: 4,
+            pendingTimeout: .seconds(120),
+            defaultMaxTokens: 4096
```
ethenotethan: are these values correct to put here? `maxConcurrentRequests` might be a model specific value. also `defaultMaxTokens` is set by the coordinator

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -682,29 +714,111 @@ public actor ProviderLoop {
...
+
+    /// Evict idle models (LRU order) until `requiredGb` is available or
+    /// no more idle models remain.
+    private func evictUntilAvailable(requiredGb: Double) async {
+        let modelsWithInflight = Set(requestToModel.values)
+
+        while availableMemoryGb() < requiredGb {
+            let candidate = modelSlots
+                .filter { !modelsWithInflight.contains($0.key) }
+                .min(by: { $0.value.lastInferenceAt < $1.value.lastInferenceAt })
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 7ea28622 — evictUntilAvailable now recomputes the in-flight set inside the loop before each await unloadModel().

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -649,24 +670,35 @@ public actor ProviderLoop {
...
+                toEvict.append(modelId)
+            }
+        }
+
+        for modelId in toEvict {
+            if let slot = modelSlots[modelId] {
+                let elapsed = now - slot.lastInferenceAt
+                logger.info("Idle timeout exceeded (\(formatDuration(elapsed)) since last activity); unloading \(modelId)")
+            }
+            await unloadModel(modelId)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 7ea28622 — tickIdleMonitor now re-validates each candidate (re-checks requestToModel and IdleTimeoutPolicy) before each await unloadModel(), so a model that became active during a prior suspension is skipped.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
Gajesh2007: Fixed in f8966082 — availableMemoryGb() now subtracts GPU.cacheMemory and applies a 0.7x safety factor: `(physicalMemory - activeMemory - cacheMemory) * 0.7`.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
Gajesh2007: Fixed in f8966082 — updateAggregateCapacity() is now called after syncWarmModelState() in the success path of ensureModelLoaded, symmetric with unloadModel.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
Gajesh2007: Fixed in f8966082 — preload Tasks are now tracked in a `preloadTasks` dictionary. On shutdown, all preload tasks are cancelled before the model cleanup sweep. CancellationError is caught to exit cleanly.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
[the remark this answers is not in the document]
Gajesh2007: Fixed in f8966082 — syncWarmModelState() now uses `modelSlots.max(by: { $0.value.lastInferenceAt < $1.value.lastInferenceAt })` to pick the most-recently-used model as currentModel. Also, lastInferenceAt is only set on actual inference completion (finishInflightRequest), not on preload, so preloads won't flip currentModel.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -682,29 +725,111 @@ public actor ProviderLoop {
...
-        // Check if already loaded
-        if state.currentModel == modelId {
+        if modelSlots[modelId] != nil {
+            return
+        }
+
+        if modelsLoading.contains(modelId) {
+            try await withCheckedThrowingContinuation { (cont: CheckedContinuation<Void, any Error>) in
+                loadingWaiters[modelId, default: []].append(cont)
+            }
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 4aa05761 — model loads are now serialized through a global gate (`isLoadingAny` + `loadGateWaiters`). Only one model loads at a time, preventing interleaved eviction decisions from overcommitting memory.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -428,23 +447,34 @@ public actor ProviderLoop {
...
+            logger.error("[\(requestId)] Model '\(modelId)' disappeared after load")
+            send.send(.inferenceError(requestId: requestId, error: "model unavailable", statusCode: 500))
+            return
+        }
+
         // 5. Register cancellation token
         let token = await cancellationRegistry.register(requestId: requestId)
 
+        // Track which model this request belongs to
+        requestToModel[requestId] = modelId
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 9e9567e0 — requestToModel is now set before the await cancellationRegistry.register() suspension point, so eviction always sees the model as active during that window.

provider-swift/Sources/ProviderCore/Server/StandaloneServer.swift
```diff
@@ -147,12 +209,28 @@ public actor StandaloneServer {
             return openAIErrorResponse(status: .badRequest, message: "Invalid request body")
         }
 
+        do {
+            try await ensureModelLoaded(chatRequest.model)
+        } catch {
+            return openAIErrorResponse(
+                status: .notFound,
+                message: "Model '\(chatRequest.model)' is not available: \(error.localizedDescription)"
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 9e9567e0 — StandaloneServer now catches StandaloneServerError (model-not-found) as 404, and all other errors (OOM, tokenizer failure, etc.) as 500.

provider-swift/Sources/ProviderCore/Server/StandaloneServer.swift
Gajesh2007: Addressed in 9e9567e0 — extracted to named constants (schedulerMaxConcurrent, schedulerPendingTimeout, schedulerDefaultMaxTokens). Making these fully configurable per-model from the coordinator catalog is being done in a stacked follow-up PR (swift-context-window branch) which adds context_length and max_output_tokens to the model catalog.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -682,29 +750,148 @@ public actor ProviderLoop {
...
+            }
+        }
+
+        // Serialize loads so concurrent eviction decisions don't interleave
+        while isLoadingAny {
+            await withCheckedContinuation { (cont: CheckedContinuation<Void, Never>) in
+                loadGateWaiters.append(cont)
+            }
+            if modelSlots[modelId] != nil { return }
+        }
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 9e9567e0 — the slot cap check is now performed after the load gate wait. The full sequence is: wait for gate -> re-check if model appeared -> check slot cap (evict if needed) -> proceed with load.

Gajesh2007:
@hankbobtheresearchoor approve this and review it on the high level

coordinator/registry/registry.go
```diff
@@ -608,9 +608,11 @@ func TruncHash(h string) string {
...
 // CatalogEntry holds metadata about an active model in the catalog.
 type CatalogEntry struct {
-	ID         string
-	WeightHash string  // expected SHA-256 weight fingerprint (empty = not enforced)
-	SizeGB     float64 // disk/GPU footprint of the model weights (zero = unknown, gate disabled)
+	ID              string
+	WeightHash      string  // expected SHA-256 weight fingerprint (empty = not enforced)
+	SizeGB          float64 // disk/GPU footprint of the model weights (zero = unknown, gate disabled)
+	ContextLength   int     // max context window in tokens (0 = default 131072)
+	MaxOutputTokens int     // max output tokens per request (0 = default 32768)
```
ethenotethan: why are we using `ContextLength` and not `MaxInputTokens`? seems like simpler abstraction that's better semantically coupled with `MaxOutputTokens`

coordinator/api/consumer.go
```diff
@@ -230,7 +230,7 @@ func stripProviderRoutingFields(parsed map[string]any) bool {
 // post-inference charge would fail silently (see GitHub issue #33). Consumers
 // who need longer generations must set max_tokens explicitly and carry the
 // balance to cover it.
-const defaultMaxOutputTokens = 8192
+const defaultMaxOutputTokens = 32768
```
ethenotethan: is this a protocol parameter that we'd eventually wanna control via the admin API?

coordinator/registry/registry.go
```diff
@@ -683,6 +685,35 @@ func (r *Registry) catalogSizeGBLocked(model string) float64 {
 	return 0
 }
 
+const (
+	DefaultContextLength   = 131072 // 128k tokens
```
ethenotethan: how is `131072` equal to 128k tokens?

coordinator/store/postgres.go
```diff
@@ -275,6 +275,15 @@ func (s *PostgresStore) migrate(ctx context.Context) error {
 			ALTER TABLE supported_models ADD COLUMN IF NOT EXISTS weight_hash TEXT NOT NULL DEFAULT '';
 		EXCEPTION WHEN others THEN NULL;
 		END $$`,
+		// Add context_length and max_output_tokens columns for per-model token limits
+		`DO $$ BEGIN
```
ethenotethan: not blocking but eventually we'll wanna segment DDL definitions from code and into separate files

coordinator/store/postgres.go
```diff
@@ -275,6 +275,15 @@ func (s *PostgresStore) migrate(ctx context.Context) error {
...
+			ALTER TABLE supported_models ADD COLUMN IF NOT EXISTS context_length INTEGER NOT NULL DEFAULT 0;
+		EXCEPTION WHEN others THEN NULL;
+		END $$`,
+		`DO $$ BEGIN
+			ALTER TABLE supported_models ADD COLUMN IF NOT EXISTS max_output_tokens INTEGER NOT NULL DEFAULT 0;
```
ethenotethan: these default values will DoS usage for existing models defined in the `supported_models` table. ie. I think this migration would break our production database

coordinator/api/consumer.go
Gajesh2007: yup. this is temporary for the next launch.

coordinator/registry/registry.go
Gajesh2007: https://www.reddit.com/r/LocalLLaMA/comments/1nf8ivi/difference_between_128k_and_131072_context_limit/

Gajesh2007:
@codex please review the latest changes addressing the two remaining unresolved issues:

1. **P1 — evictUntilAvailable now throws on insufficient memory**: Instead of silently breaking and proceeding to load (risking OOM), it now throws `InferenceError.modelLoadFailed` when no idle models can be evicted.

2. **P2 — Standalone server warm-model cache is now bounded**: Added LRU eviction with a cap of 3 cached models, preventing unbounded memory growth in --local mode.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -682,29 +754,156 @@ public actor ProviderLoop {
...
+    private func evictUntilAvailable(requiredGb: Double) async {
+        while availableMemoryGb() < requiredGb {
+            let modelsWithInflight = Set(requestToModel.values)
+            let candidate = modelSlots
+                .filter { !modelsWithInflight.contains($0.key) }
+                .min(by: { $0.value.lastInferenceAt < $1.value.lastInferenceAt })
+
+            guard let (modelId, _) = candidate else {
+                logger.warning("Low memory (\(String(format: "%.1f", availableMemoryGb())) GB free, need \(String(format: "%.1f", requiredGb)) GB) but no idle model to evict")
+                break
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 2ac45dc8 — evictUntilAvailable now throws `InferenceError.modelLoadFailed` when no idle models can be evicted, preventing OOM. The error propagates cleanly through the load gate and waiter machinery.

provider-swift/Sources/ProviderCore/Server/StandaloneServer.swift
```diff
@@ -34,20 +46,74 @@ private let standaloneLogger = Logger(
...
+
+    public func loadModel(_ modelId: String, container: MLXLMCommon.ModelContainer) async {
+        let scheduler = BatchScheduler(
+            maxConcurrentRequests: Self.schedulerMaxConcurrent,
+            pendingTimeout: Self.schedulerPendingTimeout,
+            defaultMaxTokens: Self.schedulerDefaultMaxTokens
+        )
+        await scheduler.loadModel(container: container, modelId: modelId)
+        schedulers[modelId] = scheduler
+    }
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 2ac45dc8 — StandaloneServer now caps warm-model cache at 3 entries with LRU eviction (CachedScheduler wrapper tracks lastUsedAt). Least-recently-used model is evicted via unloadModel() when at capacity.

provider-swift/Sources/ProviderCore/Server/StandaloneServer.swift
```diff
@@ -34,20 +46,74 @@ private let standaloneLogger = Logger(
...
+
+    private func ensureModelLoaded(_ modelId: String) async throws {
+        if schedulers[modelId] != nil { return }
+
+        if modelsLoading.contains(modelId) {
+            try await withCheckedThrowingContinuation { (cont: CheckedContinuation<Void, any Error>) in
+                loadingWaiters[modelId, default: []].append(cont)
+            }
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 22480de4 — StandaloneServer now has a global load gate (`isLoadingAny` + `loadGateWaiters`) mirroring ProviderLoop's pattern. Only one model loads at a time, preventing interleaved OOM.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -594,14 +637,17 @@ public actor ProviderLoop {
         ))
 
         let me = self
-        Task {
+        preloadTasks[modelId] = Task {
+            defer { Task { await me.removePreloadTask(modelId: modelId) } }
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 22480de4 — preload tasks now use UUID ownership tokens. `removePreloadTask` only removes the entry if the completing task's UUID matches, so a newer preload's entry is never removed by an older task's deferred cleanup. Old preloads are also explicitly cancelled before overwriting.

provider-swift/Tests/ProviderCoreTests/StandaloneServerTests.swift
```diff
@@ -80,15 +80,14 @@ import Testing
             headers: [.contentType: "application/json"],
             body: ByteBuffer(string: #"{"model":"mlx-test","messages":[{"role":"user","content":"hello"}],"stream":false}"#)
         ) { response in
-            #expect(response.status == .internalServerError)
-            #expect(String(buffer: response.body).contains("No model loaded"))
+            #expect(response.status == .notFound)
+            #expect(String(buffer: response.body).contains("not available"))
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 22480de4 — test assertion updated from `"not available"` to `"not found locally"` to match `StandaloneServerError.modelNotFound`. All 5 standalone tests pass.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -456,19 +497,25 @@ public actor ProviderLoop {
...
                     )
                 } catch {
-                    log.warning("[\(requestId)] Chunk encryption failed: \(error)")
+                    log.error("[\(requestId)] Chunk encryption failed: \(error)")
+                    send.send(.inferenceError(
+                        requestId: requestId,
+                        error: "response encryption failed",
+                        statusCode: 500
+                    ))
+                    return
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 22480de4 — `emitSSE` now returns `Bool` (false on encryption failure). Every call site checks the return value and `return`s from the task immediately, stopping generation and preventing conflicting terminal signals.

Gajesh2007:
@codex please review the latest changes (commit 22480de4) addressing the 4 remaining issues:

1. **P1 — Standalone model loads serialized**: Added global load gate (`isLoadingAny` + `loadGateWaiters`) so only one model loads at a time, preventing interleaved OOM in `--local` mode.
2. **P1 — Inference task aborts on encryption failure**: `emitSSE` now returns `Bool`; all call sites check the return and exit the task immediately on failure.
3. **P2 — Preload task overwrite race fixed**: Preload tasks use UUID ownership tokens so deferred cleanup cannot remove a newer task's entry. Old preloads are cancelled before overwriting.
4. **P2 — Test assertion fixed**: Updated from `"not available"` to `"not found locally"` to match `StandaloneServerError.modelNotFound`. All 5 standalone tests pass.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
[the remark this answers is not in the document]
Gajesh2007: Already fixed in 22480de4 — `emitSSE` now returns `Bool`; all call sites abort the task on `false`.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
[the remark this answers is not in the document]
Gajesh2007: Already fixed in 22480de4 — preload tasks use UUID ownership tokens for safe cleanup, and old preloads are cancelled before overwriting.

provider-swift/Tests/ProviderCoreTests/StandaloneServerTests.swift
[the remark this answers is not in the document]
Gajesh2007: Already fixed in 22480de4 — assertion updated to `"not found locally"`. All 5 standalone tests pass.

provider-swift/Sources/ProviderCore/Server/StandaloneServer.swift
```diff
@@ -33,21 +45,102 @@ private let standaloneLogger = Logger(
...
+    private static let schedulerMaxConcurrent = 4
+    private static let schedulerPendingTimeout: Duration = .seconds(120)
+    private static let schedulerDefaultMaxTokens = 4096
+
+    public func loadModel(_ modelId: String, container: MLXLMCommon.ModelContainer) async {
+        // Evict least-recently-used model if at capacity
+        if schedulers.count >= Self.maxCachedModels {
+            if let lru = schedulers.min(by: { $0.value.lastUsedAt < $1.value.lastUsedAt }) {
+                await lru.value.scheduler.unloadModel()
+                schedulers.removeValue(forKey: lru.key)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in b5591652 — LRU eviction now queries `scheduler.capacity()` and skips models with `activeRequests > 0` or `pendingRequests > 0`, preventing mid-stream eviction of schedulers with in-flight responses.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -682,29 +754,160 @@ public actor ProviderLoop {
...
+            throw InferenceError.invalidModelDirectory(
+                "Model '\(modelId)' not in advertised model list"
+            )
+        }
+
+        // Serialize loads so concurrent eviction decisions don't interleave
+        while isLoadingAny {
+            await withCheckedContinuation { (cont: CheckedContinuation<Void, Never>) in
+                loadGateWaiters.append(cont)
+            }
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in b5591652 — added `try Task.checkCancellation()` after resuming from the load gate. Cancelled preload tasks now exit immediately instead of proceeding to load after shutdown has started.

Gajesh2007:
@codex please do a deep, thorough review of the entire PR diff — not just the latest commit. Review all files changed across every commit for correctness, concurrency safety, edge cases, and potential regressions. Pay special attention to:

- Actor reentrancy and suspension-point hazards
- Memory management under concurrent multi-model traffic
- Load gate / eviction / shutdown interaction correctness
- Any remaining paths where encryption failure, cancellation, or OOM could leave state inconsistent

Latest commit: b5591652

hankbobtheresearchoor (changes requested):
## Review of #167: `feat(provider-swift): multi-model concurrent serving`

**Verdict: Request changes** — 3 🔴 must-fix concurrency bugs, 5 🟡 should-fix, 3 🔵 observations.

### Summary

Strong architecture — the per-model slot design is clean and the LRU eviction + load gate are well thought out. The code was clearly iterated through multiple review rounds (9 commits reflect this). But there are three actor-reentrancy bugs that would manifest under production load: a shutdown race where preload tasks insert models after the cleanup sweep, a capacity-iteration crash from modifying the dictionary during an `await` loop, and an LRU eviction race in StandaloneServer that can evict a model mid-request.

Build: ✅ | Tests: ✅ (1 pre-existing env failure from missing `mlx.metallib`)

### 🔴 Must fix (3)

1. **Shutdown race with preload tasks** — `run()` cancels preload tasks then snapshots + unloads all `modelSlots`. But a preload task past the cancellation check (already holding `isLoadingAny`) will complete its load and insert a model after the unload sweep. That model stays in GPU memory with no path to unload it.
2. **Actor-reentrant capacity iteration** — `updateAggregateCapacity()` iterates `modelSlots` with `await` calls. During suspension, `unloadModel()` can modify the dictionary. Swift Dictionary concurrent modification during iteration is undefined behavior.
3. **StandaloneServer LRU eviction race** — at capacity (3 models), a request hitting a cold model triggers `ensureModelLoaded(otherModel)` → `loadModel()` → eviction check. The current model's scheduler has no active/pending requests at that point (submit hasn't happened yet), so it's eligible for LRU eviction → 500 error.

### 🟡 Should fix (5)

4. Stale `inferenceActive` on encryption-failure early return
5. StandaloneServer `loadModel` bypasses serialization gate
6. Hardcoded `maxCachedModels = 3` not configurable for different memory sizes
7. Identical scheduler params for all models regardless of capability
8. StandaloneServer missing capacity/counters pattern

### 🔵 Observations (3)

9. `syncWarmModelState` uses MRU for `currentModel` — semantically odd
10. `completionBatchSize` overridden from default 32 → 4 — intentional but worth a comment
11. Multi-model behavior only covered by live MLX tests (skipped in CI)

See inline comments for details and suggested fixes.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -257,8 +286,14 @@ public actor ProviderLoop {
         logger.info("Event stream ended, shutting down")
         idleMonitorTask?.cancel()
         idleMonitorTask = nil
+        for (_, task) in preloadTasks { task.cancel() }
```
hankbobtheresearchoor: 🔴 **Must fix: Shutdown race with preload tasks**

The shutdown sequence at line 289-296 cancels preload tasks, then snapshots + unloads `modelSlots`. But `ensureModelLoaded()` only calls `try Task.checkCancellation()` at the load gate loop (line 806). A preload task past that point (already holding `isLoadingAny = true`) will complete `loadModelContainer()`, insert into `modelSlots` (line 847), and call `syncWarmModelState()` / `updateAggregateCapacity()` — all after the unload sweep has already completed.

The model stays loaded in GPU memory with no path to unload it because the event loop has exited.

**Fix options:**
- Add an `isShuttingDown` flag checked at the top of `ensureModelLoaded` (before gate acquisition)
- Loop `unloadModel` until `modelSlots.isEmpty` at shutdown (defensive, handles any late-arriving slot insertions)
- Both

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -737,11 +964,37 @@ public actor ProviderLoop {
...
+        if let modelId = requestToModel.removeValue(forKey: requestId) {
+            modelSlots[modelId]?.lastInferenceAt = .now
+            syncWarmModelState()
+        }
+    }
+
+    private func updateAggregateCapacity() async {
+        var allSlots: [BackendSlotCapacity] = []
+        var totalActive = 0
+        for (_, slot) in modelSlots {
```
hankbobtheresearchoor: 🔴 **Must fix: Actor-reentrant capacity iteration crash**

`updateAggregateCapacity()` iterates `modelSlots` directly (line 981) while making `await` calls (lines 982, 984). Between these suspension points, `unloadModel()` can run on the same actor and remove entries from `modelSlots` via `modelSlots.removeValue(forKey:)`.

Swift Dictionary iteration with concurrent modification is undefined behavior — can crash or silently skip entries.

**Fix:** Snapshot `modelSlots` before iterating:
```swift
let slots = modelSlots  // captures current dict state
for (_, slot) in slots { ... }
```

provider-swift/Sources/ProviderCore/Server/StandaloneServer.swift
```diff
@@ -33,21 +45,141 @@ private let standaloneLogger = Logger(
...
+    private static let schedulerDefaultMaxTokens = 4096
+
+    public func loadModel(_ modelId: String, container: MLXLMCommon.ModelContainer) async {
+        // Evict least-recently-used idle model if at capacity.
+        // Skip schedulers with active requests to avoid truncating in-flight responses.
+        if schedulers.count >= Self.maxCachedModels {
+            var lruKey: String?
+            var lruTime: ContinuousClock.Instant?
+            for (key, cached) in schedulers {
+                let cap = await cached.scheduler.capacity()
```
hankbobtheresearchoor: 🔴 **Must fix: LRU eviction race under concurrent load**

When `schedulers.count >= maxCachedModels` (3), a request for an already-warm model A that triggers `ensureModelLoaded(modelB)` for a cold model B will call `loadModel()` on B. Inside `loadModel()`, the LRU eviction loop (lines 85-92) checks `scheduler.capacity()` for all schedulers. At this point, model A's scheduler has `activeRequests: 0` and `pendingRequests: 0` because `scheduler.submit()` for model A hasn't been called yet — it happens at line 305/372, *after* `ensureModelLoaded` returns.

So model A is eligible for eviction, gets unloaded, and when `chatCompletionsResponse` reaches line 296, `schedulers[chatRequest.model]` is nil → 500 error "Model load succeeded but scheduler unavailable".

**Fix:** Touch the current model as "in-use" before calling `ensureModelLoaded` for other models. E.g. increment a per-model request counter that the LRU loop checks, or mark the scheduler as reserved before the `ensureModelLoaded` call.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -428,23 +463,34 @@ public actor ProviderLoop {
...
 
         // 7. Spawn inference task
+        let me = self
         let task = Task.detached {
             defer {
                 Task { await registry.finish(requestId: requestId) }
+                Task { await me.finishInflightRequest(requestId: requestId) }
             }
 
             providerState.inferenceActive = true
```
hankbobtheresearchoor: 🟡 **Should fix: Stale `inferenceActive` on encryption-failure early return**

`providerState.inferenceActive = true` is set at line 496 (Task.detached start). It's only reset by `updateAggregateCapacity()` at line 596, which runs on the **success** path. If `emitSSE()` fails (encryption error) and the task returns early (line 536, 564, 574, 587, 588), `inferenceActive` stays `true` until another request's completion calls `updateAggregateCapacity()`.

This means a failed request can leave the provider reporting as active in heartbeats when no real work is happening, misleading the coordinator's routing.

**Fix:** Add `await me.updateAggregateCapacity()` (or direct `providerState.inferenceActive = false` + check if others are active) in the `defer` block alongside `finishInflightRequest`. Or call it from `finishInflightRequest` itself since it already runs on every request completion.

provider-swift/Sources/ProviderCore/Server/StandaloneServer.swift
```diff
@@ -33,21 +45,141 @@ private let standaloneLogger = Logger(
...
         self.config = config
-        self.scheduler = scheduler
         self.models = models
     }
 
+    private static let schedulerMaxConcurrent = 4
+    private static let schedulerPendingTimeout: Duration = .seconds(120)
+    private static let schedulerDefaultMaxTokens = 4096
+
+    public func loadModel(_ modelId: String, container: MLXLMCommon.ModelContainer) async {
```
hankbobtheresearchoor: 🟡 **Should fix: `loadModel()` bypasses serialization gate**

The public `loadModel()` method doesn't check or set `isLoadingAny`. While `ensureModelLoaded` is suspended at the load gate (`while isLoadingAny` loop), another call to `loadModel()` can proceed without waiting. Currently only called from `ensureModelLoaded` itself (line 155), but it's a public method — any external caller can bypass the gate and cause concurrent model loads.

**Fix:** Either make `loadModel` private, or add the gate check. Since `ensureModelLoaded` already handles gate + LRU eviction + model path resolution, `loadModel` could be the raw "do the load" operation while `ensureModelLoaded` handles all gating. Making it private would clarify this contract.

provider-swift/Sources/ProviderCore/Server/StandaloneServer.swift
```diff
@@ -33,21 +45,141 @@ private let standaloneLogger = Logger(
...
+    private var modelsLoading: Set<String> = []
+    private var loadingWaiters: [String: [CheckedContinuation<Void, any Error>]] = [:]
+    private var isLoadingAny: Bool = false
+    private var loadGateWaiters: [CheckedContinuation<Void, Never>] = []
     private var models: [ModelInfo]
     private var serverTask: Task<Void, Never>?
 
+    /// Maximum number of models kept warm in standalone mode. Oldest models
+    /// are evicted (LRU) when this limit is reached.
+    private static let maxCachedModels = 3
```
hankbobtheresearchoor: 🟡 **Should fix: Hardcoded `maxCachedModels = 3`**

Reasonable default, but a Mac with 192GB RAM can hold 10+ small models. The 3-model cap should be configurable — either via `StandaloneServerConfig` or derived from available RAM / model sizes at server init. The provider loop already has `maxModelSlots` derived from config; standalone mode should match this configurability.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -118,11 +139,19 @@ public actor ProviderLoop {
...
-            maxConcurrentRequests: 4,
-            pendingTimeout: .seconds(120),
-            defaultMaxTokens: 4096
-        )
+        self.maxModelSlots = max(1, config.models.count)
+    }
+
+    // MARK: - Model Slot
+
+    private static let schedulerMaxConcurrent = 4
```
hankbobtheresearchoor: 🟡 **Should fix: Identical scheduler params for all models**

`schedulerMaxConcurrent = 4`, `schedulerDefaultMaxTokens = 4096`, `schedulerPendingTimeout = 120s` are applied to every model regardless of size or capability. A 0.5B model and a 70B model get the same concurrency and token budget. The `BatchScheduler.loadModel()` already computes a dynamic token budget from `config.json` architecture fields (lines 244-297 in BatchScheduler.swift). These static params could be per-model overrides from `ModelInfo` or `loopConfig` so small models can serve more concurrent requests and large models fewer.

Gajesh2007:
@hankbobtheresearchoor pushed fc26956e addressing your multi-model concurrency review.

Summary:
- ProviderLoop now drains active inference on shutdown, rejects new requests while draining, cancels/awaits preload tasks, and uses a macOS no-idle-sleep assertion while jobs are active.
- Added unloading tombstones/waiters so a model cannot be reloaded while its scheduler is still tearing down.
- Capacity aggregation now snapshots slots before awaiting, and inferenceActive is refreshed on every request exit path.
- Added backend.max_model_slots with default 3 so resident model slots are bounded independently of advertised model count.
- StandaloneServer now has configurable cache cap, request reservations around scheduler submit, private raw load path, snapshot-based LRU eviction, and 503 when all cached models are active.
- Added GlobalKVCacheBudget shared across schedulers so KV cache reservations are process-wide instead of per-model only.

Verified:
- swift test --filter standaloneServer
- swift test --filter Batching
- git diff --check

Please re-review when you have a chance.

Gajesh2007:
@codex review this

provider-swift/Sources/ProviderCore/ProviderLoop.swift
Gajesh2007: Deferring this from the multi-model PR rather than adding a size heuristic here. The constructor values are shared hard/default bounds, but the scheduler becomes model-sensitive after load: `BatchScheduler.loadModel()` derives `kvBytesPerToken` and `dynamicTokenBudgetMax` from the model `config.json`, admission clamps against live token/KV headroom, and `effectiveMaxConcurrentRequests` combines that with the adaptive cap. The adaptive cap starts conservatively and only expands toward the hard cap when observed per-request TPS supports it.

The follow-up I prefer is explicit catalog/config policy, e.g. context length, max output tokens, and optional scheduler overrides in `ModelInfo`/loop config, instead of guessing from model size in this PR.

Gajesh2007:
@codex review it one more time

provider-swift/Sources/darkbloom/StartCommand.swift
```diff
@@ -110,15 +111,8 @@ struct Start: AsyncParsableCommand {
...
         defer { ProcessLifecycle.releaseSingleInstanceLock() }
 
-        let scheduler = BatchScheduler(
-            maxConcurrentRequests: 4,
-            pendingTimeout: .seconds(120),
-            defaultMaxTokens: 4096
-        )
-
         let server = StandaloneServer(
             config: StandaloneServerConfig(port: port, host: "127.0.0.1"),
```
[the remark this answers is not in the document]
Gajesh2007: Legit. Fixed in `1dab918e`: standalone `--local` now passes `config.backend.maxModelSlots` through to `StandaloneServerConfig.maxCachedModels` using `Int(clamping:)`, so the existing `backend.max_model_slots` config controls the standalone warm-model cache cap.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -737,11 +1243,67 @@ public actor ProviderLoop {
...
+        requestToModel.removeAll()
+        syncWarmModelState()
+    }
+
+    private func finishInflightRequest(requestId: String) async {
+        let hadRegisteredTask = inflightTasks.removeValue(forKey: requestId) != nil
+        let modelId = requestToModel.removeValue(forKey: requestId)
+        if !hadRegisteredTask, modelId != nil {
+            completedBeforeTaskRegistration.insert(requestId)
+        }
```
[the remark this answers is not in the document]
Gajesh2007: I think this one is stale for the current actor flow. There is no `await` or actor suspension between `Task.detached` creation and `inflightTasks[requestId] = task`, so a queued cancel cannot remove `requestToModel` before the task handle is registered. If the detached task finishes immediately, its deferred `finishInflightRequest` actor call also waits until the current actor job completes, after registration. The `completedBeforeTaskRegistration` set remains defensive cleanup for fast completion, but the reported stale-task shutdown hang should not be reachable unless a future edit adds an await in that spawn/register window.

PR #181, opened 2026-05-18
feat: add adaptive provider batching capacity

Gajesh2007:
## Summary

This PR adds provider-reported adaptive batching capacity on top of the multi-model Swift provider branch.

- Swift `BatchScheduler` keeps a hard engine slot ceiling and learns an effective per-model concurrency cap from observed per-batch decode TPS.
- Swift admission remains KV-aware: a request must fit the model's token budget, and pending-to-active promotion re-checks active reserved tokens before allocating a row.
- Swift heartbeats now report per-slot `max_concurrency`, active/queued token budgets, observed TPS, and KV bytes per token.
- Coordinator routing consumes positive per-slot `max_concurrency` for model-aware headroom checks while preserving legacy fallback for omitted/zero values.
- Coordinator model capacity snapshots and routing gates now treat saturated per-model slots as capacity-limited rather than structurally ineligible.

## Before

```mermaid
flowchart TD
    A[Consumer request] --> B[Coordinator selects provider]
    B --> C{Provider-wide cap}
    C -->|Default or memory tier| D[Route to provider]
    D --> E[Swift ProviderLoop]
    E --> F[BatchScheduler]
    F --> G{Fixed active rows = 4}
    G -->|free row| H[BatchGenerator active batch]
    G -->|no row| I[Provider pending queue]
    H --> J[Decode]
    I --> H

    K[KV/token budget] -. checked at submit .-> F
    L[Actual per-batch TPS] -. not used for cap .-> F
    M[Multi-model slots] -. coordinator sees provider-wide cap .-> C
```

## After

```mermaid
flowchart TD
    A[Consumer request] --> B[Coordinator preflight and routing]
    B --> C{Per-model slot cap reported?}
    C -->|yes| D[Use slot max_concurrency]
    C -->|no or zero| E[Legacy provider-wide fallback]
    D --> F{Provider-wide safety cap also has headroom?}
    E --> F
    F -->|headroom| G[Route to provider]
    F -->|full| Q[Capacity rejection / queue]

    G --> H[Swift ProviderLoop]
    H --> I[BatchScheduler]
    I --> J{Request fits token budget alone?}
    J -->|no| R[token_budget_exhausted]
    J -->|yes| K[Pending queue]

    K --> L{Active rows below dynamic cap?}
    L -->|yes| M{Active reserved tokens + request <= live KV budget?}
    M -->|yes| N[Promote to BatchGenerator]
    M -->|no| K
    L -->|no| K
    N --> O[Decode burst]
    O --> P[Record batch-size TPS EWMA]
    P --> S[Expand or shrink dynamic cap]
    S --> T[Heartbeat backend_capacity slot max_concurrency]
    T --> B
```

## Testing

- `cd coordinator && go test ./...`
- `git diff --check`

Swift verification is blocked in this worktree because local path dependencies are absent:

- `libs/mlx-swift/Package.swift`
- `libs/mlx-swift-lm/Package.swift`

## Review Notes

Focused review passes were run for Swift scheduling, coordinator routing, and test coverage. Follow-up fixes addressed submit-time over-rejection, mixed-size pending promotion, fast detached task completion, model capacity readiness, capacity rejection accounting, backend-reported load saturation, and provider-wide safety headroom.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/181"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

### DIFF
--- a/coordinator/registry/registry.go
+++ b/coordinator/registry/registry.go
@@ -356,6 +356,15 @@ func (p *Provider) MaxConcurrency() int {
 	return p.maxConcurrency()
 }
 
+// MaxConcurrencyForModel returns the concurrency limit for a specific model.
+// A positive provider-reported slot cap wins; zero/missing preserves the
+// legacy provider-level fallback.
+func (p *Provider) MaxConcurrencyForModel(model string) int {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	return p.maxConcurrencyForModelLocked(model)
+}
+
 // maxConcurrency is the lock-free version (caller must hold p.mu).
 //
 // Tier values were lowered in Phase 2 of the routing-algorithm rework
@@ -397,6 +406,66 @@ func (p *Provider) maxConcurrency() int {
 		cap = 12
 	}
 	return cap
+}
+
+// maxConcurrencyForModelLocked is the lock-free model-aware concurrency cap.
+// Caller must hold p.mu.
+func (p *Provider) maxConcurrencyForModelLocked(model string) int {
+	if p.BackendCapacity != nil {
+		for _, slot := range p.BackendCapacity.Slots {
+			if slot.Model == model && slot.MaxConcurrency > 0 {
+				return slot.MaxConcurrency
+			}
+		}
+	}
+	return p.maxConcurrency()
+}
+
+func (p *Provider) pendingCountForModelLocked(model string) int {
+	count := 0
+	for _, pr := range p.pendingReqs {
+		if pr.Model == model {
+			count++
+		}
+	}
+	return count
+}
+
+func (p *Provider) hasReportedMaxConcurrencyForModelLocked(model string) bool {
+	if p.BackendCapacity == nil {
+		return false
+	}
+	for _, slot := range p.BackendCapacity.Slots {
+		if slot.Model == model && slot.MaxConcurrency > 0 {
+			return true
+		}
+	}
+	return false
+}
+
+func (p *Provider) pendingLoadForModelLocked(model string) int {
+	if !p.hasReportedMaxConcurrencyForModelLocked(model) {
+		return p.pendingCount()
+	}
+	load := p.pendingCountForModelLocked(model)
+	if p.BackendCapacity != nil {
+		for _, slot := range p.BackendCapacity.Slots {
+			if slot.Model != model {
+				continue
+			}
+			backendLoad := slot.NumRunning + slot.NumWaiting
+			if backendLoad > load {
+				load = backendLoad
+			}
+			break
+		}
+	}
+	return load
+}
+
+func (p *Provider) hasConcurrencyHeadroomForModelLocked(model string) bool {
+	return p.pendingLoadForModelLocked(model) < p.maxConcurrencyForModelLocked(model) &&
+		p.pendingCount() < p.maxConcurrency()
 }
 
 // Registry holds all connected providers and provides routing.
@@ -748,13 +817,14 @@ func (r *Registry) SetQueue(q *RequestQueue) {
 // tok/s, max Mac Studio RAM is 512 GB) so legitimate future hardware isn't
 // clamped unnecessarily.
 const (
-	maxDecodeTPS                = 500.0
-	maxPrefillTPS               = 5000.0
-	maxMemoryBandwidthGBs       = 2000.0
-	maxMemoryGB                 = 1024
-	maxMemoryGBFloat            = 1024.0
-	maxTokensPotential          = 1_000_000
-	maxTokenBudgetCap     int64 = 10_000_000_000 // 10 billion — generous safety valve for total token budget capacity
+	maxDecodeTPS                    = 500.0
+	maxPrefillTPS                   = 5000.0
+	maxMemoryBandwidthGBs           = 2000.0
+	maxMemoryGB                     = 1024
+	maxMemoryGBFloat                = 1024.0
+	maxReportedMaxConcurrency       = 24
+	maxTokensPotential              = 1_000_000
+	maxTokenBudgetCap         int64 = 10_000_000_000 // 10 billion — generous safety valve for total token budget capacity
 )
 
 // clampNonNeg returns v clamped into [0, max]; NaN/negative become 0.
@@ -810,6 +880,15 @@ func clampBackendCapacity(logger *slog.Logger, providerID string, bc *protocol.B
 		}
 		if s.NumWaiting < 0 {
 			s.NumWaiting = 0
+		}
+		if s.MaxConcurrency < 0 || s.MaxConcurrency > maxReportedMaxConcurrency {
+			logger.Warn("provider slot max_concurrency out of range, clamping",
+				"provider_id", providerID, "model", s.Model, "reported", s.MaxConcurrency)
+			if s.MaxConcurrency < 0 {
+				s.MaxConcurrency = 0
+			} else {
+				s.MaxConcurrency = maxReportedMaxConcurrency
+			}
 		}
 		if v, changed := clampNonNeg(s.ObservedDecodeTPS, maxDecodeTPS); changed {
 			logger.Warn("provider slot observed_decode_tps out of range, clamping",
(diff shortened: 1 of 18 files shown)


coordinator/registry/registry.go
```diff
@@ -356,6 +356,15 @@ func (p *Provider) MaxConcurrency() int {
 	return p.maxConcurrency()
 }
 
+// MaxConcurrencyForModel returns the concurrency limit for a specific model.
+// A positive provider-reported slot cap wins; zero/missing preserves the
+// legacy provider-level fallback.
+func (p *Provider) MaxConcurrencyForModel(model string) int {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	return p.maxConcurrencyForModelLocked(model)
```
ethenotethan, line 365: @claude is this a block or non-blocking read mutex?

Gajesh2007: It’s an exclusive `sync.Mutex`, matching the existing `Provider` locking pattern. The critical section is tiny and protects concurrent reads of `BackendCapacity` / pending request state while heartbeats and routing paths mutate them. We could move `Provider.mu` to `sync.RWMutex` later if profiling shows contention, but a plain mutex is the safer/simple production-grade default here.

coordinator/registry/registry.go
```diff
@@ -811,6 +881,15 @@ func clampBackendCapacity(logger *slog.Logger, providerID string, bc *protocol.B
 		if s.NumWaiting < 0 {
 			s.NumWaiting = 0
 		}
+		if s.MaxConcurrency < 0 || s.MaxConcurrency > maxReportedMaxConcurrency {
+			logger.Warn("provider slot max_concurrency out of range, clamping",
+				"provider_id", providerID, "model", s.Model, "reported", s.MaxConcurrency)
+			if s.MaxConcurrency < 0 {
+				s.MaxConcurrency = 0
```
ethenotethan, line 888: think it'd be easiest to set `MaxConcurrency` a non-negative variable to avoid <0 expressions ever being possible

Gajesh2007:
@codex are there any logical flaws in this or something that might break the system completely or not perform what is being requested?

hankbobtheresearchoor (commented):
## Review Summary

This is a well-structured PR. The core idea — letting the Swift provider dynamically learn per-model throughput limits and report them to the coordinator for per-model routing — is a genuine improvement over the old flat provider-wide cap. The two-sided implementation (Swift adaptive caps + Go per-model gates) is coherent.

**Verdict: Approve** — no blocking issues. A few suggestions below.

### What works well
- The dual-gate in routing: per-slot cap AND provider-wide safety cap. Protects against both model-level saturation and provider-level overload.
- The `completedBeforeTaskRegistration` race fix in `ProviderLoop` — handles the edge case correctly without adding complexity.
- The clamp on `MaxConcurrency` (0–24) in the coordinator prevents malicious heartbeat values.
- Good test coverage: 4 new Go scheduler tests + updated clamp tests + Swift forward/backward compat tests.
- The EWMA in `BatchPerformanceBucket` uses `alpha=0.25` which gives reasonable reactivity for an inference workload.

### Suggestions (see inline comments)
- 🟡 `hasConcurrencyHeadroomForModelLocked` name is misleading — it gates on provider-wide cap too
- 🔵 `pendingLoadForModelLocked` `break` after first matching slot
- 🔵 `capacityRefreshTask` queries all slots regardless of which changed

— reviewed by Hermes

coordinator/registry/registry.go
```diff
@@ -399,6 +408,66 @@ func (p *Provider) maxConcurrency() int {
...
+			if backendLoad > load {
+				load = backendLoad
+			}
+			break
+		}
+	}
+	return load
+}
+
+func (p *Provider) hasConcurrencyHeadroomForModelLocked(model string) bool {
```
hankbobtheresearchoor, line 466: 🟡 **Suggestion:** The name  implies a model-level-only check, but the function also gates on the provider-wide . This double-gate is **deliberate and correct** — it prevents a model-level green light from accidentally overloading the provider. Consider adding a brief doc comment explaining the dual purpose: the second term is the provider-wide safety valve, not redundant with the per-slot cap.

hankbobtheresearchoor, line 466: 🟡 **Suggestion:** The name `hasConcurrencyHeadroomForModelLocked` implies a model-level-only check, but the function also gates on the provider-wide `pendingCount() < maxConcurrency()`. This double-gate is deliberate and correct — it prevents a model-level green light from accidentally overloading the provider. Consider adding a brief doc comment explaining that the second term is the provider-wide safety valve.

coordinator/registry/registry.go
```diff
@@ -399,6 +408,66 @@ func (p *Provider) maxConcurrency() int {
...
+	if p.BackendCapacity != nil {
+		for _, slot := range p.BackendCapacity.Slots {
+			if slot.Model != model {
+				continue
+			}
+			backendLoad := slot.NumRunning + slot.NumWaiting
+			if backendLoad > load {
+				load = backendLoad
+			}
+			break
```
hankbobtheresearchoor, line 460: 🔵 **Observation:** `pendingLoadForModelLocked` uses `break` after matching the first slot for a model. If a provider ever reports duplicate model entries across multiple slots, only the first slot’s backend load would be considered. This is fine for the current single-slot-per-model architecture, but a comment noting the assumption would help future readers.

provider-swift/Sources/ProviderCore/ProviderLoop.swift
```diff
@@ -770,6 +786,22 @@ public actor ProviderLoop {
         return remMinutes == 0 ? "\(hours)h" : "\(hours)h\(remMinutes)m"
     }
 
+    // MARK: - Capacity Refresh
+
+    private func startCapacityRefreshMonitor() {
```
hankbobtheresearchoor, line 791: 🔵 **Observation:** `capacityRefreshTask` calls `updateAggregateCapacity()` at half the heartbeat interval, which queries **all** model slots’ schedulers for capacity. For a provider with many loaded models, most slots haven’t changed state between refreshes. Consider passing the affected model ID and only refreshing that slot — or keep the full refresh since it’s cheap (just reading actor properties) and simpler.

coordinator/registry/registry.go
```diff
@@ -748,13 +817,14 @@ func (r *Registry) SetQueue(q *RequestQueue) {
...
-	maxTokensPotential          = 1_000_000
-	maxTokenBudgetCap     int64 = 10_000_000_000 // 10 billion — generous safety valve for total token budget capacity
+	maxDecodeTPS                    = 500.0
+	maxPrefillTPS                   = 5000.0
+	maxMemoryBandwidthGBs           = 2000.0
+	maxMemoryGB                     = 1024
+	maxMemoryGBFloat                = 1024.0
+	maxReportedMaxConcurrency       = 24
+	maxTokensPotential              = 1_000_000
+	maxTokenBudgetCap         int64 = 10_000_000_000 // 10 billion — generous safety valve for total token budget capacity
```
ethenotethan, line 827: One thing worth noting: if we ever need to change these values, it would require having to upgrade the provider binary. Meaning that these expressions are tightly coupled with a particular provider release. If that's undesired, we could parameterize these via the coordinator admin and update them unilaterally.

ethenotethan (approved):
iiuc this is adding a self reporting mechanism to the provider where the provider self-determines its concurrency limits based on constant profiling of token throughput wrt decode rates, GPU availability, memory bandwidth, etc. ran some q&a using agents to understand how variability in prompt sizes could impact the self reporting which is claimed to be handled. also looked at the variables we're profiling and they seem correct for profiling in-time token throughput capabilities. 

@hankbobtheresearchoor approved after reviewing. I pre-embedded him with repo context and an llm-wiki that's enriched with theoretical context. 

have open questions about how this handles all different token generation architectures (e.g, spec-decoding with MTP drafters vs single target model next token prediction) and race conditions but will leave those for another day. otherwise approving on intuition, a few hours of layman analysis, and vibes. 

iiuc there is some synchronization risk since the heartbeats are done on 5 second interval but rabbit hole for another day

PR #186, opened 2026-05-18
fix(payments): try overage charge before clamping in handleComplete

hankbobtheresearchoor:
## What

Cherry-picked the provider.go half of PR #161 onto a clean origin/master base to avoid the 16-file merge conflict tangle.

The overage-charge-before-clamp change is applied cleanly; the TODO comments flag where a follow-up `len(x)/4` → `len(x)` upgrade should land.

## Changes

| File | Change |
|------|--------|
| `coordinator/api/provider.go` | Try consumer overage charge before clamping in `handleComplete` |
| `coordinator/api/consumer.go` | TODO comments for `len(x)/4` → `len(x)` upgrade in `approximateTokenCount` |

## How it works

When `totalCost > pr.ReservedMicroUSD` (e.g. a provider's custom output price exceeds the standard platform rate):

1. **Master (old):** immediately clamp, silently shorting the provider.
2. **This PR:** charge the consumer the `overage = totalCost - reserved` first. The consumer already proved sufficient balance at reservation time. Only clamp if the charge fails (e.g. balance drained mid-flight). The provider gets paid according to the reported total cost.

The clamp stays as circuit breaker against provider over-reporting, but should never fire on honest work.

This replaces #161 for the provider.go portion.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/186"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

### DIFF
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -289,6 +289,10 @@ func approximateTokenCount(v any) int {
 		if x == "" {
 			return 0
 		}
+		// TODO: len(x)/4 underestimates tokens for code, non-English text,
+		// and chat template expansions, which can silently short providers.
+		// Switch to len(x) as a universal upper bound — every BPE tokenizer
+		// starts with one token per byte and can only merge.
 		tokens := len(x) / 4
 		if tokens < 1 {
 			tokens = 1
@@ -299,6 +303,7 @@ func approximateTokenCount(v any) int {
 		if err != nil {
 			return 0
 		}
+		// TODO: len(b)/4 same issue as above; switch to len(b) with the fix.
 		tokens := len(b) / 4
 		if tokens < 1 {
 			tokens = 1
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -291,8 +291,15 @@ func approximateTokenCount(v any) int {
 		}
 		// TODO: len(x)/4 underestimates tokens for code, non-English text,
 		// and chat template expansions, which can silently short providers.
-		// Switch to len(x) as a universal upper bound — every BPE tokenizer
-		// starts with one token per byte and can only merge.
+		// Options, ordered by precision/effort:
+		// 1. len(x) — universal upper bound; every BPE tokenizer starts with
+		//    one token per byte and can only merge. Conservative, trivial.
+		// 2. Model-family ratio — classify the model to identify the
+		//    tokenizer family, then apply an appropriate chars/token ratio
+		//    (e.g. ~3.5 for English LLaMA, ~2.5 for code-heavy inputs).
+		//    Better precision than raw len(x) with low overhead.
+		// 3. Run the actual tokenizer for exact count (most precise,
+		//    highest overhead; worth it when the cost delta matters).
 		tokens := len(x) / 4
 		if tokens < 1 {
 			tokens = 1
@@ -303,7 +310,8 @@ func approximateTokenCount(v any) int {
 		if err != nil {
 			return 0
 		}
-		// TODO: len(b)/4 same issue as above; switch to len(b) with the fix.
+		// TODO: len(b)/4 same issue as above; see the string-branch
+		// comment for the three options (len(b), model-family ratio, tokenizer).
 		tokens := len(b) / 4
 		if tokens < 1 {
 			tokens = 1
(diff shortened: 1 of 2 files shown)


coordinator/api/consumer.go
```diff
@@ -289,6 +289,10 @@ func approximateTokenCount(v any) int {
 		if x == "" {
 			return 0
 		}
+		// TODO: len(x)/4 underestimates tokens for code, non-English text,
+		// and chat template expansions, which can silently short providers.
+		// Switch to len(x) as a universal upper bound — every BPE tokenizer
```
ethenotethan: @hankbobtheresearchoor update with a comment that it could be like the VEX, but we could also run tokenizer-based heuristics where we classify based on the respective model family, noticing the specific tokenizer used, then allocating a ratio accordingly.

hankbobtheresearchoor:
I'll take a look at this.

hankbobtheresearchoor:
I'll take a look at this.

coordinator/api/consumer.go
hankbobtheresearchoor: Done. Updated the TODO in `approximateTokenCount` to list three options ordered by precision/effort:

1. **`len(x)`** — universal upper bound (every BPE tokenizer can only merge, starts at one token per byte)
2. **Model-family ratio** — classify the model to identify the tokenizer family, then apply an appropriate chars/token ratio (e.g. ~3.5 for English LLaMA, ~2.5 for code-heavy inputs). This is the VEX-like approach you suggested.
3. **Actual tokenizer** — exact count, highest overhead

Applied to both the string and `json.Marshal` branches. Commit: `5b787dc`

hankbobtheresearchoor:
Updated the `approximateTokenCount` TODO comments per Ethan's review suggestion.

The TODO now lists **three options** for improving token estimation, ordered by precision/effort:

1. **`len(x)`** — universal upper bound (every BPE tokenizer starts with one token per byte and can only merge). Conservative, trivial.
2. **Model-family ratio** — classify the model to identify the tokenizer family, then apply an appropriate chars/token ratio (e.g. ~3.5 for English LLaMA, ~2.5 for code-heavy). Better precision than raw `len(x)` with low overhead.
3. **Actual tokenizer** — run the real tokenizer for exact count. Most precise, highest overhead; worth it when the cost delta matters.

Applied to both the `string` and `json.Marshal` branches in `approximateTokenCount`.

`5b787dc` docs(consumer): expand token-count TODO with three-tier approach

ethenotethan:
@hankbobtheresearchoor can you also file a linear issue titled `better input token` heuristic as well?

hankbobtheresearchoor:
I'll take a look at this.

hankbobtheresearchoor (commented):
## Review Summary

**Verdict:** Looks good. The overage-charge-before-clamp approach is correct and handles the key edge case cleanly.

### What changed
- `handleComplete` now tries `s.ledger.Charge(consumerKey, overage, requestID)` before falling back to clamping
- Settlement comments updated to describe the three possible cases (under, exact, over)

### What's right
- **Correct ordering**: overage charge → then settlement (refund unused). No double-charge possible — if overage succeeds, `totalCost > reserved` so settlement is a no-op. If it fails, `totalCost = reserved` so settlement is also a no-op.
- **Safe fallback**: clamp still exists as circuit breaker. If consumer balance was drained mid-flight, the provider still gets something.
- **Idempotent**: `msg.RequestID` passed to Charge for dedup.
- **Metrics**: `store.debit.latency_ms` with `op:overage_charge` tracks debit latency regardless of outcome — useful for SLO monitoring.

### Suggestions (non-blocking)
- The `approximateTokenCount` TODO in consumer.go is good context — the underestimation is the root cause of why overages happen. Filed [DAR-39](https://linear.app/darkbloom/issue/DAR-39/better-input-token-heuristic) to track it.

hankbobtheresearchoor:
Done. Created [DAR-39: better input token heuristic](https://linear.app/darkbloom/issue/DAR-39/better-input-token-heuristic).

PR #205, opened 2026-05-24
fix: verify model registry artifact bytes

Gajesh2007:
## Summary

Fixes model registry registration accepting same-size corrupted R2 artifacts.

Previously registration only verified each manifest file with `HEAD` status and `Content-Length`. That could accept an object whose bytes did not match `manifest.files[].sha256`, leaving a promotable `ready` model version that providers would later reject during download.

This PR changes registration-time artifact verification to:

- `GET` each manifest file with bounded 8-worker concurrency.
- Stream the response body through SHA-256 without loading full artifacts into memory.
- Compare streamed SHA-256 against `manifest.files[].sha256`.
- Keep `Content-Length` and actual-byte-count checks.
- Use a response-header timeout instead of a whole-request timeout so large model artifacts can stream for longer than 30 seconds.

## Validation

- `go test ./coordinator/...`
- `git diff --check`

## Notes

This intentionally increases model registration cost because the coordinator now reads all artifact bytes once. Registration is an operator/admin path, and the benefit is catching corrupt or overwritten R2 objects before a model version can be stored/promoted as ready.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/205"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782243062&installation_id=133247490&pr_number=205&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F205&signature=7dd40358dc9931bd67b8a0064922d8d7e022aadbbbb42c32f4a7eb9628d075d9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/coordinator/api/model_registry_handlers.go
+++ b/coordinator/api/model_registry_handlers.go
@@ -99,8 +99,8 @@ func (s *Server) handleRegisterModel(w http.ResponseWriter, r *http.Request) {
 		writeJSON(w, http.StatusBadRequest, errorResponse("invalid_request_error", err.Error()))
 		return
 	}
-	if err := verifyManifestFiles(r.Context(), registryCDNBaseURL(), manifest, s.logger); err != nil {
-		writeJSON(w, http.StatusBadRequest, errorResponse("invalid_request_error", "manifest file verification failed: "+err.Error()))
+	if err := verifyManifestArtifacts(r.Context(), registryCDNBaseURL(), manifest, s.logger); err != nil {
+		writeJSON(w, http.StatusBadRequest, errorResponse("invalid_request_error", "manifest artifact verification failed: "+err.Error()))
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


Gajesh2007:
Closing intentionally. This approach makes the coordinator stream every model artifact during registration to recompute SHA-256, which is not the architecture we want. Full byte verification should happen on the high-bandwidth publish/GCP path; the coordinator should stay on metadata/manifest registration plus cheap availability checks.

PR #248, opened 2026-05-29
fix(provider-swift): map defective chat-template render to 422, not cascading 500 (#242)

### LINKED ISSUE #242: [bug] mlx-community/gemma-4-26b-a4b-it-8bit chat_template is not portable to swift-jinja
### Component

provider-swift (Swift CLI)

### What happened?

When tool definitions are included in calls to /v1/chat/completions to providers using the swift-jinja template interpreter, `500 upper filter requires string` is always thrown.  This impacts all agent harnesses that send tools. This occurs because the chat_template.jinja for this particular model is defective and only succeeds for other services due to undocumented permissiveness in the CPython jinja2 used by oMLX and other backends.

Defect specifics:
- CPython jinja2 is permissive about `Undefined | upper` and propagates `Undefined / ""` through the filter without raising

Patch description:
- throughout chat_template.jinja, replace `X | upper` with `(X | default('')) | upper` to properly treat `missing/None` as the empty string, failing all comparisons without crashing

Darkbloom actions:
- must patch chat_template.jinja
- must re-vend the patched chat_template.jinja from a new R2 prefix
- must bump `aggregate_sha256` in the coordinator catalog
- should consider upstreaming the patch to mlx-community/gemma-4-26b-a4b-it-8bit
- should expand test coverage to include cases that send tool definitions
- should include defensive try/catch in provider-swift's `MultiModelBatchSchedulerEngine` to downgrade the cascading `model load failed` into a clean 4xx error

### Expected behavior

Chat completions for mlx-community/gemma-4-26b-a4b-it-8bit from harnesses that send tool definitions should succeed for Darkbloom as they succeed when this model is loaded locally in oMLX

### Steps to reproduce

set Darkbloom as a provider in any harness that sends tool definitions

### Environment

computers

### Logs

```shell

```

### Confirmations

- [x] I searched existing issues and this isn't a duplicate

shwniscool:
## What

Fixes the `500 upper filter requires string` thrown by `provider-swift` whenever a request carrying **tool definitions** is routed to a model whose `chat_template.jinja` is not portable to swift-jinja — specifically `mlx-community/gemma-4-26b-a4b-it-8bit` (#242).

## Why

The model's `chat_template.jinja` renders `X | upper` against an `Undefined`/`None` value. CPython's jinja2 (used by oMLX and other backends) is permissive and propagates `Undefined`/`""` through the filter, so the template "works" everywhere else. swift-jinja is stricter and raises `upper filter requires string`.

The render happens inside `MultiModelBatchSchedulerEngine.streamChatCompletion` (and `applyTemplate`), which **rethrew the raw error verbatim**. It then fell through `mapInferenceErrorToStatus` to a generic **500**. The coordinator reads a 500 as a provider fault and reroutes the request — so a deterministic, request-shaped failure turns into a cascading `model load failed` across every provider serving the model, impacting all agent harnesses that send tools.

## What this PR changes (provider-side defensive guard — the "should" item in the issue)

- **New typed error** `MultiModelBatchSchedulerEngineError.templateRenderingFailed(String)`.
- **Defensive `try/catch`** around `applyChatTemplate` in both `streamChatCompletion` and `applyTemplate`: any non-typed render failure is wrapped into `.templateRenderingFailed` (the verbatim underlying message is preserved for operator debugging); already-typed engine errors pass through unchanged.
- **Status mapping** `.templateRenderingFailed → 422` in `ProviderLoop+ErrorMapping`. A 422 fails the request cleanly without marking the provider faulty or triggering a reroute.
- **Tests** (`MultiModelBatchSchedulerEngineTests`): the 422 mapping, verbatim-message preservation, and a tokenizer-driven `applyTemplate` test that sends a tool definition and asserts the swift-jinja failure surfaces as `.templateRenderingFailed` → 422 (no live model required).
- CHANGELOG entry under Provider (Swift) → Bug Fixes.

## Remaining items from the issue (NOT in this PR — they are publish/infra, not repo code)

These require artifacts/access outside this repo and the actual re-vended template:

- [ ] **Patch `chat_template.jinja`** — replace `X | upper` with `(X | default('')) | upper` throughout. The template is not in this repo; it ships inside the model snapshot on R2.
- [ ] **Re-vend the patched template from a new R2 prefix.**
- [ ] **Bump `aggregate_sha256` in the coordinator catalog** — this lives in the `model_versions` Postgres table (`r2_prefix` + `aggregate_sha256`), populated at publish time, not in a static repo file. It must be set to the real digest produced by the re-vend.
- [ ] (Optional) Upstream the template patch to `mlx-community/gemma-4-26b-a4b-it-8bit`.

Once the patched template is re-vended, the 422 guard here remains valuable as defense-in-depth against any future non-portable template.

## Verification note

`provider-swift` is a macOS/MLX Swift package and its dependencies (`MLXLMServer`, `mlx-swift-lm`) do not build in this Linux CI sandbox, so I could not run `swift test` here. The new tests are written against the verified upstream type signatures (`ApplyTemplateRequest`, `OpenAITool`, `MLXLMCommon.Tokenizer`) and mirror existing test patterns; please confirm they pass in the macOS test job.

Resolves #242 (provider-side guard).

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/248"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782666937&installation_id=133247490&pr_number=248&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F248&signature=f563baf2549a487494101f2a2750cad4064de311a22b582c78c81853d933f379"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/provider-swift/Sources/ProviderCore/Inference/MultiModelBatchSchedulerEngine.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/MultiModelBatchSchedulerEngine.swift
@@ -168,9 +168,20 @@ public struct MultiModelBatchSchedulerEngine: MLXServerEngine, Sendable {
             promptTokens = try tokenizer.inner.applyChatTemplate(
                 messages: messages, tools: toolSpecs, additionalContext: nil
             )
-        } catch {
+        } catch let error as MultiModelBatchSchedulerEngineError {
             await releaseBox.fire()
             throw error
+        } catch {
+            // #242: a defective `chat_template.jinja` (e.g. gemma-4-26b's
+            // `X | upper` on an Undefined value when tools are present)
+            // throws inside swift-jinja. Wrap it as a typed, request-shaped
+            // 422 instead of letting the raw render error fall through to a
+            // generic 500 that the coordinator misreads as a provider fault
+            // and reroutes into a cascading `model load failed`.
+            await releaseBox.fire()
+            throw MultiModelBatchSchedulerEngineError.templateRenderingFailed(
+                "chat template failed to render request: \(error.localizedDescription)"
+            )
         }
 
         let maxTokens = request.maxTokens ?? defaultMaxTokens
@@ -321,11 +332,23 @@ public struct MultiModelBatchSchedulerEngine: MLXServerEngine, Sendable {
         let tokenizer = try await resolveTokenizer(modelId: request.model)
         let messages = request.messages.map { $0.templateMessageDict() }
         let tools = request.tools?.map { $0.toolSpec() }
-        let tokens = try tokenizer.inner.applyChatTemplate(
-            messages: messages,
-            tools: tools,
-            additionalContext: nil
-        )
+        let tokens: [Int]
+        do {
+            tokens = try tokenizer.inner.applyChatTemplate(
+                messages: messages,
+                tools: tools,
+                additionalContext: nil
+            )
+        } catch let error as MultiModelBatchSchedulerEngineError {
+            throw error
+        } catch {
+            // #242: same defensive wrap as `streamChatCompletion` so a
+            // template that throws on tool definitions surfaces as a clean
+            // 422 from `/apply-template` rather than a generic 500.
+            throw MultiModelBatchSchedulerEngineError.templateRenderingFailed(
+                "chat template failed to render request: \(error.localizedDescription)"
+            )
+        }
         return TokenizeResponse(tokens: tokens)
     }
 
--- a/provider-swift/Sources/ProviderCore/Inference/MultiModelBatchSchedulerEngine.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/MultiModelBatchSchedulerEngine.swift
@@ -168,9 +168,19 @@ public struct MultiModelBatchSchedulerEngine: MLXServerEngine, Sendable {
             promptTokens = try tokenizer.inner.applyChatTemplate(
                 messages: messages, tools: toolSpecs, additionalContext: nil
             )
-        } catch let error as MultiModelBatchSchedulerEngineError {
+        } catch let MultiModelBatchSchedulerEngineError.templateRenderingFailed(message) {
+            // Trust boundary (review #248): only `.templateRenderingFailed`
+            // is passed through unchanged, to avoid double-wrapping its
+            // message. We deliberately do NOT pass through every typed
+            // engine error — the tokenizer is a distinct trust domain from
+            // the scheduler, so scheduler-origin admission errors
+            // (`queueFull`, `tokenBudgetExhausted`, `requestRejected`, ...)
+            // cannot originate here. Any other failure out of this render
+            // block is, by construction, a template problem and is wrapped
+            // below — so a future typed error thrown from the tokenizer
+            // path can't silently slip out with the wrong status code.
             await releaseBox.fire()
-            throw error
+            throw MultiModelBatchSchedulerEngineError.templateRenderingFailed(message)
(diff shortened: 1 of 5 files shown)


hankbobtheresearchoor (approved):
## Review Summary

Clean, well-scoped PR. The root cause is clear (swift-jinja is stricter than CPython jinja2), the fix is defensive in the right place (engine level, before the error hits the coordinator), and the 422 mapping is semantically correct (the request is unprocessable given this model's template, not a transient provider fault).

**Build + test results (macOS, M3 Ultra):**
- `swift build --product darkbloom -c debug` — clean (1674 steps)
- All 3 new unit tests pass (status mapping, message preservation, applyTemplate integration)
- All 3 existing error-map tests still pass
- Full ProviderCore test suite passes

One inoffensive observation below — nothing blocking.

LGTM 🚢

provider-swift/Sources/ProviderCore/Inference/MultiModelBatchSchedulerEngine.swift
```diff
@@ -168,9 +168,20 @@ public struct MultiModelBatchSchedulerEngine: MLXServerEngine, Sendable {
             promptTokens = try tokenizer.inner.applyChatTemplate(
                 messages: messages, tools: toolSpecs, additionalContext: nil
             )
-        } catch {
+        } catch let error as MultiModelBatchSchedulerEngineError {
```
hankbobtheresearchoor: 🔵 **Observation (non-blocking):** This `catch let error as MultiModelBatchSchedulerEngineError` catches *every* typed engine error, not just template-related ones — `queueFull`, `tokenBudgetExhausted`, `requestRejected`, etc. Today that's harmless because `applyChatTemplate` won't throw capacity/congestion errors (those come from the scheduler, not the tokenizer). But a future maintainer adding a new typed engine error that *could* be thrown from tokenization path — e.g. a tokenizer-warming-timeout — would have it silently pass through here and get the wrong status code downstream. Consider narrowing this catch to only `.templateRenderingFailed` (the one case that can actually originate from the tokenizer), or adding a comment noting the trust boundary between tokenizer errors and scheduler errors.

shwniscool: Good call — narrowed in ece91e66. The pass-through now matches only `.templateRenderingFailed` (so its message isn't double-wrapped), and everything else out of the render block is wrapped as `.templateRenderingFailed`. Added a comment at both call sites documenting the tokenizer↔scheduler trust boundary you flagged, so a future typed error on the tokenizer path can't silently slip out with the wrong status.

PR #268, opened 2026-06-03
fix(fleet): reliability fixes — trust, memory routing, billing, SE keys, connectivity

Gajesh2007:
## Summary

Fixes the seven systemic root causes behind the fleet's reliability problems, diagnosed from **96 provider log reports across 40 Macs** (at diagnosis time, **45% of registered providers were untrusted / non-serving**, ~3,700 GB of GPU RAM offline). Each fix is verified against source and ships with a regression test.

| # | Issue | Impact | Where |
|---|-------|--------|-------|
| 1 | Outbound stream dies after reconnect → trust-timeout cascade | ~13 machines pinned untrusted | `CoordinatorClient.swift`, `provider.go` |
| 2 | ACME upgrade leg never retried + stale-trust restore | ~9 stuck `self_signed` | `provider.go`, `registry.go` |
| 3 | Memory admission skipped physical-fit on token-budget path | ~6,400 failed loads / request bounces | `scheduler.go`, `registry.go` |
| 4 | Billing zero-usage: terminal output zeroed observed tokens | direct revenue leak | `BatchScheduler+EngineBridge.swift`, `ProviderLoop.swift`, `provider.go` |
| 5 | SE key loads but can't sign (-25308 / poisoned v2) | silent untrust | `PersistentEnclaveKey.swift` |
| 6 | Reconnect herd + no reachability signal | softens #1 availability driver | `ExponentialBackoff.swift`, `CoordinatorClient.swift` |
| 7 | ECDSA status-sig lockout (cause unconfirmed) | 2 non-recovering | `provider.go`, `attestation.go` |

## Details

**1. Trust-timeout cascade.** The Swift provider reused a single-shot `AsyncStream` for outbound frames; after the first reconnect its iterator was finished, so every challenge response was silently dropped while heartbeats/pongs (which call `ws.send` directly) kept the socket "online" — the box looked online but answered no challenges → `MarkUntrustedTransient`. Fix: per-connection `OutboundRouter` + coordinator force-reconnect after `MaxConsecutiveChallengeTimeoutsBeforeReconnect=6`. Regression: `outboundDeliverySurvivesReconnect`.

**2. ACME retry + stale-trust restore.** `applyACMETrust` ran at registration *before* attestation binding completed, so its key-match check failed on ordering and never re-ran. Now the connect-time ACME result is stashed and retried on each passing challenge (mirroring MDM). `RestoreProviderState` is capped at `self_signed` so hardware trust is re-earned via a live challenge.

**3. Memory admission.** `freeMemoryAdmits` did only token-budget math for modern providers and never checked physical fit, so a 93.7 GB model was dispatched to 48/64 GB boxes that 503'd at load. New always-on `modelFitsHardware()` gate (shared with the pre-warm path so they can't drift), cold-load only (a resident/serving model is never rejected). A never-fits model now reports `model_too_large` (distinct counter) instead of `over_capacity` (which would 429-retry forever).

**4. Billing zero-usage.** `recordFinish` overwrote the live observed token count with the terminal `RequestOutput`, which can report 0 → coordinator bills \$0 and refunds. Now `max(observed, terminal)`; provider falls back to a content-frame floor if the usage chunk is lost entirely; coordinator emits `billing.zero_usage_complete`.

**5. SE key self-test/auto-repair.** `loadOrCreateVerified` runs a self-test sign, attempts one delete + re-mint, and otherwise throws so the provider falls back to an ephemeral identity and keeps serving at `self_signed` instead of being silently pinned untrusted.

**6. Connectivity.** Equal-jitter backoff breaks the synchronized reconnect herd; `NWPathMonitor` reachability is surfaced in reconnect logs + telemetry to separate "coordinator down" from "box lost internet" (the dominant, operator-side cause).

**7. ECDSA status-sig.** Instruments the failure path (plain-sig-passed, Go canonical bytes, per-field lengths, `status_sig_failed` metric) so the two stuck boxes are diagnosable from logs, and hardens the latent escaping divergence — `BuildStatusCanonical` now uses `SetEscapeHTML(false)` so `<`,`>`,`&` match the Swift encoder (no-op for current base64/hex values).

## Testing
- **Coordinator:** `go build ./...` + `go test ./...` green (11 packages). New tests: absolute-fit / `model_too_large` routing, `BuildStatusCanonical` no-HTML-escape golden.
- **Provider:** `swift build` green; new tests pass — `outboundDeliverySurvivesReconnect`, `recordFinishUsesMaxObservedTokens` (billing), jitter-backoff bounds. MLX-runtime tests skipped locally (no metallib in worktree).

## Follow-ups (noted, not in this PR)
- Backfill `TotalSizeBytes` for catalog model versions (the fit gate fails open on `SizeGB==0`).
- Coordinator periodic `conn.Ping` for faster half-open detection (90s sweeper + #1 force-reconnect already cover it).
- Same `SetEscapeHTML(false)` hardening for the plain attestation-blob canonical (not currently failing anyone).
- Apply `NWPathMonitor` to *gate* reconnects (await link-up), not just report.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/268"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783106773&installation_id=133247490&pr_number=268&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F268&signature=9be57398fbc0c8bf6dc648ddb42818abe4de52f3bf8dbc72780bb8652be3138d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/api/adaptive_capacity_integration_test.go
+++ b/coordinator/api/adaptive_capacity_integration_test.go
@@ -285,7 +285,7 @@ func TestAdaptiveCapacityIntegrationOmittedMaxConcurrencyUsesLegacyFallback(t *t
 	for i := range 4 {
 		p.AddPending(&registry.PendingRequest{RequestID: fmt.Sprintf("legacy-%d", i), ProviderID: p.ID, Model: model, RequestedMaxTokens: 128})
 	}
-	candidates, rejections := reg.QuickCapacityCheck(model, 10, 128)
+	candidates, rejections, _ := reg.QuickCapacityCheck(model, 10, 128)
 	if candidates != 1 || rejections != 0 {
 		t.Fatalf("QuickCapacityCheck candidates=%d rejections=%d, want 1/0 with legacy fallback", candidates, rejections)
 	}
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -143,6 +143,12 @@ func (s *Server) refundProviderExtra(pr *registry.PendingRequest) {
 	s.ddIncr("billing.reservation_extra_refunds", []string{"model:" + pr.Model})
 }
 
+// errModelTooLarge is the dispatch error returned when providers serve the
+// requested model but none of them has enough total memory to ever load it.
+// Distinct from "no provider available" so the caller rejects fast instead of
+// queuing for 120s — queueing can't help a model that will never fit.
+const errModelTooLarge = "model too large for any available provider"
+
 // dispatchOneProvider encrypts and sends an inference request to a single
 // provider. It returns the pending request and provider on success, or an
 // error string on failure. The excludeProviders set is updated on failure.
@@ -198,6 +204,11 @@ func (s *Server) dispatchOneProvider(
 
 	provider, decision = s.registry.ReserveProviderEx(model, pr, excludeList()...)
 	if provider == nil {
+		// Providers serve this model but none can physically fit it: don't make
+		// the caller queue/retry for something that will never load.
+		if decision.CandidateCount == 0 && decision.CapacityRejections == 0 && decision.ModelTooLargeRejections > 0 {
+			return nil, nil, decision, errModelTooLarge, http.StatusServiceUnavailable
+		}
 		return nil, nil, decision, "no provider available", http.StatusServiceUnavailable
 	}
 	pendingCleanup := true
@@ -1134,6 +1145,16 @@ func (s *Server) handleChatCompletions(w http.ResponseWriter, r *http.Request) {
 			isResponsesAPI, timing, excludeProviders,
 		)
 		if provider == nil {
+			// No online provider has enough memory to ever fit this model.
+			// Retrying and queueing are both pointless — reject immediately
+			// with a clear, non-retryable error.
+			if dispatchErr == errModelTooLarge {
+				s.ddIncr("routing.decisions", []string{"model:" + model, "model_type:" + s.registry.ModelType(model), "outcome:model_too_large"})
+				lastErr = dispatchErr
+				lastErrCode = dispatchErrCode
+				break
+			}
+
 			// dispatchOneProvider may have found a provider but rejected it
 			// (payout destination missing, insufficient funds, encryption
 			// missing). In that case it already added the provider to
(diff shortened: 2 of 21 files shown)


Gajesh2007:
Addressed the latest review comments in f9f490a5:

- **Codex P2 — memory multiplier too permissive** (`scheduler.go:141`): catalog `SizeGB` is the raw weight (`TotalSizeBytes/1e9`), but the provider needs `weight × 1.2 × 2.0 ≈ 2.4×`. Bumped `modelMemoryHeadroomFactor` 2.0 → **2.4** so we stop dispatching/pre-warming models the provider rejects at load (e.g. a 28 GB-weight model on a 64 GB Mac). Cold-load only; resident (`running`/`idle`) models stay exempt. Regression test added.
- **Codex P2 — fresh attestation clobbered on restore** (`registry.go:620`): `RestoreProviderState` now caps trust at `self_signed` (hardware must be re-earned) **without** forcing `Attested=false`, so a provider that just passed live SE attestation no longer reports `X-Provider-Attested:false`.

Earlier P2s (idle-resident gate exemption, preflight `model_too_large`, empty-delta billing) were fixed in prior commits.

- **Threat-model CI — provider-inflated billing floor**: valid concern, but **not safe to fix with a naive clamp**. `requestedMaxTokens` defaults to 256 when the consumer omits `max_tokens`, so clamping billed completion tokens to it would massively *under-bill* legitimate long generations. This is fundamentally the pre-existing provider-self-reports-usage trust boundary; the right fix is coordinator-side independent token validation (per-model ceiling), which I'm tracking as a separate follow-up rather than rushing an under-billing bug into this PR.

PR #280, opened 2026-06-06
feat(provider): revision-aware model update with background download

ethenotethan:
## What

Adds the ability to **download a newer model revision in the background and replace the cached copy** (Slack request).

### New: `darkbloom models update [<model-id>]`
Compares each installed model's recorded revision against the coordinator catalog; when a model's revision changed, downloads the new revision into a staging dir and **atomically swaps it in** (the manifest path keeps the old snapshot serving until the swap completes). Flags:
- `--background` — runs the update in a detached child process, logging to `~/.darkbloom/model-update.log` so it survives the shell exiting.
- `--include-untracked` — also refresh installed models that predate revision tracking (off by default, so the first run doesn't re-download the whole cache).
- `--coordinator` / `--r2-cdn` overrides, mirroring `models download`.

With no `<model-id>`, every installed catalog model is checked. Not-installed models are skipped (this updates existing copies; use `models download` to install).

### Revision tracking
- After every download, persist a per-model `.darkbloom-revision.json` (`version` + `aggregate_sha256`) next to the HF `snapshots/`/`refs/` dirs.
- `ModelDownloader.revisionStatus(for:)` compares it to the catalog — prefers the content hash (catches re-publishes that reuse a version string), falls back to the version string.
- New reusable API: `localRevision(for:)`, `revisionStatus(for:)`, `downloadIfOutdated(model:replaceUntracked:onProgress:)`.

### UX
- `models catalog` now flags entries that have an update available.

## Tests
- Unit tests for the pure revision comparison (`compareRevision`) and the revision-file read/write round-trip in `ModelCatalogTests`.

## Design notes
- Deliberately scoped to an **operator-initiated CLI**, not an in-process hot-swap inside the running attested provider — swapping attested weights at runtime has attestation implications and was left out of scope.
- `aggregate_sha256` is the primary revision key; `version` is the fallback.

## ⚠️ Not built here
`provider-swift` is macOS-only (Darwin / MLX / Metal) and was **not compiled** in the Linux authoring sandbox (no Swift toolchain). CI / a macOS build must verify compilation and run the tests.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/280"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783361513&installation_id=133247490&pr_number=280&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F280&signature=e36f292bdf2eddc95c51536591c478bf77723af0f7ce9ba10d46ffc17d942d4b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/provider-swift/Sources/ProviderCore/Models/ModelCatalog.swift
+++ b/provider-swift/Sources/ProviderCore/Models/ModelCatalog.swift
@@ -558,6 +558,53 @@ public struct ModelCatalogClient: Sendable {
     }
 }
 
+// MARK: - Revision tracking
+
+/// Revision record persisted alongside a downloaded model so a later run can
+/// tell whether the cached copy matches the catalog's current revision without
+/// re-hashing multi-GB weights. Written to
+/// `~/.cache/huggingface/hub/models--{org}--{name}/.darkbloom-revision.json`.
+public struct LocalRevision: Codable, Sendable, Equatable {
+    public let modelID: String
+    public let version: String?
+    public let aggregateSHA256: String?
+
+    public init(modelID: String, version: String?, aggregateSHA256: String?) {
+        self.modelID = modelID
+        self.version = version
+        self.aggregateSHA256 = aggregateSHA256
+    }
+
+    enum CodingKeys: String, CodingKey {
+        case modelID = "model_id"
+        case version
+        case aggregateSHA256 = "aggregate_sha256"
+    }
+}
+
+/// How the cached copy of a model compares to the catalog's current revision.
+public enum RevisionStatus: Sendable, Equatable {
+    /// No copy of the model is present on disk.
+    case notInstalled
+    /// Cached revision matches the catalog.
+    case upToDate
+    /// Installed, but with no recorded revision (predates revision tracking)
+    /// or insufficient revision info on one side to compare.
+    case unknown
+    /// Cached revision differs from the catalog's current revision.
+    case outdated(local: String?, catalog: String?)
+}
+
+/// Result of a revision-aware download (`downloadIfOutdated`).
+public enum UpdateOutcome: Sendable, Equatable {
+    case notInstalled
+    case upToDate
+    /// Installed but with no recorded revision, and `replaceUntracked` was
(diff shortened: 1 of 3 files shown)


Gajesh2007:
Closing as superseded by the manifest/prefetch architecture that landed after this was authored (#283 + the registry manifest work):

- The downloader is now revision-aware by construction: each file is skip-if-valid against the CURRENT catalog manifest, so `darkbloom models update`'s job is already what `darkbloom models download <id>` does on a re-published model — it fetches exactly the changed files and re-verifies per-file + aggregate SHA-256. No separate `.darkbloom-revision.json` tracking is needed; the manifest hashes are the revision key.
- For connected providers, the coordinator-driven path (desired_models → background prefetch_model → verified-on-disk → hot-swap) covers fleet model updates without operator action, which was the Slack ask's underlying need.
- The `--background` detached child process would add a second uncoordinated writer to the model cache underneath a serving, attested provider — that interacts badly with the weight-hash refresh/TOCTOU logic from #297.

Worth salvaging as small follow-ups on top of the manifest hashes (not stamp files): the `models catalog` 'update available' flag, and a `models update` convenience verb that just loops installed models through the existing skip-if-valid downloader. Thanks — the UX instinct here was right; the mechanism just got built underneath it in the meantime.

PR #237, opened 2026-05-29
feat(kv-cache): encrypted SSD KV cache — primitives, verification, and flag-gated engine integration

anupsv:
## Summary

Encrypted-at-rest KV cache for the Swift provider: persist prefill KV to
disk encrypted, reload after restart/eviction to skip re-prefill. 12
commits, ~3.5k LOC, **103 KV-cache tests green**.

> **DRAFT — two gates before merge:**
> 1. **Submodule PR must merge first.** This branch bumps
>    `libs/mlx-swift-lm` to a branch (`feat/encrypted-prefix-cache-persistence`)
>    adding the additive `PrefixCachePersistence` hook. That submodule PR
>    has to land before this can go to master.
> 2. **2-Mac smoke test pending.** End-to-end behavior with the flag ON
>    can't run in CI — needs the M5 Thunderbolt pair. All unit/integration
>    coverage here runs on `swift test`.

## What's in it

**Crypto primitives (P0)** — `EncryptedKVStore` (AES-256-GCM, per-file
random DEK, HKDF-derived per-chunk nonces, metadata-as-AAD tamper
binding, atomic write + dir fsync) and `KVCacheKEK` (envelope: per-file
DEK wrapped by an SE-derived KEK held in Keychain). Async + sync
(`writeSync`/`readSync`) paths share the format.

**Cache machinery (P1–P3)** — `PrefixCacheRAM` (LRU), `PrefixDigest` +
`PrefixCacheIndex` (exact-checkpoint lookup), `PrefixCacheManager`
(orchestration actor with the MB-1 model-binding guard), and
`KVCacheSerializer` (`[KVCache]` ↔ encryptable bytes, bf16-exact).

**Live integration (Path 2)** — backs the **engine's own** in-GPU block
prefix cache with our encryption: a `PrefixCachePersistence` hook
(submodule) calls `EncryptedPrefixCachePersistence` on block evict/load,
so evicted blocks are AES-GCM-encrypted to SSD and reloaded instead of
re-prefilled. Gated behind the default-off `DARKBLOOM_PREFIX_CACHE` flag.

## Verified before building

- Cross-model cacheability + the rotating-cache snapshot/restore hazard
  resolved **empirically** (a review claimed `temporalOrder` scrambles on
  restore; tests prove it's correct when state+metaState round-trip
  together).
- Qwen3.5/3.6/3.7 MoE + Gemma-4 26B-A4B + GPT-OSS-20B confirmed from
[...]

PR #290, opened 2026-06-09
feat(kv-cache): sliding 5-minute SSD TTL + per-server --local cache scope

anupsv:
Two follow-ups to the merged encrypted SSD KV cache (#237), both from @Gajesh2007's review comments. Clean branch off `master` — **8 files, +247/−5** (the prompt_cache_key scoping landed with #237 already; this is the remaining work).

## 1. Sliding 5-minute SSD TTL

`expiresAt` was carried in the metadata but **never enforced** — checkpoint lifetime was purely capacity-driven, so a low-traffic prefix could sit encrypted on SSD indefinitely. This adds a real **sliding** TTL:

- **Sliding off `lastHitAt`** (the mutable, plaintext index field bumped on every SSD hit) — so a hit refreshes the window **without re-sealing** the GCM-AAD-bound file. Rule: expired when `now() - lastHitAt > ttl`.
- **Lazy enforcement**: `loadFromSSD` checks expiry *before* decrypt → expired = miss + drop file & index entry + refresh accountant (`ttlExpirations++`).
- **Proactive reclaim**: `reapExpired()` at load-time `reconcileWithDisk` reclaims expired entries even for a model never looked up again (covers both the indexed-entry and orphan-file paths).
- `expiresAt = createdAt + ttl` is now stamped on write (a conservative hard floor for the orphan check); `nil` when disabled.
- **Config**: `DARKBLOOM_PREFIX_CACHE_TTL_SECONDS` — default **300** (matches Anthropic/OpenAI prompt-cache defaults), `0` = infinite.
- **Scope**: SSD checkpoint tier only — the RAM tier keeps its deterministic LRU.
- **Privacy**: bounds how long prompt-derived KV lingers on disk, shrinking the TB-007 cross-tenant TTFT-oracle window.

## 2. Per-server cache scope for the `--local` path

The standalone HTTP server can't carry a per-request `prompt_cache_key` (the upstream `OpenAIChatCompletionRequest` has no such field), so the per-tenant scoping merged in #237 only flowed on the coordinator WebSocket path. This adds a **fixed per-server** scope via `DARKBLOOM_PREFIX_CACHE_SCOPE` (`""` = unscoped, default) so cross-tenant isolation can be validated on a single box.

## Tests
- Unit (deterministic, injectable clock): TTL expiry → miss+drop; sliding-keeps-hot-alive (5×200s hits over a 300s TTL never expire); `ttl=0` → infinite; reconcile reaps (orphan + indexed paths); pure `resolveTTLSeconds` policy. All cache suites green; build clean on top of current `master`.

## Verified live on an M5 Max
- **TTL**: real 300s default — populate 8 SSD checkpoints, idle 330s, restart → reconcile reaped all 8 (**8 → 0**). **Control** with `TTL=0` over the identical idle kept all 8 (**8 → 8**) — proving the reclamation is TTL-driven, not some other cleanup.
- **Scoping**: two scopes sharing one cache dir, identical prompts → file count **24 → 48** (tenant B could not hit tenant A's cached prefixes; same-scope reuse works).

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/290"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783558822&installation_id=133247490&pr_number=290&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F290&signature=4b20ec1b7c8019630a243d29f20c910be122902c5a7cb15f38109d40f75eb9af"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/provider-swift/Sources/ProviderCore/Inference/BatchScheduler+EngineBridge.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/BatchScheduler+EngineBridge.swift
@@ -282,6 +282,27 @@ extension BatchScheduler {
     /// the interval is 0 (disabled) or there is no checkpoint manager (engine
     /// tier / cache off). Snapshots are cheap (a struct copy across the manager
     /// actor), so a 2-minute cadence is free.
+    /// Periodic steady-state TTL sweep (PR #290 review). Cadence: half the
+    /// TTL, clamped to [60, 300]s — frequent enough that an expired entry
+    /// outlives its TTL by at most ~cadence, rare enough to be free (the sweep
+    /// is an in-memory index scan; file deletes only when something expired).
+    /// Independent of the stats logger so disabling observability never
+    /// disables the privacy reaper.
+    func startTTLReaper() {
+        ttlReapTask?.cancel()
+        ttlReapTask = nil
+        let ttl = Self.prefixCacheTTLSeconds()
+        guard ttl > 0, let mgr = checkpointManager else { return }
+        let cadence = Swift.max(60, Swift.min(ttl / 2, 300))
+        ttlReapTask = Task.detached { [weak mgr] in
+            while !Task.isCancelled {
+                try? await Task.sleep(for: .seconds(Double(cadence)))
+                if Task.isCancelled { return }
+                await mgr?.reapExpiredTick()
+            }
+        }
+    }
+
     func startPrefixCacheStatsLogger() {
         prefixCacheStatsTask?.cancel()
         prefixCacheStatsTask = nil
--- a/provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
@@ -596,6 +596,9 @@ public actor BatchScheduler {
                         // TB-016 sub-feature B: min persist threshold (16384
                         // for Gemma, 0 otherwise). Env override available.
                         minPersistTokens: Self.prefixCacheMinPersistTokens(arch: modelId),
+                        // Sliding SSD TTL (default 5min; 0 = infinite). Bounds
+                        // how long prompt-derived KV lingers on disk.
+                        ttlSeconds: Self.prefixCacheTTLSeconds(),
                         now: nowFn,
                         accountant: diskAccountant,
                         modelKey: backing.modelKey)
@@ -940,6 +943,23 @@ public actor BatchScheduler {
         }
         // Default: 16384 for Gemma, 0 otherwise.
         return PrefixCachePastWindow.isProven(arch: arch) ? 16384 : 0
+    }
+
+    /// Sliding TTL (seconds) for persisted SSD prefix checkpoints. Default 300
+    /// (5 min, matching Anthropic/OpenAI prompt-cache defaults); `0` disables
+    /// (infinite — capacity-driven eviction only). Operator override:
+    /// `DARKBLOOM_PREFIX_CACHE_TTL_SECONDS`.
+    static let defaultPrefixCacheTTLSeconds: Int64 = 300
+    static func prefixCacheTTLSeconds() -> Int64 {
+        resolveTTLSeconds(env: ProcessInfo.processInfo.environment["DARKBLOOM_PREFIX_CACHE_TTL_SECONDS"])
+    }
+
+    /// Pure TTL policy (testable). Unset / malformed / negative ⇒ default;
+    /// `0` ⇒ disabled (infinite); a positive value sets the sliding window.
+    static func resolveTTLSeconds(env: String?) -> Int64 {
+        guard let v = env else { return defaultPrefixCacheTTLSeconds }
+        guard let n = Int64(v), n >= 0 else { return defaultPrefixCacheTTLSeconds }
+        return n  // 0 ⇒ disabled
     }
 
     /// Set the post-load budgets driven by architecture + physical
--- a/provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
@@ -103,6 +103,12 @@ public actor BatchScheduler {
     /// only — the engine tier (`EncryptedPrefixCachePersistence`) keeps no
     /// hit/miss counters, so pure-attention `.engine` models log nothing here.
     var prefixCacheStatsTask: Task<Void, Never>?
+    /// Steady-state TTL reaper (PR #290 review): reapExpired otherwise runs
+    /// only at load-time reconcile, so entries going cold while the model
+    /// stays loaded would sit on disk until restart (the lazy read-path check
+    /// fires only when the same prefix is looked up again). Started with the
+    /// engine, cancelled in `stopCurrentEngine`.
+    var ttlReapTask: Task<Void, Never>?
     /// Interval (seconds) for the stats logger. Default 120s when a checkpoint
     /// manager is active (one info line every two minutes is negligible even
     /// across a fleet, and gives hit-rate observability out of the box). A
@@ -330,6 +336,9 @@ public actor BatchScheduler {
         // Periodic checkpoint-tier hit/miss logger (no-op if disabled or
         // engine-tier model). Cancelled in stopCurrentEngine.
         startPrefixCacheStatsLogger()
+        // Steady-state TTL sweep for the checkpoint SSD tier (no-op when TTL
+        // disabled or engine-tier model). Cancelled in stopCurrentEngine.
+        startTTLReaper()
     }
 
     /// Snapshot model bytes + tokenizer + architecture out of the
@@ -1371,6 +1380,8 @@ public actor BatchScheduler {
         await logPrefixCacheStats()
         prefixCacheStatsTask?.cancel()
         prefixCacheStatsTask = nil
+        ttlReapTask?.cancel()
+        ttlReapTask = nil
 
         if let engine = self.engine {
             _ = engine.core.abortAllRequests()
--- a/provider-swift/Sources/ProviderCore/Inference/MultiModelBatchSchedulerEngine.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/MultiModelBatchSchedulerEngine.swift
@@ -122,7 +122,8 @@ public struct MultiModelBatchSchedulerEngine: MLXServerEngine, Sendable {
         acquire: @escaping @Sendable (String) async throws -> AcquiredModel,
         tokenizerProvider: @escaping @Sendable (String?) async throws -> TokenizerHandle,
         availableModels: @escaping @Sendable () async -> [String],
-        defaultMaxTokens: Int = 4096
+        defaultMaxTokens: Int = 4096,
+        cacheScope: String = ""
     ) {
         self.acquire = acquire
         self.tokenizerProvider = tokenizerProvider
@@ -133,7 +134,11 @@ public struct MultiModelBatchSchedulerEngine: MLXServerEngine, Sendable {
         self.releaseModel = { _ in }
         self.defaultMaxTokens = defaultMaxTokens
         self.reasoningEffort = nil
-        self.cacheScope = ""
+        // Fixed per-server scope for the standalone (--local) path, which can't
+        // carry a per-request prompt_cache_key through the upstream router. ""
+        // ⇒ unscoped (default). Set via DARKBLOOM_PREFIX_CACHE_SCOPE; used to
+        // exercise/validate cross-tenant isolation on a single box.
+        self.cacheScope = cacheScope
     }
 
     // MARK: - MLXServerEngine
--- a/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
+++ b/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
@@ -468,9 +468,45 @@ public actor PrefixCacheManager: PrefixCacheOwner {
 
     private func loadFromSSD(tokens: [Int], scope: String = "") async -> PrefixLookupResult? {
         guard let index, let kek, let cacheDir else { return nil }
-        guard let entry = index.findLongestCheckpoint(
+
+        // Sliding TTL: an entry is expired when it hasn't been hit within
+        // `ttlSeconds` (lastHitAt is bumped on every hit — SSD loads AND RAM
+        // serves — so hot prefixes stay warm). Checked BEFORE decrypt — an
+        // expired entry is reclaimed (same drop path as a corrupt file) and
+        // the search CONTINUES with the next-longest checkpoint: an expired 8k
+        // checkpoint must not mask a shorter, still-fresh one (e.g. a hot
+        // shared system-prefix). Each drop removes the entry from the index,
+        // so findLongestCheckpoint yields the next candidate; progress is
+        // guaranteed. ttlSeconds == 0 disables (infinite retention). Bounds
+        // how long prompt-derived KV survives on disk (TB-007 window shrink).
+        var selected: PrefixIndexEntry?
+        // Defensive cap: dropUnusableSSDFile no-ops when the manager closed
+        // mid-loop (it must not delete in a handed-off dir), which would
+        // otherwise refetch the same entry forever. The `closed` re-check
+        // breaks that cycle; the cap is belt-and-braces against any future
+        // drop path that leaves the entry behind.
+        var attempts = boundaries.count + 1
+        while attempts > 0, !closed, let candidate = index.findLongestCheckpoint(
             modelHash: binding.modelHash, tokens: tokens, boundaries: boundaries, scope: scope
-        ) else { return nil }
+        ) {
+            attempts -= 1
+            if ttlSeconds > 0 {
+                // Overflow-safe age: index.json is plaintext and can be left
+                // corrupt by a crash — an extreme lastHitAt (e.g. Int64.min)
+                // must be treated as expired/dropped, not trap the provider.
+                let (age, overflow) = now().subtractingReportingOverflow(candidate.lastHitAt)
+                if overflow || age > ttlSeconds {
+                    stats.ttlExpirations += 1
+                    let rel = "\(modelDirComponent)/\(candidate.digestHex).\(EncryptedKVStore.fileExtension)"
+                    await dropUnusableSSDFile(
+                        cacheDir.appendingPathComponent(rel), digestHex: candidate.digestHex, index: index)
+                    continue
+                }
+            }
+            selected = candidate
+            break
+        }
+        guard let entry = selected else { return nil }
 
         // Path safety: the on-disk index JSON is plaintext and NOT
         // authenticated, so a tampered entry.relativePath could contain
@@ -481,18 +517,6 @@ public actor PrefixCacheManager: PrefixCacheOwner {
         // digest) instead of trusting the stored path.
         let relPath = "\(modelDirComponent)/\(entry.digestHex).\(EncryptedKVStore.fileExtension)"
         let fileURL = cacheDir.appendingPathComponent(relPath)
-
-        // Sliding TTL: an entry is expired when it hasn't been hit within
-        // `ttlSeconds` (lastHitAt is bumped on every hit, so hot prefixes stay
-        // warm). Checked BEFORE decrypt — an expired entry is a miss, and its
-        // file + index entry are reclaimed (same drop path as a corrupt file).
-        // ttlSeconds == 0 disables (infinite retention). Bounds how long
-        // prompt-derived KV survives on disk (TB-007 oracle-window shrink).
-        if ttlSeconds > 0, now() - entry.lastHitAt > ttlSeconds {
-            stats.ttlExpirations += 1
-            await dropUnusableSSDFile(fileURL, digestHex: entry.digestHex, index: index)
-            return nil
-        }
 
         // MB-1: validate metadata BEFORE unwrap/decrypt. A wrong-model
         // file decrypts cleanly (AAD is its own metadata), so the cipher
--- a/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
+++ b/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
@@ -308,6 +308,19 @@ public actor PrefixCacheManager: PrefixCacheOwner {
         return overflow ? Int64.max : sum
     }
 
+    /// THE sliding-TTL expiry predicate — single source of truth for the read
+    /// path (loadFromSSD) and the sweeps (reapExpired). Expired when the entry
+    /// hasn't been hit within `ttlSeconds`; `lastHitAt` is bumped on every hit
+    /// (SSD loads AND RAM serves), so hot prefixes stay warm. Overflow-safe:
+    /// index.json is plaintext and can be crash-corrupted — an extreme
+    /// `lastHitAt` (e.g. Int64.min) counts as expired/reapable, never a trap.
+    /// `ttlSeconds == 0` ⇒ never expires (infinite retention).
+    private func isExpired(lastHitAt: Int64) -> Bool {
+        guard ttlSeconds > 0 else { return false }
+        let (age, overflow) = now().subtractingReportingOverflow(lastHitAt)
+        return overflow || age > ttlSeconds
+    }
+
     /// Time-coalesced index save after a hit-recency touch (PR #290 review).
     /// `index.touch` only mutates in memory; if the process dies before a
     /// graceful flush, restart reconcile would reap recently-hot entries off
@@ -535,21 +548,14 @@ public actor PrefixCacheManager: PrefixCacheOwner {
             modelHash: binding.modelHash, tokens: tokens, boundaries: boundaries, scope: scope
         ) {
             attempts -= 1
-            if ttlSeconds > 0 {
-                // Overflow-safe age: index.json is plaintext and can be left
-                // corrupt by a crash — an extreme lastHitAt (e.g. Int64.min)
-                // must be treated as expired/dropped, not trap the provider.
-                let (age, overflow) = now().subtractingReportingOverflow(candidate.lastHitAt)
-                if overflow || age > ttlSeconds {
-                    stats.ttlExpirations += 1
-                    let rel = "\(modelDirComponent)/\(candidate.digestHex).\(EncryptedKVStore.fileExtension)"
-                    await dropUnusableSSDFile(
-                        cacheDir.appendingPathComponent(rel), digestHex: candidate.digestHex, index: index)
-                    continue
-                }
+            guard isExpired(lastHitAt: candidate.lastHitAt) else {
+                selected = candidate
+                break
             }
-            selected = candidate
-            break
+            stats.ttlExpirations += 1
+            let rel = "\(modelDirComponent)/\(candidate.digestHex).\(EncryptedKVStore.fileExtension)"
+            await dropUnusableSSDFile(
+                cacheDir.appendingPathComponent(rel), digestHex: candidate.digestHex, index: index)
         }
         guard let entry = selected else { return nil }
 
@@ -1093,11 +1099,7 @@ public actor PrefixCacheManager: PrefixCacheOwner {
     /// manager is closed (a closed manager must not mutate a handed-off dir).
     private func reapExpired(index: PrefixCacheIndex, cacheDir: URL) {
         guard !closed, ttlSeconds > 0 else { return }
-        // Saturating: an operator-set "effectively infinite" TTL (e.g.
-        // Int64.max) must mean "reap nothing", not trap on underflow.
-        let (diff, underflow) = now().subtractingReportingOverflow(ttlSeconds)
-        let cutoff = underflow ? Int64.min : diff
-        for entry in index.entries(modelHash: binding.modelHash) where entry.lastHitAt < cutoff {
+        for entry in index.entries(modelHash: binding.modelHash) where isExpired(lastHitAt: entry.lastHitAt) {
             let url = cacheDir.appendingPathComponent(
                 "\(modelDirComponent)/\(entry.digestHex).\(EncryptedKVStore.fileExtension)")
             try? FileManager.default.removeItem(at: url)
--- a/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
+++ b/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
@@ -216,6 +216,16 @@ public actor PrefixCacheManager: PrefixCacheOwner {
     private var unsavedWrites = 0
     /// Save the index after this many new writes (or on shutdown/idle flush).
     private static let saveCoalesceThreshold = 8
+    /// Last wall-clock second a touch-driven index save ran (PR #290 review):
+    /// hit-recency (`lastHitAt`) is bumped in memory on every RAM/SSD hit, but
+    /// without periodic persistence a crash loses those bumps and the next
+    /// restart's reconcile reaps recently-hot entries off stale timestamps.
+    /// Saves are time-coalesced (at most one per `touchSaveIntervalSeconds`),
+    /// bounding both the hot-path fsync cost and the crash-loss window.
+    private var lastTouchSaveAt: Int64 = 0
+    /// Max recency staleness a crash can cause (seconds). 60s against the 300s
+    /// default TTL means a crash under-credits an entry's recency by ≤ 20%.
+    static let touchSaveIntervalSeconds: Int64 = 60
     /// TB-016 sub-feature B: pinned digests are always eligible for
     /// persistence regardless of minPersistTokens. Internal-only (no
     /// public pin() API this phase).
@@ -296,6 +306,40 @@ public actor PrefixCacheManager: PrefixCacheOwner {
     private func saturatingExpiry(from base: Int64) -> Int64 {
         let (sum, overflow) = base.addingReportingOverflow(ttlSeconds)
         return overflow ? Int64.max : sum
+    }
+
+    /// Time-coalesced index save after a hit-recency touch (PR #290 review).
+    /// `index.touch` only mutates in memory; if the process dies before a
+    /// graceful flush, restart reconcile would reap recently-hot entries off
+    /// stale persisted timestamps. At most one save per
+    /// `touchSaveIntervalSeconds` bounds the crash-loss window without putting
+    /// an fsync on every warm request. TTL-gated: with TTL off nothing reaps
+    /// on recency, so the persistence urgency disappears.
+    private func persistRecencyIfDue(_ index: PrefixCacheIndex) {
+        guard ttlSeconds > 0, !closed, index.isDirty else { return }
+        let t = now()
+        guard t - lastTouchSaveAt >= Self.touchSaveIntervalSeconds else { return }
+        if (try? index.save()) != nil {
+            lastTouchSaveAt = t
+            unsavedWrites = 0
+        }
+    }
+
+    /// Steady-state TTL sweep (PR #290 review): reconcile-time reaping only
+    /// covers model (re)load, and the lazy read-path check only fires when the
+    /// SAME prefix is looked up again — so entries that go cold while the
+    /// model stays loaded would otherwise sit on disk until restart. Called
+    /// periodically by BatchScheduler while the engine is alive. Persists the
+    /// shrunken index and pushes the corrected footprint to the accountant
+    /// (mirroring dropUnusableSSDFile). No-op when TTL disabled / SSD off /
+    /// closed.
+    public func reapExpiredTick() async {
+        guard ssdEnabled, let index, let cacheDir else { return }
+        let before = stats.ttlExpirations
+        reapExpired(index: index, cacheDir: cacheDir)
+        guard stats.ttlExpirations > before else { return }
+        if index.isDirty { _ = try? index.save() }
+        await notifyAccountant()
     }
 
     public init(
@@ -409,6 +453,7 @@ public actor PrefixCacheManager: PrefixCacheOwner {
                         // enabled so ttl=0 behavior stays byte-identical.
                         if ttlSeconds > 0 {
                             index.touch(modelHash: binding.modelHash, digestHex: digestHex, now: now())
+                            persistRecencyIfDue(index)
                         }
                     } else if hit.tokenCount >= minPersistTokens {
                         // TB-016 sub-feature B: 2nd-use promotion. RAM hit above
@@ -618,6 +663,7 @@ public actor PrefixCacheManager: PrefixCacheOwner {
             )
         }
         index.touch(modelHash: binding.modelHash, digestHex: entry.digestHex, now: now())
+        persistRecencyIfDue(index)
 
         return PrefixLookupResult(caches: caches, tokenCount: entry.tokenCount, tier: .ssd)
     }
--- a/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
+++ b/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
@@ -390,16 +390,27 @@ public actor PrefixCacheManager: PrefixCacheOwner {
         for cp in checkpoints.reversed() {
             if let hit = ram.get(modelHash: binding.modelHash, digest: cp.digest) {
                 stats.ramHits += 1
-                // TB-016 sub-feature B: 2nd-use promotion. If this RAM hit is
-                // above the persist threshold and NOT already on SSD, schedule
-                // a detached promotion (no blocking the lookup actor).
                 let digestHex = cp.digest.dbkvHexString
-                if ssdEnabled,
-                   hit.tokenCount >= minPersistTokens,
-                   index?.entry(modelHash: binding.modelHash, digestHex: digestHex) == nil {
-                    let cpScope = scope
-                    Task.detached { [weak self] in
-                        await self?.persistDigest(cp.digest, scope: cpScope)
+                if ssdEnabled, let index {
+                    if index.entry(modelHash: binding.modelHash, digestHex: digestHex) != nil {
+                        // The prefix is ALSO on SSD. A RAM hit must slide the
+                        // SSD entry's lastHitAt: the sliding TTL is "time since
+                        // last use", and use includes RAM serves. Without this,
+                        // a RAM-hot prefix older than the TTL gets reaped as
+                        // "expired" the moment RAM pressure evicts it and the
+                        // next lookup falls through to SSD. Gated on TTL being
+                        // enabled so ttl=0 behavior stays byte-identical.
+                        if ttlSeconds > 0 {
+                            index.touch(modelHash: binding.modelHash, digestHex: digestHex, now: now())
+                        }
+                    } else if hit.tokenCount >= minPersistTokens {
+                        // TB-016 sub-feature B: 2nd-use promotion. RAM hit above
+                        // the persist threshold and NOT already on SSD — schedule
+                        // a detached promotion (no blocking the lookup actor).
+                        let cpScope = scope
+                        Task.detached { [weak self] in
+                            await self?.persistDigest(cp.digest, scope: cpScope)
+                        }
                     }
                 }
                 return PrefixLookupResult(caches: hit.caches, tokenCount: hit.tokenCount, tier: .ram)
--- a/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
+++ b/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
@@ -1047,7 +1047,10 @@ public actor PrefixCacheManager: PrefixCacheOwner {
     /// manager is closed (a closed manager must not mutate a handed-off dir).
     private func reapExpired(index: PrefixCacheIndex, cacheDir: URL) {
         guard !closed, ttlSeconds > 0 else { return }
-        let cutoff = now() - ttlSeconds
+        // Saturating: an operator-set "effectively infinite" TTL (e.g.
+        // Int64.max) must mean "reap nothing", not trap on underflow.
+        let (diff, underflow) = now().subtractingReportingOverflow(ttlSeconds)
+        let cutoff = underflow ? Int64.min : diff
         for entry in index.entries(modelHash: binding.modelHash) where entry.lastHitAt < cutoff {
             let url = cacheDir.appendingPathComponent(
                 "\(modelDirComponent)/\(entry.digestHex).\(EncryptedKVStore.fileExtension)")
--- a/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
+++ b/provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
@@ -124,6 +124,8 @@ public struct PrefixCacheManagerStats: Sendable, Equatable {
     public var prefixHashMismatches = 0
     public var ssdReadErrors = 0
     public var diskEvictions = 0
+    /// SSD entries dropped because they passed the sliding TTL (expired).
+    public var ttlExpirations = 0
 }
 
 // MARK: - Manager
@@ -154,6 +156,14 @@ public actor PrefixCacheManager: PrefixCacheOwner {
     /// Half-life (seconds) for recency decay in benefit-per-byte scoring.
     /// TB-016 sub-feature C: freshly-promoted entries stay hot.
     private let evictionHalfLifeSeconds: Double
+    /// Sliding TTL (seconds) for persisted SSD checkpoints. An entry is expired
+    /// when `now() - lastHitAt > ttlSeconds` (sliding: `lastHitAt` is bumped on
+    /// every hit, so hot prefixes stay warm). 0 ⇒ no TTL (infinite — legacy /
+    /// capacity-driven only). Enforced on read (loadFromSSD) and proactively
+    /// reclaimed in the disk sweep. Privacy: bounds how long KV derived from one
+    /// prompt lingers on disk, shrinking the TB-007 cross-tenant TTFT-oracle
+    /// window. SSD tier only — the RAM tier keeps its deterministic LRU.
+    private let ttlSeconds: Int64
     private let now: @Sendable () -> Int64
 
     /// Phase 3: global disk accountant (process-wide, shared across models).
@@ -270,6 +280,17 @@ public actor PrefixCacheManager: PrefixCacheOwner {
         scopeByDigest[digestHex] ?? ""
     }
 
+    /// Absolute expiry stamped into a freshly-written file's metadata:
+    /// `createdAt + ttl`, or nil when TTL is disabled. This is the EARLIEST
+    /// possible expiry (a never-hit entry, where lastHitAt == createdAt); the
+    /// live check slides off the mutable `lastHitAt` in the index, so a hit
+    /// extends effective lifetime without re-sealing the file. Informational +
+    /// a hard floor for any future offline reaper; the authoritative runtime
+    /// check is the lastHitAt comparison in loadFromSSD/sweep.
+    private func expiresAtForWrite() -> Int64? {
+        ttlSeconds > 0 ? now() + ttlSeconds : nil
+    }
+
     public init(
         binding: PrefixCacheModelBinding,
         ram: PrefixCacheRAM,
@@ -282,6 +303,7 @@ public actor PrefixCacheManager: PrefixCacheOwner {
         minPersistTokens: Int = 0,
         prefillCostPerToken: Double = 1.0,
         evictionHalfLifeSeconds: Double = 86400,
+        ttlSeconds: Int64 = 0,
         now: @escaping @Sendable () -> Int64 = { Int64(Date().timeIntervalSince1970) },
         accountant: GlobalDiskAccountant? = nil,
         modelKey: String = ""
@@ -300,6 +322,7 @@ public actor PrefixCacheManager: PrefixCacheOwner {
         self.minPersistTokens = minPersistTokens
         self.prefillCostPerToken = prefillCostPerToken
         self.evictionHalfLifeSeconds = evictionHalfLifeSeconds
+        self.ttlSeconds = max(0, ttlSeconds)
         self.now = now
         self.accountant = accountant
         self.modelKey = modelKey
@@ -440,6 +463,18 @@ public actor PrefixCacheManager: PrefixCacheOwner {
         // digest) instead of trusting the stored path.
         let relPath = "\(modelDirComponent)/\(entry.digestHex).\(EncryptedKVStore.fileExtension)"
         let fileURL = cacheDir.appendingPathComponent(relPath)
+
+        // Sliding TTL: an entry is expired when it hasn't been hit within
+        // `ttlSeconds` (lastHitAt is bumped on every hit, so hot prefixes stay
+        // warm). Checked BEFORE decrypt — an expired entry is a miss, and its
+        // file + index entry are reclaimed (same drop path as a corrupt file).
+        // ttlSeconds == 0 disables (infinite retention). Bounds how long
+        // prompt-derived KV survives on disk (TB-007 oracle-window shrink).
+        if ttlSeconds > 0, now() - entry.lastHitAt > ttlSeconds {
+            stats.ttlExpirations += 1
+            await dropUnusableSSDFile(fileURL, digestHex: entry.digestHex, index: index)
+            return nil
+        }
 
         // MB-1: validate metadata BEFORE unwrap/decrypt. A wrong-model
         // file decrypts cleanly (AAD is its own metadata), so the cipher
@@ -604,6 +639,7 @@ public actor PrefixCacheManager: PrefixCacheOwner {
                 kvHeads: binding.kvHeads, headDim: binding.headDim, tokenCount: checkpointLength,
                 tokenPrefixHash: digestHex, kvCacheClass: "mixed",
                 metaState: [layoutJSON], chunkPlaintextSizes: chunks.map { $0.count }, createdAt: now(),
+                expiresAt: expiresAtForWrite(),
                 scope: scope
             )
             try await EncryptedKVStore.write(to: fileURL, metadata: meta, chunks: chunks, kek: kek)
@@ -701,6 +737,7 @@ public actor PrefixCacheManager: PrefixCacheOwner {
                     metaState: [layoutJSON],
                     chunkPlaintextSizes: chunks.map { $0.count },
                     createdAt: now(),
+                    expiresAt: expiresAtForWrite(),
                     scope: scopeFor(digestHex)
                 )
                 try await EncryptedKVStore.write(to: fileURL, metadata: meta, chunks: chunks, kek: kek)
@@ -804,6 +841,11 @@ public actor PrefixCacheManager: PrefixCacheOwner {
         let fm = FileManager.default
         let suffix = ".\(EncryptedKVStore.fileExtension)"
 
+        // Proactively reclaim TTL-expired entries before re-indexing, so a
+        // restart doesn't resurrect stale KV (and disk is freed even if the
+        // model is never looked up again this session).
+        reapExpired(index: index, cacheDir: cacheDir)
+
         // Drop index entries whose backing file vanished.
         for entry in index.entries(modelHash: binding.modelHash) {
             let url = modelDir.appendingPathComponent("\(entry.digestHex)\(suffix)")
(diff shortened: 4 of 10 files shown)


ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- **[INFO]** `:1` — 
- **[MEDIUM]** `:1` — 
- **[INFO]** `:1` — 

### Additive_complexity — 4 finding(s) (2 blocking)
- **[MEDIUM]** `:1` — 
- **[INFO]** `:1` — 
- **[MEDIUM]** `:1` — 
- **[INFO]** `:1` — 

_7 finding(s) total, 3 blocking. Verdict: REQUEST_CHANGES._

Gajesh2007:
@codex review this

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -938,6 +991,23 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+    private func reapExpired(index: PrefixCacheIndex, cacheDir: URL) {
+        guard !closed, ttlSeconds > 0 else { return }
+        let cutoff = now() - ttlSeconds
+        for entry in index.entries(modelHash: binding.modelHash) where entry.lastHitAt < cutoff {
+            let url = cacheDir.appendingPathComponent(
+                "\(modelDirComponent)/\(entry.digestHex).\(EncryptedKVStore.fileExtension)")
+            try? FileManager.default.removeItem(at: url)
+            index.remove(modelHash: binding.modelHash, digestHex: entry.digestHex)
+            stats.ttlExpirations += 1
+        }
```
ethenotethan, line 1108: 🟡 **[MEDIUM]** ⚡ reapExpired iterates all entries without bounds, potentially causing O(n) allocation spike

💡 _Suggestion:_ Add batching or limit the number of entries processed per call, especially for large indexes with many expired entries

📊 _Score: 3×4 = 12 · Category: Unbounded allocations_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -834,12 +876,22 @@ public actor PrefixCacheManager: PrefixCacheOwner {
                 try? fm.removeItem(at: url)  // foreign / corrupt / mislabeled
                 continue
             }
+            // TTL: don't resurrect an orphan whose own metadata says it's
+            // already expired (expiresAt = createdAt + ttl at write). Delete it.
+            if ttlSeconds > 0, let exp = meta.expiresAt, now() > exp {
+                try? fm.removeItem(at: url)
+                stats.ttlExpirations += 1
+                continue
+            }
```
ethenotethan, line 986: 🔵 **[INFO]** ⚡ reconcileWithDisk performs synchronous file operations in a loop without yielding

💡 _Suggestion:_ Add periodic Task.yield() calls in the file enumeration loop to prevent blocking the actor for extended periods

📊 _Score: 2×3 = 6 · Category: Blocking I/O on hot paths_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -441,6 +464,18 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+        // `ttlSeconds` (lastHitAt is bumped on every hit, so hot prefixes stay
+        // warm). Checked BEFORE decrypt — an expired entry is a miss, and its
+        // file + index entry are reclaimed (same drop path as a corrupt file).
+        // ttlSeconds == 0 disables (infinite retention). Bounds how long
+        // prompt-derived KV survives on disk (TB-007 oracle-window shrink).
+        if ttlSeconds > 0, now() - entry.lastHitAt > ttlSeconds {
+            stats.ttlExpirations += 1
+            await dropUnusableSSDFile(fileURL, digestHex: entry.digestHex, index: index)
+            return nil
+        }
```
ethenotethan: 🟡 **[MEDIUM]** ⚡ TTL expiry check calls now() on every SSD lookup even when TTL is disabled

💡 _Suggestion:_ Move the ttlSeconds > 0 check before calling now() to avoid unnecessary timestamp computation when TTL is disabled

provider-swift/Sources/ProviderCore/Inference/MultiModelBatchSchedulerEngine.swift
```diff
@@ -122,7 +122,8 @@ public struct MultiModelBatchSchedulerEngine: MLXServerEngine, Sendable {
         acquire: @escaping @Sendable (String) async throws -> AcquiredModel,
         tokenizerProvider: @escaping @Sendable (String?) async throws -> TokenizerHandle,
         availableModels: @escaping @Sendable () async -> [String],
-        defaultMaxTokens: Int = 4096
+        defaultMaxTokens: Int = 4096,
```
ethenotethan, line 125: 🔵 **[INFO]** 🏷️ cacheScope parameter uses bare string type instead of typed wrapper

💡 _Suggestion:_ Consider wrapping in a CacheScope newtype or enum to prevent string confusion and enable compile-time validation

📊 _Score: 2×1 = 2 · Category: stringly-typed_

provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
```diff
@@ -942,6 +945,23 @@ public actor BatchScheduler {
...
+    static func prefixCacheTTLSeconds() -> Int64 {
+        resolveTTLSeconds(env: ProcessInfo.processInfo.environment["DARKBLOOM_PREFIX_CACHE_TTL_SECONDS"])
+    }
+
+    /// Pure TTL policy (testable). Unset / malformed / negative ⇒ default;
+    /// `0` ⇒ disabled (infinite); a positive value sets the sliding window.
+    static func resolveTTLSeconds(env: String?) -> Int64 {
+        guard let v = env else { return defaultPrefixCacheTTLSeconds }
+        guard let n = Int64(v), n >= 0 else { return defaultPrefixCacheTTLSeconds }
+        return n  // 0 ⇒ disabled
```
ethenotethan, line 971: 🔵 **[INFO]** 🧩 Unnecessary wrapper function for simple environment variable parsing

💡 _Suggestion:_ Inline the environment variable parsing directly in prefixCacheTTLSeconds() or use a generic env parsing helper if this pattern is repeated elsewhere

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -270,6 +280,17 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+    /// Absolute expiry stamped into a freshly-written file's metadata:
+    /// `createdAt + ttl`, or nil when TTL is disabled. This is the EARLIEST
+    /// possible expiry (a never-hit entry, where lastHitAt == createdAt); the
+    /// live check slides off the mutable `lastHitAt` in the index, so a hit
+    /// extends effective lifetime without re-sealing the file. Informational +
+    /// a hard floor for any future offline reaper; the authoritative runtime
+    /// check is the lastHitAt comparison in loadFromSSD/sweep.
+    private func expiresAtForWrite() -> Int64? {
+        ttlSeconds > 0 ? now() + ttlSeconds : nil
+    }
```
ethenotethan, line 309: 🔵 **[INFO]** 🧩 Single-use helper function that could be inlined

💡 _Suggestion:_ Inline the TTL calculation directly where used, or make it a computed property if accessed frequently

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -938,6 +991,23 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+        guard !closed, ttlSeconds > 0 else { return }
+        let cutoff = now() - ttlSeconds
+        for entry in index.entries(modelHash: binding.modelHash) where entry.lastHitAt < cutoff {
+            let url = cacheDir.appendingPathComponent(
+                "\(modelDirComponent)/\(entry.digestHex).\(EncryptedKVStore.fileExtension)")
+            try? FileManager.default.removeItem(at: url)
+            index.remove(modelHash: binding.modelHash, digestHex: entry.digestHex)
+            stats.ttlExpirations += 1
+        }
+    }
```
ethenotethan, line 1109: 🔵 **[INFO]** 🧩 TTL expiry check duplicated between reapExpired and loadFromSSD

💡 _Suggestion:_ Extract the expiry check logic into a shared helper method to avoid duplication

📊 _Score: 1×2 = 2 · Category: duplicate logic_

provider-swift/Sources/ProviderCore/Inference/MultiModelBatchSchedulerEngine.swift
```diff
@@ -133,7 +134,11 @@ public struct MultiModelBatchSchedulerEngine: MLXServerEngine, Sendable {
         self.releaseModel = { _ in }
         self.defaultMaxTokens = defaultMaxTokens
         self.reasoningEffort = nil
-        self.cacheScope = ""
+        // Fixed per-server scope for the standalone (--local) path, which can't
+        // carry a per-request prompt_cache_key through the upstream router. ""
+        // ⇒ unscoped (default). Set via DARKBLOOM_PREFIX_CACHE_SCOPE; used to
+        // exercise/validate cross-tenant isolation on a single box.
+        self.cacheScope = cacheScope
```
ethenotethan, line 141: 🔵 **[INFO]** 🧩 Adding another configuration parameter when the existing pattern could be extended

💡 _Suggestion:_ Consider if cacheScope could be part of the existing configuration structure rather than a separate parameter

📊 _Score: 1×2 = 2 · Category: over-configuration_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:991-1008` — reapExpired iterates all entries without bounds, potentially causing O(n) allocation spike
  - _Suggestion:_ Add batching or limit the number of entries processed per call, especially for large indexes with many expired entries
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:841-885` — reconcileWithDisk performs synchronous file operations in a loop without yielding
  - _Suggestion:_ Add periodic Task.yield() calls in the file enumeration loop to prevent blocking the actor for extended periods
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:464-477` — TTL expiry check calls now() on every SSD lookup even when TTL is disabled
  - _Suggestion:_ Move the ttlSeconds > 0 check before calling now() to avoid unnecessary timestamp computation when TTL is disabled

### Type_diligence — 1 finding(s)
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/MultiModelBatchSchedulerEngine.swift:125` — cacheScope parameter uses bare string type instead of typed wrapper
  - _Suggestion:_ Consider wrapping in a CacheScope newtype or enum to prevent string confusion and enable compile-time validation

### Additive_complexity — 4 finding(s)
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:948-962` — Unnecessary wrapper function for simple environment variable parsing
  - _Suggestion:_ Inline the environment variable parsing directly in prefixCacheTTLSeconds() or use a generic env parsing helper if this pattern is repeated elsewhere
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:289-292` — Single-use helper function that could be inlined
  - _Suggestion:_ Inline the TTL calculation directly where used, or make it a computed property if accessed frequently
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:994-1009` — TTL expiry check duplicated between reapExpired and loadFromSSD
  - _Suggestion:_ Extract the expiry check logic into a shared helper method to avoid duplication
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/MultiModelBatchSchedulerEngine.swift:137-141` — Adding another configuration parameter when the existing pattern could be extended
  - _Suggestion:_ Consider if cacheScope could be part of the existing configuration structure rather than a separate parameter

_8 finding(s) total, 2 blocking. Verdict: REQUEST_CHANGES._

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -367,16 +390,27 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+                    } else if hit.tokenCount >= minPersistTokens {
+                        // TB-016 sub-feature B: 2nd-use promotion. RAM hit above
+                        // the persist threshold and NOT already on SSD — schedule
+                        // a detached promotion (no blocking the lookup actor).
+                        let cpScope = scope
+                        Task.detached { [weak self] in
+                            await self?.persistDigest(cp.digest, scope: cpScope)
+                        }
                     }
                 }
```
ethenotethan, line 480: 🟡 **[MEDIUM]** ⚡ Synchronous file I/O in hot path during RAM hit processing

💡 _Suggestion:_ Move the index.touch() call to a detached Task to avoid blocking the lookup path on disk I/O

📊 _Score: 3×4 = 12 · Category: blocking_io_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -938,6 +1002,23 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+    private func reapExpired(index: PrefixCacheIndex, cacheDir: URL) {
+        guard !closed, ttlSeconds > 0 else { return }
+        let cutoff = now() - ttlSeconds
+        for entry in index.entries(modelHash: binding.modelHash) where entry.lastHitAt < cutoff {
+            let url = cacheDir.appendingPathComponent(
+                "\(modelDirComponent)/\(entry.digestHex).\(EncryptedKVStore.fileExtension)")
+            try? FileManager.default.removeItem(at: url)
+            index.remove(modelHash: binding.modelHash, digestHex: entry.digestHex)
+            stats.ttlExpirations += 1
+        }
```
ethenotethan, line 1108: 🟡 **[MEDIUM]** ⚡ Unbounded loop over all entries during TTL reaping without batching

💡 _Suggestion:_ Add batching/yielding in the reapExpired loop to prevent blocking on large caches: process entries in chunks and yield control periodically

📊 _Score: 4×3 = 12 · Category: unbounded_loop_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -834,12 +887,22 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+            // already expired (expiresAt = createdAt + ttl at write). Delete it.
+            if ttlSeconds > 0, let exp = meta.expiresAt, now() > exp {
+                try? fm.removeItem(at: url)
+                stats.ttlExpirations += 1
+                continue
+            }
             let bytes = (try? fm.attributesOfItem(atPath: url.path)[.size] as? Int) ?? nil
+            // Seed lastHitAt from the file's own createdAt (NOT now()), so a
+            // re-indexed orphan keeps its real age — re-indexing can't reset the
+            // sliding TTL clock and indefinitely extend a stale entry.
```
ethenotethan, line 990: 🟡 **[MEDIUM]** ⚡ Synchronous file operations in reconcileWithDisk without yielding

💡 _Suggestion:_ Add periodic Task.yield() calls in the reconciliation loops to prevent blocking the actor for extended periods during large directory scans

📊 _Score: 3×3 = 9 · Category: blocking_io_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -604,6 +650,7 @@ public actor PrefixCacheManager: PrefixCacheOwner {
                 kvHeads: binding.kvHeads, headDim: binding.headDim, tokenCount: checkpointLength,
                 tokenPrefixHash: digestHex, kvCacheClass: "mixed",
```
ethenotethan, line 734: 🔵 **[INFO]** 🏷️ expiresAt metadata field accessed without type assertion

💡 _Suggestion:_ Add type assertion when accessing expiresAt from metadata: `guard let expiresAt = meta.expiresAt as? Int64 else { ... }`

📊 _Score: 3×2 = 6 · Category: missing_type_assertion_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -834,12 +887,22 @@ public actor PrefixCacheManager: PrefixCacheOwner {
                 try? fm.removeItem(at: url)  // foreign / corrupt / mislabeled
                 continue
             }
+            // TTL: don't resurrect an orphan whose own metadata says it's
+            // already expired (expiresAt = createdAt + ttl at write). Delete it.
+            if ttlSeconds > 0, let exp = meta.expiresAt, now() > exp {
```
ethenotethan: 🔵 **[INFO]** 🏷️ expiresAt metadata field accessed without type assertion

💡 _Suggestion:_ Add type assertion when accessing expiresAt from metadata: `guard let exp = meta.expiresAt as? Int64 else { continue }`

📊 _Score: 3×2 = 6 · Category: missing_type_assertion_

provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
ethenotethan, line 971: 🔵 **[INFO]** 🧩 Unnecessary wrapper function for simple environment variable parsing

💡 _Suggestion:_ Inline the TTL parsing logic directly in prefixCacheTTLSeconds() or use a generic env parsing helper if this pattern repeats elsewhere

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
ethenotethan, line 309: 🔵 **[INFO]** 🧩 Single-use helper function for simple conditional expression

💡 _Suggestion:_ Inline the logic: `ttlSeconds > 0 ? now() + ttlSeconds : nil` directly at call sites

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
ethenotethan, line 1108: 🟡 **[MEDIUM]** 🧩 TTL expiry check duplicated between reapExpired and loadFromSSD

💡 _Suggestion:_ Extract shared TTL check into a helper method: `isExpired(lastHitAt: Int64) -> Bool` to avoid maintaining the same logic in two places

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
ethenotethan, line 480: 🟡 **[MEDIUM]** 🧩 Complex nested conditional logic for RAM hit handling with SSD touch

💡 _Suggestion:_ Extract the SSD touch logic into a separate method to reduce nesting and improve readability of the main lookup path

📊 _Score: 3×3 = 9 · Category: additive complexity_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Performance — 3 finding(s) (3 blocking)
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:390-415` — Synchronous file I/O in hot path during RAM hit processing
  - _Suggestion:_ Move the index.touch() call to a detached Task to avoid blocking the lookup path on disk I/O
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:1005-1019` — Unbounded loop over all entries during TTL reaping without batching
  - _Suggestion:_ Add batching/yielding in the reapExpired loop to prevent blocking on large caches: process entries in chunks and yield control periodically
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:852-900` — Synchronous file operations in reconcileWithDisk without yielding
  - _Suggestion:_ Add periodic Task.yield() calls in the reconciliation loops to prevent blocking the actor for extended periods during large directory scans

### Type_diligence — 2 finding(s)
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:651` — expiresAt metadata field accessed without type assertion
  - _Suggestion:_ Add type assertion when accessing expiresAt from metadata: `guard let expiresAt = meta.expiresAt as? Int64 else { ... }`
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:892` — expiresAt metadata field accessed without type assertion
  - _Suggestion:_ Add type assertion when accessing expiresAt from metadata: `guard let exp = meta.expiresAt as? Int64 else { continue }`

### Additive_complexity — 4 finding(s) (2 blocking)
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:953-962` — Unnecessary wrapper function for simple environment variable parsing
  - _Suggestion:_ Inline the TTL parsing logic directly in prefixCacheTTLSeconds() or use a generic env parsing helper if this pattern repeats elsewhere
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:289-292` — Single-use helper function for simple conditional expression
  - _Suggestion:_ Inline the logic: `ttlSeconds > 0 ? now() + ttlSeconds : nil` directly at call sites
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:1005-1019` — TTL expiry check duplicated between reapExpired and loadFromSSD
  - _Suggestion:_ Extract shared TTL check into a helper method: `isExpired(lastHitAt: Int64) -> Bool` to avoid maintaining the same logic in two places
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:393-415` — Complex nested conditional logic for RAM hit handling with SSD touch
  - _Suggestion:_ Extract the SSD touch logic into a separate method to reduce nesting and improve readability of the main lookup path

_9 finding(s) total, 5 blocking. Verdict: REQUEST_CHANGES._

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -441,6 +464,18 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
 
+        // Sliding TTL: an entry is expired when it hasn't been hit within
+        // `ttlSeconds` (lastHitAt is bumped on every hit, so hot prefixes stay
+        // warm). Checked BEFORE decrypt — an expired entry is a miss, and its
+        // file + index entry are reclaimed (same drop path as a corrupt file).
+        // ttlSeconds == 0 disables (infinite retention). Bounds how long
+        // prompt-derived KV survives on disk (TB-007 oracle-window shrink).
+        if ttlSeconds > 0, now() - entry.lastHitAt > ttlSeconds {
+            stats.ttlExpirations += 1
+            await dropUnusableSSDFile(fileURL, digestHex: entry.digestHex, index: index)
```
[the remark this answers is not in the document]
anupsv: Fixed in `c0709d67`. You're right — the sliding TTL is *time since last use*, and use includes RAM serves. The RAM-hit branch now touches the SSD index entry (`lastHitAt = now`) whenever the hit digest also exists on SSD, so a RAM-hot prefix can no longer be reaped as expired when RAM pressure evicts it and the next lookup falls through to SSD.

Details:
- Gated on `ttlSeconds > 0`, so `ttl=0` behavior is byte-identical (and the disabled path does no extra `now()`/touch work).
- The touch is an in-memory dict update on the already-held manager actor; index persistence stays coalesced as before (no new write pressure on the hot path).
- The same branch restructure keeps the TB-016 2nd-use promotion path unchanged for the no-SSD-entry case.

Regression test `ttlRamHitSlidesSSDRecency`: 5 RAM serves spanning 1000s (each gap < the 300s TTL), then RAM eviction → the SSD fallthrough must hit, not reap. Verified it **fails without the fix** (entry reaped, `ttlExpirations=1`) and passes with it. Full cache suite (109 tests) green.

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -834,12 +876,22 @@ public actor PrefixCacheManager: PrefixCacheOwner {
                 try? fm.removeItem(at: url)  // foreign / corrupt / mislabeled
                 continue
             }
+            // TTL: don't resurrect an orphan whose own metadata says it's
+            // already expired (expiresAt = createdAt + ttl at write). Delete it.
+            if ttlSeconds > 0, let exp = meta.expiresAt, now() > exp {
```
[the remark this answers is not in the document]
anupsv: Fixed in `f3fe3f95`. The reconcile orphan branch now treats nil `expiresAt` as `createdAt + ttl` — exactly what `expiresAtForWrite` stamps on new files — so legacy files (pre-TTL, or written while TTL was disabled) get identical semantics and can't be resurrected past the privacy window on upgrade/restart.

Notes:
- They were never *served* stale (the lazy read-path check expires on next lookup) — the gap was disk lingering, which matters because the TTL's stated guarantee is bounding how long prompt-derived KV survives on SSD.
- The addition is **saturating** (new `saturatingExpiry` helper, also adopted by `expiresAtForWrite`) — this also covers the write-site half of your overflow comment; the read-side guard + parse clamp land separately.
- Regression test asserts on the **physical file** after reconcile (a lookup-based assertion would be masked by the lazy check) — verified it fails without the fix. A companion test guards against over-deletion: a younger-than-TTL legacy orphan stays re-indexed and servable.

Full cache suite (111 tests) green.

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -367,16 +397,27 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+                        // a RAM-hot prefix older than the TTL gets reaped as
+                        // "expired" the moment RAM pressure evicts it and the
+                        // next lookup falls through to SSD. Gated on TTL being
+                        // enabled so ttl=0 behavior stays byte-identical.
+                        if ttlSeconds > 0 {
+                            index.touch(modelHash: binding.modelHash, digestHex: digestHex, now: now())
+                        }
+                    } else if hit.tokenCount >= minPersistTokens {
+                        // TB-016 sub-feature B: 2nd-use promotion. RAM hit above
+                        // the persist threshold and NOT already on SSD — schedule
```
ethenotethan, line 473: 🟡 **[MEDIUM]** ⚡ Task.detached blocks on persistDigest without timeout or cancellation

💡 _Suggestion:_ Add timeout and cancellation handling to the detached task, or use structured concurrency with TaskGroup to prevent unbounded resource usage

📊 _Score: 3×4 = 12 · Category: blocking_io_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -938,6 +1016,23 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+    private func reapExpired(index: PrefixCacheIndex, cacheDir: URL) {
+        guard !closed, ttlSeconds > 0 else { return }
+        let cutoff = now() - ttlSeconds
+        for entry in index.entries(modelHash: binding.modelHash) where entry.lastHitAt < cutoff {
+            let url = cacheDir.appendingPathComponent(
+                "\(modelDirComponent)/\(entry.digestHex).\(EncryptedKVStore.fileExtension)")
+            try? FileManager.default.removeItem(at: url)
+            index.remove(modelHash: binding.modelHash, digestHex: entry.digestHex)
+            stats.ttlExpirations += 1
+        }
```
ethenotethan: 🔵 **[INFO]** ⚡ reapExpired iterates all entries for single model without early termination

💡 _Suggestion:_ Consider adding an index by lastHitAt or early termination if entries are sorted by recency to avoid O(n) scan on every reconcile

📊 _Score: 2×3 = 6 · Category: inefficient_loop_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -834,12 +894,29 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+            // Seed lastHitAt from the file's own createdAt (NOT now()), so a
+            // re-indexed orphan keeps its real age — re-indexing can't reset the
+            // sliding TTL clock and indefinitely extend a stale entry.
             index.record(PrefixIndexEntry(
                 modelHash: binding.modelHash, digestHex: digestHex,
                 tokenCount: meta.tokenCount,
                 relativePath: "\(modelDirComponent)/\(name)",
-                fileBytes: bytes ?? 0, createdAt: meta.createdAt, lastHitAt: now()))
+                fileBytes: bytes ?? 0, createdAt: meta.createdAt, lastHitAt: meta.createdAt))
         }
```
ethenotethan, line 996: 🔵 **[INFO]** ⚡ saturatingExpiry called twice for same createdAt in reconcile loop

💡 _Suggestion:_ Cache the expiry calculation result when processing multiple entries with the same createdAt timestamp

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -282,6 +310,7 @@ public actor PrefixCacheManager: PrefixCacheOwner {
         minPersistTokens: Int = 0,
```
ethenotethan, line 367: 🔵 **[INFO]** 🏷️ expiresAtForWrite() returns Int64? but could use a more specific type

💡 _Suggestion:_ Consider using a dedicated ExpiryTime type or TimeInterval to make the semantics clearer and prevent confusion with other Int64 timestamps

📊 _Score: 2×1 = 2 · Category: bare_interface_

provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
```diff
@@ -942,6 +945,23 @@ public actor BatchScheduler {
...
+    }
```
ethenotethan, line 972: 🔵 **[INFO]** 🧩 resolveTTLSeconds function could be simplified with guard-let pattern

💡 _Suggestion:_ Replace the nested guard statements with a single guard-let: `guard let v = env, let n = Int64(v), n >= 0 else { return defaultPrefixCacheTTLSeconds }`

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -270,6 +280,24 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+    private func expiresAtForWrite() -> Int64? {
+        ttlSeconds > 0 ? saturatingExpiry(from: now()) : nil
+    }
+
+    /// Saturating `base + ttlSeconds`. An operator-set "effectively infinite"
+    /// TTL (e.g. Int64.max) must clamp to never-expires, not trap on overflow.
+    private func saturatingExpiry(from base: Int64) -> Int64 {
+        let (sum, overflow) = base.addingReportingOverflow(ttlSeconds)
+        return overflow ? Int64.max : sum
+    }
```
ethenotethan, line 309: 🔵 **[INFO]** 🧩 expiresAtForWrite and saturatingExpiry could be combined into single function

💡 _Suggestion:_ Merge these into one function: `private func expiresAtForWrite() -> Int64? { guard ttlSeconds > 0 else { return nil }; let (sum, overflow) = now().addingReportingOverflow(ttlSeconds); return overflow ? Int64.max : sum }`

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -367,16 +397,27 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+                    } else if hit.tokenCount >= minPersistTokens {
+                        // TB-016 sub-feature B: 2nd-use promotion. RAM hit above
+                        // the persist threshold and NOT already on SSD — schedule
+                        // a detached promotion (no blocking the lookup actor).
+                        let cpScope = scope
+                        Task.detached { [weak self] in
+                            await self?.persistDigest(cp.digest, scope: cpScope)
+                        }
                     }
                 }
```
ethenotethan, line 480: 🟡 **[MEDIUM]** 🧩 nested conditional logic in lookup makes control flow harder to follow

💡 _Suggestion:_ Extract the SSD touch/promotion logic into a separate method like `handleSSDEntryForRAMHit(digestHex:hit:)` to flatten the nested conditions

📊 _Score: 3×4 = 12 · Category: additive complexity_

provider-swift/Tests/ProviderCoreTests/KVCache/PrefixCacheManagerTests.swift
```diff
@@ -572,3 +572,208 @@ func managerScopedSSDRoundtripAndIsolation() async throws {
...
+        kek: kek ?? KVCacheKEK(wrapper: InMemoryKeyWrappingService(),
+                               storage: InMemoryWrappedKEKStorage(identifier: UUID().uuidString)),
+        cacheDir: cacheDir, ssdEnabled: true,
+        boundaries: [4, 8],
+        ttlSeconds: ttl,
+        now: { clock.value },
+        modelKey: "test-model"
+    )
+    return (mgr, cacheDir)
+}
```
ethenotethan, line 597: 🔵 **[INFO]** 🧩 makeTTLManager duplicates manager creation pattern from other tests

💡 _Suggestion:_ Extract common manager creation logic into a shared test helper that accepts configuration parameters

📊 _Score: 2×2 = 4 · Category: duplicate logic_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:409-415` — Task.detached blocks on persistDigest without timeout or cancellation
  - _Suggestion:_ Add timeout and cancellation handling to the detached task, or use structured concurrency with TaskGroup to prevent unbounded resource usage
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:1026-1033` — reapExpired iterates all entries for single model without early termination
  - _Suggestion:_ Consider adding an index by lastHitAt or early termination if entries are sorted by recency to avoid O(n) scan on every reconcile
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:904-920` — saturatingExpiry called twice for same createdAt in reconcile loop
  - _Suggestion:_ Cache the expiry calculation result when processing multiple entries with the same createdAt timestamp

### Type_diligence — 1 finding(s)
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:310` — expiresAtForWrite() returns Int64? but could use a more specific type
  - _Suggestion:_ Consider using a dedicated ExpiryTime type or TimeInterval to make the semantics clearer and prevent confusion with other Int64 timestamps

- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:948-963` — resolveTTLSeconds function could be simplified with guard-let pattern
  - _Suggestion:_ Replace the nested guard statements with a single guard-let: `guard let v = env, let n = Int64(v), n >= 0 else { return defaultPrefixCacheTTLSeconds }`
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:283-299` — expiresAtForWrite and saturatingExpiry could be combined into single function
  - _Suggestion:_ Merge these into one function: `private func expiresAtForWrite() -> Int64? { guard ttlSeconds > 0 else { return nil }; let (sum, overflow) = now().addingReportingOverflow(ttlSeconds); return overflow ? Int64.max : sum }`
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:400-422` — nested conditional logic in lookup makes control flow harder to follow
  - _Suggestion:_ Extract the SSD touch/promotion logic into a separate method like `handleSSDEntryForRAMHit(digestHex:hit:)` to flatten the nested conditions
- 🔵 **[INFO]** `provider-swift/Tests/ProviderCoreTests/KVCache/PrefixCacheManagerTests.swift:577-597` — makeTTLManager duplicates manager creation pattern from other tests
  - _Suggestion:_ Extract common manager creation logic into a shared test helper that accepts configuration parameters

_8 finding(s) total, 2 blocking. Verdict: REQUEST_CHANGES._

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -441,6 +464,18 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+        // Sliding TTL: an entry is expired when it hasn't been hit within
+        // `ttlSeconds` (lastHitAt is bumped on every hit, so hot prefixes stay
+        // warm). Checked BEFORE decrypt — an expired entry is a miss, and its
+        // file + index entry are reclaimed (same drop path as a corrupt file).
+        // ttlSeconds == 0 disables (infinite retention). Bounds how long
+        // prompt-derived KV survives on disk (TB-007 oracle-window shrink).
+        if ttlSeconds > 0, now() - entry.lastHitAt > ttlSeconds {
+            stats.ttlExpirations += 1
+            await dropUnusableSSDFile(fileURL, digestHex: entry.digestHex, index: index)
+            return nil
```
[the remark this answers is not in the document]
anupsv: Fixed in `1d35b6f1`. The expiry check is now a selection loop: each expired candidate is reaped (the drop removes it from the index, so `findLongestCheckpoint` naturally yields the next-longest — progress guaranteed) and the first still-fresh checkpoint serves. So an expired 8k checkpoint no longer masks a hot shared system-prefix at 4k.

Two guards worth noting: a `closed` re-check breaks the cycle where `dropUnusableSSDFile` intentionally no-ops on a closed manager (a naive loop would refetch the same entry forever), plus a defensive iteration cap of `boundaries.count + 1`.

Regression test `ttlExpiredLongestFallsBackToShorterCheckpoint`: cp8 goes stale while cp4 stays warm (kept fresh by the RAM-hit touch from the earlier fix), RAM evicted → the lookup reaps cp8 and serves cp4 from SSD (`tokenCount == 4`, `ttlExpirations == 1`). Verified it fails without the fix. Full cache suite (112 tests) green.

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -441,6 +464,18 @@ public actor PrefixCacheManager: PrefixCacheOwner {
         let relPath = "\(modelDirComponent)/\(entry.digestHex).\(EncryptedKVStore.fileExtension)"
         let fileURL = cacheDir.appendingPathComponent(relPath)
 
+        // Sliding TTL: an entry is expired when it hasn't been hit within
+        // `ttlSeconds` (lastHitAt is bumped on every hit, so hot prefixes stay
+        // warm). Checked BEFORE decrypt — an expired entry is a miss, and its
+        // file + index entry are reclaimed (same drop path as a corrupt file).
+        // ttlSeconds == 0 disables (infinite retention). Bounds how long
+        // prompt-derived KV survives on disk (TB-007 oracle-window shrink).
+        if ttlSeconds > 0, now() - entry.lastHitAt > ttlSeconds {
```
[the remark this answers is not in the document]
anupsv: Fixed in `1d35b6f1` (same commit as the shorter-checkpoint fallback, since it rewrites this exact comparison). The age computation now uses `subtractingReportingOverflow`; overflow (e.g. a crash-corrupted `lastHitAt = Int64.min` in the plaintext index) is treated as **expired → drop the corrupt entry**, instead of trapping the provider mid-lookup. The write-side counterpart (`now() + ttl`) was made saturating in `f3fe3f95` (`saturatingExpiry`), and `reapExpired`'s cutoff gets the same treatment in the TTL-clamp follow-up.

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -270,6 +280,17 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
 
+    /// Absolute expiry stamped into a freshly-written file's metadata:
+    /// `createdAt + ttl`, or nil when TTL is disabled. This is the EARLIEST
+    /// possible expiry (a never-hit entry, where lastHitAt == createdAt); the
+    /// live check slides off the mutable `lastHitAt` in the index, so a hit
+    /// extends effective lifetime without re-sealing the file. Informational +
+    /// a hard floor for any future offline reaper; the authoritative runtime
+    /// check is the lastHitAt comparison in loadFromSSD/sweep.
+    private func expiresAtForWrite() -> Int64? {
+        ttlSeconds > 0 ? now() + ttlSeconds : nil
```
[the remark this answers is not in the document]
anupsv: Done across three commits — all TTL arithmetic now saturates, so a huge `DARKBLOOM_PREFIX_CACHE_TTL_SECONDS` (e.g. `Int64.max`) genuinely means infinite retention with no traps: the write stamp `now()+ttl` saturates to `Int64.max` (`f3fe3f95` — this was the reachable trap you flagged), the read-path age check uses overflow-reporting subtraction (`1d35b6f1`), and `reapExpired`'s cutoff saturates to reap-nothing (latest commit; honestly, that site can't underflow with a positive wall clock — it's defense-in-depth + uniformity). We chose saturation over a parse clamp so "effectively infinite" stays a first-class, well-defined operator choice. End-to-end test `ttlEffectivelyInfiniteTTLNeverTrapsOrReaps`: `ttl=Int64.max` through store/flush, reconcile, and a ~32-year-later lookup — no trap, no reap, SSD hit.

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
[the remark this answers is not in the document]
anupsv: Fixed in `b23897c7`. Touches (both the SSD-hit touch and the RAM-hit touch added earlier) now trigger a **time-coalesced index save** — `persistRecencyIfDue`, at most one save per 60s, TTL-gated — so a crash loses at most 60s of recency against the 300s default TTL instead of everything since the last graceful flush. No fsync-per-warm-request: the coalescing bounds hot-path I/O.

Regression test `ttlTouchRecencyPersistsAcrossCrash`: persisted lastHitAt=1000, warm hit at t=1200 (touch + coalesced save), hard crash (no graceful teardown), restart at t=1400 where the reap cutoff is 1100 → the entry must survive reconcile off the persisted touch. Fails without the fix (persisted recency would still be 1000 → reaped).

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -938,6 +991,23 @@ public actor PrefixCacheManager: PrefixCacheOwner {
     /// index.json without bound and can fill the volume. 0 budget = no cap.
     ///
     /// TB-016 sub-feature C: Uses benefit-per-byte scoring instead of LRU.
+    /// Proactively drop SSD entries past the sliding TTL (file + index entry),
+    /// independent of the disk budget. Runs at load-time reconcile so expired
+    /// KV is reclaimed even for a model that's never looked up again; the
+    /// loadFromSSD check covers steady-state. No-op when TTL is disabled or the
+    /// manager is closed (a closed manager must not mutate a handed-off dir).
```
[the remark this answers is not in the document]
anupsv: Fixed in `b23897c7`. New `PrefixCacheManager.reapExpiredTick()` (reap + coalesced index save + accountant notify, mirroring `dropUnusableSSDFile`'s discipline) driven by a periodic `BatchScheduler` task at cadence `max(60, min(ttl/2, 300))`s — so an expired entry outlives its TTL by at most ~one cadence while the model stays loaded. Lifecycle mirrors the stats logger (started with the engine, cancelled in `stopCurrentEngine`) but is deliberately **independent of the stats logger** so disabling observability never disables the privacy reaper. The sweep is an in-memory index scan; file deletes happen only when something actually expired.

Test `ttlReapExpiredTickReapsWhileModelStaysLoaded`: cold entry reclaimed by the tick with no reconcile/restart, asserted on the physical file.

provider-swift/Sources/ProviderCore/Server/LocalInferenceHTTP.swift
```diff
@@ -60,13 +60,15 @@ func makeLocalInferenceApplication(
     acquire: @escaping @Sendable (String) async throws -> MultiModelBatchSchedulerEngine.AcquiredModel,
     tokenizerProvider: @escaping @Sendable (String?) async throws -> TokenizerHandle,
     availableModels: @escaping @Sendable () async -> [String],
+    cacheScope: String = "",
```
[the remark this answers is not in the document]
anupsv: Fixed in `b23897c7`. `ProviderLoop.startLocalEndpoint` now passes `DARKBLOOM_PREFIX_CACHE_SCOPE` into `makeLocalInferenceApplication` exactly like the standalone `--local` path, so both local HTTP surfaces honor the per-server scope consistently. The coordinator WS path is unchanged — it carries the real per-request `prompt_cache_key` scope and never consults this env var.

provider-swift/Sources/ProviderCore/Inference/BatchScheduler+EngineBridge.swift
```diff
@@ -282,6 +282,27 @@ extension BatchScheduler {
...
+        let ttl = Self.prefixCacheTTLSeconds()
+        guard ttl > 0, let mgr = checkpointManager else { return }
+        let cadence = Swift.max(60, Swift.min(ttl / 2, 300))
+        ttlReapTask = Task.detached { [weak mgr] in
+            while !Task.isCancelled {
+                try? await Task.sleep(for: .seconds(Double(cadence)))
+                if Task.isCancelled { return }
+                await mgr?.reapExpiredTick()
+            }
+        }
```
ethenotethan, line 303: 🟡 **[MEDIUM]** ⚡ TTL reaper task lacks proper cancellation handling in error paths

💡 _Suggestion:_ Add defer { ttlReapTask?.cancel() } or wrap the sleep/reap loop in a cancellation-aware pattern to ensure cleanup on task failure

📊 _Score: 2×4 = 8 · Category: unbounded goroutines_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -427,9 +513,45 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+        // drop path that leaves the entry behind.
+        var attempts = boundaries.count + 1
+        while attempts > 0, !closed, let candidate = index.findLongestCheckpoint(
             modelHash: binding.modelHash, tokens: tokens, boundaries: boundaries, scope: scope
-        ) else { return nil }
+        ) {
+            attempts -= 1
+            if ttlSeconds > 0 {
+                // Overflow-safe age: index.json is plaintext and can be left
+                // corrupt by a crash — an extreme lastHitAt (e.g. Int64.min)
```
ethenotethan: 🟡 **[MEDIUM]** ⚡ Potential O(n) loop in TTL expiry check during SSD lookup

💡 _Suggestion:_ Consider adding an early break or limit on attempts to prevent excessive iteration when many entries are expired

📊 _Score: 3×3 = 9 · Category: repeated work_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -938,6 +1086,26 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+        // Int64.max) must mean "reap nothing", not trap on underflow.
+        let (diff, underflow) = now().subtractingReportingOverflow(ttlSeconds)
+        let cutoff = underflow ? Int64.min : diff
+        for entry in index.entries(modelHash: binding.modelHash) where entry.lastHitAt < cutoff {
+            let url = cacheDir.appendingPathComponent(
+                "\(modelDirComponent)/\(entry.digestHex).\(EncryptedKVStore.fileExtension)")
+            try? FileManager.default.removeItem(at: url)
+            index.remove(modelHash: binding.modelHash, digestHex: entry.digestHex)
+            stats.ttlExpirations += 1
+        }
```
ethenotethan, line 1108: 🔵 **[INFO]** ⚡ File deletion in reapExpired runs synchronously during lookup

💡 _Suggestion:_ Consider making file deletions asynchronous or batched to avoid blocking the lookup path when many files need cleanup

📊 _Score: 2×3 = 6 · Category: blocking I/O on hot paths_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -270,6 +290,58 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+    /// on recency, so the persistence urgency disappears.
+    private func persistRecencyIfDue(_ index: PrefixCacheIndex) {
+        guard ttlSeconds > 0, !closed, index.isDirty else { return }
+        let t = now()
+        guard t - lastTouchSaveAt >= Self.touchSaveIntervalSeconds else { return }
+        if (try? index.save()) != nil {
+            lastTouchSaveAt = t
+            unsavedWrites = 0
+        }
+    }
```
ethenotethan, line 339: 🟡 **[MEDIUM]** ⚡ Synchronous index save on hot path during touch operations

💡 _Suggestion:_ The time-coalesced save still blocks the request thread - consider making index.save() asynchronous or use a background queue

📊 _Score: 2×4 = 8 · Category: blocking I/O on hot paths_

provider-swift/Sources/ProviderCore/Inference/BatchScheduler+EngineBridge.swift
```diff
@@ -282,6 +282,27 @@ extension BatchScheduler {
...
+    }
```
ethenotethan, line 304: 🔵 **[INFO]** 🧩 TTL reaper setup duplicates pattern from stats logger with minor variations

💡 _Suggestion:_ Extract common periodic task setup pattern into a helper method to reduce duplication

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -270,6 +290,58 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+    /// shrunken index and pushes the corrected footprint to the accountant
+    /// (mirroring dropUnusableSSDFile). No-op when TTL disabled / SSD off /
+    /// closed.
+    public func reapExpiredTick() async {
+        guard ssdEnabled, let index, let cacheDir else { return }
+        let before = stats.ttlExpirations
+        reapExpired(index: index, cacheDir: cacheDir)
+        guard stats.ttlExpirations > before else { return }
+        if index.isDirty { _ = try? index.save() }
+        await notifyAccountant()
```
ethenotethan, line 355: 🟡 **[MEDIUM]** 🧩 Multiple small helper methods for TTL logic could be consolidated

💡 _Suggestion:_ Consider merging expiresAtForWrite, saturatingExpiry, and persistRecencyIfDue into fewer, more cohesive methods

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
```diff
@@ -427,9 +513,45 @@ public actor PrefixCacheManager: PrefixCacheOwner {
...
+                    stats.ttlExpirations += 1
+                    let rel = "\(modelDirComponent)/\(candidate.digestHex).\(EncryptedKVStore.fileExtension)"
+                    await dropUnusableSSDFile(
+                        cacheDir.appendingPathComponent(rel), digestHex: candidate.digestHex, index: index)
+                    continue
+                }
+            }
+            selected = candidate
+            break
+        }
```
ethenotethan, line 559: 🔴 **[CRITICAL]** 🧩 Complex nested loop with defensive caps and multiple exit conditions in loadFromSSD

💡 _Suggestion:_ Extract TTL expiry check into a separate method and simplify the control flow with early returns

📊 _Score: 4×4 = 16 · Category: additive complexity_

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
ethenotethan, line 1108: 🟡 **[MEDIUM]** 🧩 TTL expiry logic duplicated between reapExpired and loadFromSSD

💡 _Suggestion:_ Extract common TTL expiry check (age calculation and comparison) into a shared helper method

📊 _Score: 3×3 = 9 · Category: duplicate logic_

provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
```diff
@@ -942,6 +954,23 @@ public actor BatchScheduler {
...
```
ethenotethan, line 972: 🔵 **[INFO]** 🧩 TTL configuration adds another environment variable when existing patterns could be extended

💡 _Suggestion:_ Consider if TTL could be configured through existing cache configuration mechanisms rather than a separate env var

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler+EngineBridge.swift:291-303` — TTL reaper task lacks proper cancellation handling in error paths
  - _Suggestion:_ Add defer { ttlReapTask?.cancel() } or wrap the sleep/reap loop in a cancellation-aware pattern to ensure cleanup on task failure
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:520-540` — Potential O(n) loop in TTL expiry check during SSD lookup
  - _Suggestion:_ Consider adding an early break or limit on attempts to prevent excessive iteration when many entries are expired
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:1094-1106` — File deletion in reapExpired runs synchronously during lookup
  - _Suggestion:_ Consider making file deletions asynchronous or batched to avoid blocking the lookup path when many files need cleanup
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:318-326` — Synchronous index save on hot path during touch operations
  - _Suggestion:_ The time-coalesced save still blocks the request thread - consider making index.save() asynchronous or use a background queue

### Additive_complexity — 5 finding(s) (3 blocking)
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler+EngineBridge.swift:285-304` — TTL reaper setup duplicates pattern from stats logger with minor variations
  - _Suggestion:_ Extract common periodic task setup pattern into a helper method to reduce duplication
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:290-342` — Multiple small helper methods for TTL logic could be consolidated
  - _Suggestion:_ Consider merging expiresAtForWrite, saturatingExpiry, and persistRecencyIfDue into fewer, more cohesive methods
- 🔴 **[CRITICAL]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:513-553` — Complex nested loop with defensive caps and multiple exit conditions in loadFromSSD
  - _Suggestion:_ Extract TTL expiry check into a separate method and simplify the control flow with early returns
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:1089-1106` — TTL expiry logic duplicated between reapExpired and loadFromSSD
  - _Suggestion:_ Extract common TTL expiry check (age calculation and comparison) into a shared helper method
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:954-972` — TTL configuration adds another environment variable when existing patterns could be extended
  - _Suggestion:_ Consider if TTL could be configured through existing cache configuration mechanisms rather than a separate env var

_9 finding(s) total, 6 blocking. Verdict: REQUEST_CHANGES._

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
anupsv: Fixed in `31c9275c`. Extracted the expiry predicate into a single `isExpired(lastHitAt:)` (overflow-safe, TTL-gated) — the selection loop body is now a `guard isExpired else { selected; break }` plus the drop. The attempts cap and `closed` re-check stay (they guard real teardown races, documented inline), but the control flow reads linearly now. Strictly behavior-preserving; all 115 cache tests pass unchanged.

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
anupsv: Fixed in `31c9275c` — `isExpired(lastHitAt:)` is now the single source of truth for both `loadFromSSD` and `reapExpired` (the cutoff form is gone). Equivalence verified at every edge: boundary (`lastHitAt < now−ttl ⇔ now−lastHitAt > ttl`), corrupt `Int64.min` timestamp (reaped in both forms), future timestamp (kept), `Int64.max` TTL (keeps everything).

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
anupsv: Closed by `31c9275c` (same finding as the later review's duplicate-logic comment): `isExpired(lastHitAt:)` is now the shared predicate for the read path and both sweeps.

provider-swift/Sources/ProviderCore/Inference/BatchScheduler+EngineBridge.swift
anupsv: Declining — there is no error path to handle: the loop body is `try? await Task.sleep` (non-escaping) plus a non-throwing actor call, so the task cannot fail; it exits only via cancellation. A `defer { ttlReapTask?.cancel() }` inside the task would be a self-cancel no-op. Lifecycle is owned by the scheduler (cancelled in `stopCurrentEngine`, mirroring the established `prefixCacheStatsTask` pattern); two independent reviewers verified all teardown paths (model swap, unload, idle-eviction) cancel it.

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
anupsv: Declining — the suggested limit already exists: the loop is capped at `attempts = boundaries.count + 1` (≤ ~8 — one prompt's checkpoint ladder, never the index size), plus a `closed` re-check. It cannot iterate "many entries"; it considers only the ladder of checkpoints applicable to the one prompt being looked up.

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
anupsv: Declining — the premise doesn't hold: `reapExpired` never runs during lookup. It runs at load-time reconcile and on the periodic tick (both off the request path). The lookup path deletes at most one prompt's expired ladder candidates (bounded by `boundaries.count`), which is the intended self-healing-on-read, and reuses the same drop path corrupt files already take.

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
anupsv: Declining — measured against what it costs: the save is time-coalesced to at most one per 60s, the index is bounded by the disk budget (hundreds of entries ⇒ ms-scale encode+write), and it runs on the manager actor which already performs synchronous saves on every entry drop. Making it async would let the save race index mutation (the index is actor-confined by design) to shave ~1ms/min. If profiling ever shows it on a flame graph we can revisit with an actor-hop snapshot, but today the complexity buys nothing measurable.

provider-swift/Sources/ProviderCore/Inference/BatchScheduler+EngineBridge.swift
anupsv: Declining for now (rule of three): two ~15-line starters with different guards, cadence math, and bodies. A generic periodic-task helper would save ~5 lines and add indirection between the reader and the lifecycle. If a third periodic task appears we'll extract.

provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift
anupsv: Declining — these are three distinct concerns, not fragments of one: `expiresAtForWrite` is write-stamp policy, `saturatingExpiry` is the overflow primitive (it exists precisely to dedupe its two callers), `persistRecencyIfDue` is durability coalescing. Merging them produces a grab-bag method with a conditional personality. (The genuinely shared logic — the expiry predicate — did get consolidated: see `isExpired` in `31c9275c`.)

provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
anupsv: Declining — `DARKBLOOM_PREFIX_CACHE_TTL_SECONDS` IS the existing configuration mechanism: it follows the established `DARKBLOOM_PREFIX_CACHE_*` family exactly (`MAX_GB`, `DISK_GB`, `MIN_PERSIST_TOKENS`, `STATS_INTERVAL_SECS`), same resolver pattern, same precedence (env > default), documented in the same table in docs/ssd-kv-cache.md.

Gajesh2007 (approved):

PR #283, opened 2026-06-07
Background model prefetch + zero-downtime hot-swap (DAR-225)

Gajesh2007:
## Background model prefetch + zero-downtime hot-swap

Lets us move a public model name from one build (quant) to another with **no downtime** and **without consumers ever seeing the quant**. First user: migrate `gemma-4-26b` from `…-fp8` (~27 GB) to `…-qat-4bit` (~15.6 GB) — higher quality, lower memory, and at `min_ram_gb≈22` it opens the entire 32 GB Mac tier (fp8 was gated to 64 GB+).

Closes the work tracked in **DAR-225** (blocks DAR-189; subsumes DAR-154 download-resume; related DAR-175/DAR-123).

The design is **declarative**: an alias carries a single `desired_build` pointer (plus an optional still-acceptable `previous_build`). The coordinator pushes that desired state to providers; each provider converges on its own — background-prefetch the desired build, verify it on disk, then hard-swap off the old one. There are no routing weights, no ramp dial, and no central migration controller — convergence is just providers reconciling against the declared desired state.

### Before → After

**Before** — routing is by exact HuggingFace-path id, so the quant is part of the public name and swapping it breaks clients:

```mermaid
flowchart LR
    C["Consumer"] -->|"model: mlx-community/gemma-4-26b-a4b-it-fp8"| CO["Coordinator<br/>(exact-id match)"]
    CO --> P1["Provider · fp8 (27 GB)"]
    CO -.->|"swapping the quant means changing the public id"| X["clients break +<br/>manual, downtime-y cutover"]
```

**After** — consumers send a stable alias; the coordinator resolves it to a concrete build, preferring the Desired build when a provider can route it and falling back to the still-acceptable Previous build if Desired is momentarily saturated. The desired build is pushed declaratively; providers prefetch + verify it in the background and hard-swap once ready:

```mermaid
flowchart LR
    A["Admin"] -->|"set desired_build (declarative)"| CO["Coordinator"]
    C["Consumer"] -->|"model: gemma-4-26b (stable alias)"| CO
    CO -->|"ResolveModel: prefer Desired when routable"| P2["Provider · qat-4bit<br/>(Desired)"]
    CO -.->|"fall back to Previous if Desired is saturated"| P1["Provider · fp8<br/>(Previous)"]
    CO -->|"desired_models (declarative push)"| P1
    P1 -->|"prefetch + verify, then models_update (hash-gated)"| CO
    P1 -.->|"hard-swap: drop fp8 once qat verified"| P2
```

[...]

PR #299, opened 2026-06-10
fix(coordinator): never list a build behind a public alias (hide retired builds)

Gajesh2007:
## Never expose a concrete build behind a public alias

A consumer must only ever see the **alias** (e.g. `gemma-4-26b`) — the concrete quant build it resolves to (`…-fp8`, `…-qat-4bit`) is a backend routing detail and must never appear in a public listing.

### The bug
`/v1/models` and `/v1/models/openrouter` built their "hidden builds" set from only the alias's **desired** and **previous** builds. Once a build was **fully retired** — the operator clears `previous_build`, so the old build moves into the alias's `retired_builds` lineage but is left **registered/active** — it was no longer desired-or-previous, so it stopped being hidden and **reappeared as its own raw entry**.

Concretely, after retiring fp8 on the `gemma-4-26b` migration:

```
GET /v1/models →
  gemma-4-26b                                  ← the alias (correct)
  mlx-community/gemma-4-26b-a4b-it-fp8         ← ⚠️ the raw quant, leaked back
```

On the OpenRouter feed it's worse: the retired build is listed `is_ready: true` with **zero providers** → requests to it queue and 429 after the 120s timeout (a marketplace black-hole).

### The fix
Hide **every in-catalog build a public alias references** — desired, previous, **and** the retired lineage — in both listing helpers (`aliasModelEntries`, `openRouterAliasEntries`), regardless of whether the alias itself is advertisable right now.

- Only **in-catalog** builds are hidden — a build absent from the catalog can't appear in the listing anyway, and gating on catalog membership preserves the "desired build not yet registered" case (an alias mid-registration still doesn't hide a phantom id).
- Capacity still aggregates only the routable **desired/previous** members; retired builds are **hide-only** and never contribute capacity.

After the fix, `GET /v1/models` shows only `gemma-4-26b` at every stage — rollout, retirement, and after.

### Tests
- `TestListModelsHidesRetiredAliasBuild` and `TestOpenRouterModelsHidesRetiredAliasBuild` reproduce the exact leak (retired build still registered) and assert only the alias is listed. **Both verified to fail without the fix.**
- Existing alias-listing tests still pass (incl. `TestAliasModelEntriesDesiredNotInCatalog`, which pins the not-yet-registered-desired behavior).
- Full coordinator `go test ./...` + `go vet` clean.

Follow-up to #283 (DAR-225). Coordinator-only; no protocol or provider changes.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/299"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783643070&installation_id=133247490&pr_number=299&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F299&signature=638bbdc0217b39371a28b821914a8cfbf12492f6d6ca43cf7e41b7eb64ade6bd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -3378,13 +3378,31 @@ func buildNonStreamingResponse(requestID, model string, msg extractedMessage, us
 // including attestation metadata (trust level, Secure Enclave status,
 // provider count) for each model. Capacity fields (routable_providers,
 // warm_providers, can_accept) are included from the live capacity snapshot.
+// hideAliasBuild marks a concrete build id as hidden from the standalone model
+// listing — a build behind a public alias is only ever exposed through that
+// alias. Only builds that are actually in the catalog are hidden: a build absent
+// from the catalog can't appear in the listing anyway, and adding it would
+// pollute the hidden set (e.g. an alias pointing at a not-yet-registered build).
+// Empty ids are ignored.
+func hideAliasBuild(hidden map[string]struct{}, catalogByID map[string]store.SupportedModel, buildID string) {
+	if buildID == "" {
+		return
+	}
+	if _, inCatalog := catalogByID[buildID]; inCatalog {
+		hidden[buildID] = struct{}{}
+	}
+}
+
 // aliasModelEntries builds the consumer-facing /v1/models entries for active
 // public aliases and returns the set of underlying build ids those aliases
-// cover (so the caller can hide them from the default listing). Each alias entry
-// derives its metadata from its primary build — the desired build, or the
-// previous build if the desired one isn't in the catalog yet — and aggregates
-// live capacity across both the desired and previous builds so the alias's
-// routable/warm counts reflect every quant currently serving it.
+// cover (so the caller can hide them from the default listing). The hidden set
+// covers EVERY build an alias references — desired, previous, and the retired
+// lineage — so a concrete quant build never appears as its own entry once it is
+// behind an alias. Each alias entry derives its metadata from its primary build
+// — the desired build, or the previous build if the desired one isn't in the
+// catalog yet — and aggregates live capacity across the desired and previous
+// builds so the alias's routable/warm counts reflect every quant currently
+// serving it (retired builds are hide-only and never contribute capacity).
 func (s *Server) aliasModelEntries(
 	capByModel map[string]*registry.ModelCapacity,
 	catalogByID map[string]store.SupportedModel,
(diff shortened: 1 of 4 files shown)


coordinator/api/consumer.go
```diff
@@ -3378,13 +3378,31 @@ func buildNonStreamingResponse(requestID, model string, msg extractedMessage, us
...
+// Empty ids are ignored.
+func hideAliasBuild(hidden map[string]struct{}, catalogByID map[string]store.SupportedModel, buildID string) {
+	if buildID == "" {
+		return
+	}
+	if _, inCatalog := catalogByID[buildID]; inCatalog {
+		hidden[buildID] = struct{}{}
+	}
+}
+
```
ethenotethan, line 3395: 🟡 **[MEDIUM]** ⚡ hideAliasBuild performs redundant catalog lookup for every build ID

💡 _Suggestion:_ Pre-filter build IDs to non-empty values before calling hideAliasBuild, or combine the empty check with catalog lookup in a single operation

coordinator/api/consumer.go
```diff
@@ -3378,13 +3378,31 @@ func buildNonStreamingResponse(requestID, model string, msg extractedMessage, us
...
+// pollute the hidden set (e.g. an alias pointing at a not-yet-registered build).
+// Empty ids are ignored.
+func hideAliasBuild(hidden map[string]struct{}, catalogByID map[string]store.SupportedModel, buildID string) {
+	if buildID == "" {
+		return
+	}
+	if _, inCatalog := catalogByID[buildID]; inCatalog {
+		hidden[buildID] = struct{}{}
+	}
+}
```
ethenotethan, line 3394: 🔵 **[INFO]** 🧩 hideAliasBuild helper function adds indirection for simple map assignment

💡 _Suggestion:_ Inline the logic directly in the loops where it's used - the function is only 6 lines and used in 3 places, making it harder to follow the flow

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:3387-3395` — hideAliasBuild performs redundant catalog lookup for every build ID
  - _Suggestion:_ Pre-filter build IDs to non-empty values before calling hideAliasBuild, or combine the empty check with catalog lookup in a single operation

### Additive_complexity — 1 finding(s)
- 🔵 **[INFO]** `coordinator/api/consumer.go:3387-3394` — hideAliasBuild helper function adds indirection for simple map assignment
  - _Suggestion:_ Inline the logic directly in the loops where it's used - the function is only 6 lines and used in 3 places, making it harder to follow the flow

_2 finding(s) total, 1 blocking. Verdict: REQUEST_CHANGES._

PR #297, opened 2026-06-09
fix(coordinator+provider): kill false "model swap" untrusts on busy multi-model providers

Gajesh2007:
# Fix false "model swap" hard-untrusts on busy multi-model providers

## The incident

Two of our busiest providers (both dual-model: `gemma-4-26b` + `gpt-oss-20b`, both on v0.5.16, both with **bit-perfect weights on disk**) were repeatedly hard-untrusted with `active model hash mismatch — possible model swap`. Restarts didn't help ("untrusted a few minutes after start"), re-downloading didn't help, and a local verifier script run on the box confirmed the disk matched the published manifests exactly.

The "wrong" hash the coordinator kept logging — `61bfc04e…` — turned out to be **the correct, official aggregate hash of gpt-oss-20b**.

## Root cause: a race in the coordinator's active-model check

The challenge response's `active_model_hash` is a bare hash — it does not say which model it describes. To validate it, the coordinator *guessed* the model from the provider's **last heartbeat** (up to a full heartbeat interval stale):

```
16:24:06.3Z  gemma request finishes            ← last heartbeat: "current = gemma"
16:24:06.5Z  gpt-oss request dispatched        ← provider's current model is now gpt-oss
16:24:08.4Z  challenge arrives; provider answers in 190 ms with
             active_model_hash = hash(gpt-oss) = 61bfc04e…   ← correct, for gpt-oss
             coordinator compares against catalog[gemma] = a4722b60…
             mismatch → "model swap" → MarkUntrusted (hard)
```

On a multi-model box both models are loaded and serving simultaneously — "current model" is just *whichever model served the last request*, a label that flips every few seconds under interleaved traffic. The provider answers about its current model *now*; the coordinator validates against the model it heard about *up to 30 s ago*. When traffic flips the model inside that window, a perfectly correct hash of model B is graded against model A's answer key.

The bug only fires on **busy multi-model** providers (31 of 69 fleet providers are dual-model and exposed; the two highest-scored boxes hit it first because routing gives them the densest interleaving). Single-model providers can never hit it — which is why "other machines on the same version have no issues."

A hard untrust also stops the challenge loop, so one race hit sticks until the provider reconnects.

## The fix (coordinator side — deploys without touching the fleet)

v0.5.16 providers **already send the model-keyed map** `model_hashes = {model_id: hash}` in every challenge response. The coordinator now validates that map entry-by-entry against the catalog — exact, race-free, and strictly stronger than checking only one model:

[...]

PR #294, opened 2026-06-09
feat(coordinator): admin-gated /data state-export for EigenCloud→GCP migration (DAR-70)

Gajesh2007:
## Summary

Adds an **admin-gated endpoint to export the coordinator's TEE-sealed on-disk state (`/data`)** as a downloadable archive, so the prod coordinator can be migrated off the EigenCloud TEE onto a GCP Confidential VM **without re-enrolling the provider fleet**.

This is the **DAR-70** piece of the EigenCloud→GCP migration: the step-ca root + intermediate **CA private keys** and the MicroMDM **BoltDB device database** are born inside the TEE on first boot and have never been exfiltrated — this endpoint is the only way to lift them out so the existing fleet keeps authenticating after the move.

## Endpoint

`GET /v1/admin/state-export` — triple-gated, fail-closed:

1. **`EIGENINFERENCE_STATE_EXPORT_ENABLED`** (default `false`) → returns **404** when off (route inert until toggled + restart).
2. **`isAdminAuthorized`** — reuses the existing constant-time admin-key / Privy-admin check.
3. **Encrypted by default** — if `EIGENINFERENCE_STATE_EXPORT_RECIPIENT` (an `age1…` recipient) is set, output is age-encrypted (`darkbloom-state-<ts>.zip.age`); a raw zip requires `EIGENINFERENCE_STATE_EXPORT_ALLOW_PLAINTEXT=true`, otherwise **412**.

Root resolution: `EIGENINFERENCE_STATE_EXPORT_ROOT` → `USER_PERSISTENT_DATA_PATH` → `/mnt/disks/userdata`.

## What it captures

Zips `step-ca/**` (config, secrets incl. password, certs) and `micromdm/**` (BoltDB, push/server certs, the hidden **`.push_imported`** sentinel), excluding `*.log`. Resolves the symlinked `/data` root. Streams the archive (no full-archive buffering).

## Consistency — the hard part

MicroMDM holds an exclusive bbolt lock on its live DB, so a naive copy can capture a torn write → a corrupt snapshot would silently de-route the fleet on restore. The export is **two-phase**:

[...]

PR #292, opened 2026-06-09
feat(provider): darkbloom restart + graceful download-first auto-update

Gajesh2007:
## Why

The provider's background auto-update kills in-flight inference. Today (`ProviderLoop.performAutoUpdateCheck`):

1. The 30-min monitor **skips entirely** if any request is in flight (`if !requestToModel.isEmpty { return }`) — a busy provider may never update.
2. When idle it calls `SelfUpdater.update()` (check + download + verify + install in one shot). The idle check is a snapshot at the *start*; new requests are accepted and served *during* the multi-second/minute download, the new binary lands on disk under them, then `restartAfterUpdate()` bounces the process — **cutting off active requests**.
3. There's no first-class `restart` verb, and `darkbloom start` (no args) is interactive (TUI picker), so operators can't cleanly bounce the existing selection.

## Flow: before vs after

**Before** — the idle check is a one-time snapshot; requests served during the download get bounced by the restart:

```mermaid
flowchart TD
    T[Auto-update tick · every 30m] --> Q{Any request in flight?}
    Q -->|Yes| SKIP[Skip entirely — retry in 30m]
    Q -->|No| DL[check + download + verify + install<br/>single shot]
    DL --> RS[restartAfterUpdate]
    RS --> BOUNCE[Process bounces]
    NEW[New request arrives<br/>during download] -. accepted + served .-> BOUNCE
    BOUNCE -.->|killed mid-generation| DROP[(In-flight request dropped)]
```

**After** — download happens while still serving; we only stop accepting and drain *after* a successful install, then hot-swap:

```mermaid
flowchart TD
    T[Auto-update tick · every 30m] --> CHK[check for update<br/>runs while serving]
    CHK --> AV{Update available?}
    AV -->|No| IDLE[stay idle / keep serving]
    AV -->|Yes| INST[download + verify + install<br/>WHILE STILL SERVING]
    INST --> OK{Install succeeded?}
    OK -->|No| RESUME[resume serving<br/>zero downtime · retry next tick]
    OK -->|Yes| DRAIN[begin draining<br/>refuse new reqs · 503 reroute / local 503]
    DRAIN --> WAIT[wait for in-flight to reach zero<br/>up to updateDrainTimeout = 120s]
    WAIT --> D{Drained?}
    D -->|Yes| HS[hot-swap restart]
    D -->|No · 120s timeout| FC[force-cancel stragglers] --> HS
    NEW[New request during install] -. still accepted + served .-> INST
```

[...]

PR #291, opened 2026-06-09
feat: coordinated drain + restart for provider auto-updates

Gajesh2007:
## Summary

Implements coordinated drain + restart for provider auto-updates. When a new release is registered, the coordinator fans out drain commands to all online providers, which stop accepting new requests, finish in-flight work, apply the binary update via SelfUpdater, and restart.

### Changes

**Protocol (Go + Swift):**
- New message types: `drain_command` (coordinator->provider), `provider_draining` (provider->coordinator)
- `DrainReason` enum: `update`, `maintenance`, `decommission`
- Heartbeat includes `draining`, `drain_reason`, `drain_deadline`

**Provider (ProviderLoop.swift):**
- Drain state machine: `isDraining`, `drainDeadline`, `drainReason`, progress counters
- Admission gate: rejects new requests with 503 + Retry-After during drain
- `runDrainMonitor()`: waits for in-flight completion or deadline, sends progress updates
- `restartAfterUpdate()`: calls `ProcessLifecycle.restartAfterUpdate()` (launchd bootout/bootstrap/kickstart or execv)
- Runs `SelfUpdater.update()` before restart to apply binary update
- Deadline parsing: tries fractional seconds first, falls back to Go time.RFC3339

**Coordinator Registry (registry.go):**
- `Draining` field on Provider struct
- `UpdateProviderDraining()`: updates state from draining messages
- `GetOnlineProviders()`: returns online providers for fan-out
- `Disconnect()`: clears drain state on reconnect

**Coordinator Scheduler (scheduler.go):**
- `snapshotProviderLocked`: skips draining providers
- `providerCanAdmitLocked`: skips draining providers
- `QuickCapacityCheck`: skips draining providers
- `ModelCapacitySnapshot`: skips draining providers

**Release Handlers (release_handlers.go):**
- `triggerProviderDrainForRelease()`: fans out drain commands to all online providers on new release
- Feature flag: `ENABLE_DRAIN_UPDATE=false` to disable (enabled by default)

### Verification
- Coordinator: `go build ./...`, `go test ./...` all pass
- Provider: `swift build`, `swift test` pass (non-MLX tests; MLX error is test-env only)
- gofmt clean

### Follow-up (not blocking)
- Add canary/batching to fleet drain (currently all-at-once)
- Add drain command authentication (signature/nonce)
- Add test coverage for drain flow
- Make SelfUpdater respect commanded version
- Make drainTimeoutSecs authoritative
- Restart only on successful .updated (not on failure/no-op)

---
[...]

PR #303, opened 2026-06-10
v0.6.0 launch: APNs grace, vision routing, 4-bit cutover, graceful update

Gajesh2007:
## v0.6.0 launch — fixes for all seven launch objectives

Implements the code-level fixes from the pre-release audit so v0.6.0 is **safe to deploy** and every launch objective can be turned on operationally afterward. This is one PR; **no tag/release** is created here (release stays human-gated).

Branch base: master @ `8adfe150`.

### What's in it

| Objective | Change |
|---|---|
| **APNs binary verification** | Code-attestation enforcement is now **deadline-driven**: `SetCodeAttestor` only *configures* (challenge + measure); enforcement begins at `APNS_ENFORCE_AFTER` (RFC3339). Unset ⇒ perpetual grace/observe, so setting `APNS_*` secrets **never deroutes the fleet**. After the deadline, un-attested (and too-old-to-attest) providers stop routing — "everyone must have it" after the 1-day grace. The gate reads policy live, so grace→enforce flips with **no provider reconnect**. Added periodic re-challenge so a single dropped background push doesn't strand a capable provider. |
| **Gemma 8-bit → 4-bit migration** | Coordinator **alias takeover**: a public alias can adopt the live concrete id `gemma-4-26b`, absorbing it as the `previous_build` fallback while `desired_build` points at the new 4-bit build. Opt-in, fail-closed, and **does not change the absorbed model's catalog weight hash** (so it never untrusts the live fleet; asserted by test). This is the missing bootstrap for the cutover — inert until an admin invokes it. |
| **SSD KV caching** | Already default-on in 0.6.0; this PR makes the documented opt-out **reachable** by passing `DARKBLOOM_PREFIX_CACHE` through the launchd plist `EnvironmentVariables` (launchd doesn't inherit the installing shell). |
| **Gemma 4 multimodal** | **Vision-aware routing** so a media request is never silently answered image-blind: providers advertise `is_vision` per build (from `vision_config`); the coordinator detects image/video parts (it sees plaintext at routing time) and only routes to a vision-capable provider, failing fast with a clear error when none exists. Media-aware **routing** token estimate (flat per-image cost, not base64 length). Console upload button now lights only from catalog modality (the `/gemma-?4/i` family bridge is removed), so the operator's capability flip is the single switch. **WS 1 MiB cap fix**: the provider raises `maximumMessageSize` to 32 MiB so an image frame can't tear down the whole session and cancel unrelated in-flight requests. |
| **Graceful downloads / swap / update** | Shipped in #292/#298; this PR carries the **0.6.0 version bump** and the alias hot-swap bootstrap above. |
| **Model hashing fix** | Shipped in #297; no code change here — see deploy guardrails (catalog freeze). |
| **EigenCloud → GCP migration** | State-export shipped in #294; no change here. |

Plus **DAR-130**: tool JSON schemas are normalized at the inbound decode boundary (default-fill a missing `type` on each parameter) so a Gemma chat template's `{{ value['type'] | upper }}` can't 500 — provider-side only, the `libs/mlx-swift-lm` submodule is untouched.

### Review

Both quality-gate reviewers (codex + Claude) ran. The first codex pass flagged three real issues in vision routing — a `p.Models` data race (the helper read it without `p.mu`), a billing under-reservation (media-aware count dropped `tool_calls`/`name` fields), and a Responses-API media-detection gap. All three are fixed in `63bdf5f3` and re-verified resolved. Billing now stays a guaranteed upper bound; routing/ITPM is the media-aware estimate.

### Tests

- Coordinator: `go build ./...` + `go test ./...` — 15 packages, 0 failures (new tests pass under `-race`).
- Provider: `swift build` + `swift test` — 678 tests pass.
- Console: `next build` + `eslint` (0 errors) + 240 vitest pass.

New coverage: APNs grace/enforce/live-flip/re-challenge, `is_vision` protocol symmetry, vision routing helpers + fail-fast, media detection (chat + Responses) + media-aware estimate, alias-takeover (incl. catalog-hash-unchanged), tool-schema defaulting, WS cap, launchd env passthrough.

---

## ⚠️ Deploy guardrails (operational — read before shipping)

These are runtime/KMS actions, not code, and gate the safe rollout:

- **KMS — leave unset this cycle:** `APNS_KEY_ID` / `APNS_TEAM_ID` / `APNS_AUTH_KEY_P8_*` (configuring them starts grace/observe; only set `APNS_ENFORCE_AFTER = release + 24h` once the fleet is on 0.6.0 and attesting), `EIGENINFERENCE_MIN_PROVIDER_VERSION` (don't bump — would deroute the <0.6.0 fleet), `EIGENINFERENCE_BINARYHASH_ENFORCE` (default-off is the intended new posture), `EIGENINFERENCE_STATE_EXPORT_ENABLED`.
- **Deploy order:** coordinator first, **catalog frozen** (no publish/promote/status ops) until ~2× 5-min challenge cycles verify clean (`grep "model weight hash mismatch"`, capacity stable). Then register the provider release in a low-traffic window (the ≤0.5.16 updater is drainless).
- **Turn features on last, after fleet convergence:** publish the 4-bit build under a **new** id (inert), then the alias takeover; flip catalog `vision`/`video` capabilities + ship the console image button only once the gemma fleet is on 0.6.0.
- **Size-limit chain (images):** provider WS 32 MiB > Caddy `request_body` 25 MB > coordinator sealed-body cap **16 MiB** (the effective single-image ceiling). Consistent for single-image launch; raise the coordinator's 16 MiB sealed cap first (then Caddy) if multi-image support is wanted later.

## Out of scope (follow-up PRs)

- #291 heartbeat `draining` flag (next release cycle — first fleet-wide graceful drain is 0.6.0 → next).
- Metallib template-hash **union** (newest-wins today; 0.6.0's metallib is byte-identical so safe this release — fix before any MLX-bumping release).
- Remote (coordinator-pushed) SSD-cache kill switch; durable console image storage; batched multimodal; making Responses-API images actually serve (today they're gated + fail-fast, the conversion still strips them).

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/303"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783663597&installation_id=133247490&pr_number=303&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F303&signature=fd91067de20cdb36933437462951129d427e3303fb8a486c0be8b100f9ee5065"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/api/code_attest_throttle.go
+++ b/coordinator/api/code_attest_throttle.go
@@ -1,0 +1,97 @@
+package api
+
+import (
+	"sync"
+	"time"
+)
+
+// codeAttestThrottle keeps APNs code-identity pushes within Apple's background-
+// push budget and reuses a recent attestation across reconnects, so the
+// coordinator never spams silent pushes.
+//
+// Apple throttles silent/background notifications to roughly 2-3 per device per
+// hour and drops the rest; we are forced to use background pushes (an alert push
+// would require UNUserNotificationCenter authorization, which breaks the
+// no-notification-center invariant). So the coordinator must NOT re-challenge on
+// a fixed interval — attestation is per-connection (the binary cannot change
+// without the process, and thus the WebSocket, restarting), so a single challenge
+// per connection suffices, with bounded retries only on delivery failure.
+//
+// Keyed by the Secure Enclave public key — the stable per-device identity that
+// survives reconnects. Two knobs:
+//   - reuseWindow: how long a successful attestation is honored for a NEW
+//     connection from the same device+version without re-pushing. Bounds the
+//     staleness of the proof (a malicious binary swap within the window could ride
+//     a prior attestation), so it is kept short and version-gated. Within a single
+//     live connection the proof is exact regardless of this window.
+//   - pushCooldown: minimum spacing between pushes to the same device — the hard
+//     rate-limit backstop. At 20m that is <= 3 pushes/hour/device even under
+//     reconnect churn and retries.
+type codeAttestThrottle struct {
+	mu       sync.Mutex
+	attested map[string]codeAttestRecord // seKey -> last successful attestation
+	lastPush map[string]time.Time        // seKey -> last push (device-level rate limit)
+
+	reuseWindow  time.Duration
+	pushCooldown time.Duration
+	maxAttempts  int
+	now          func() time.Time
+}
+
+type codeAttestRecord struct {
+	at      time.Time
+	version string
+}
+
+func newCodeAttestThrottle() *codeAttestThrottle {
+	return &codeAttestThrottle{
+		attested:     make(map[string]codeAttestRecord),
+		lastPush:     make(map[string]time.Time),
+		reuseWindow:  30 * time.Minute,
+		pushCooldown: 20 * time.Minute, // <= 3 pushes/hour/device (APNs background budget)
+		maxAttempts:  3,
+		now:          time.Now,
+	}
+}
+
+// reuseAttestation reports whether the device attested recently with the SAME
+// binary version, so a fresh connection can inherit the proof without a push.
+func (t *codeAttestThrottle) reuseAttestation(seKey, version string) bool {
+	if seKey == "" {
+		return false
+	}
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	r, ok := t.attested[seKey]
+	return ok && r.version == version && t.now().Sub(r.at) < t.reuseWindow
+}
+
+// allowPush reports whether enough time has passed since the last push to this
+// device to send another within the background-push budget.
+func (t *codeAttestThrottle) allowPush(seKey string) bool {
+	if seKey == "" {
+		return true // no device identity to throttle on; fall back to the loop's cap
+	}
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	last, ok := t.lastPush[seKey]
+	return !ok || t.now().Sub(last) >= t.pushCooldown
+}
+
+func (t *codeAttestThrottle) recordPush(seKey string) {
+	if seKey == "" {
+		return
+	}
+	t.mu.Lock()
+	t.lastPush[seKey] = t.now()
+	t.mu.Unlock()
+}
+
+func (t *codeAttestThrottle) recordAttested(seKey, version string) {
+	if seKey == "" {
+		return
+	}
+	t.mu.Lock()
+	t.attested[seKey] = codeAttestRecord{at: t.now(), version: version}
+	t.mu.Unlock()
+}
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -505,7 +505,7 @@ func approximateTokenCountUpperBound(v any) int {
 func estimatePromptTokens(parsed map[string]any) int {
 	total := 0
 	if v, ok := parsed["messages"]; ok {
-		total += messagesPromptTokens(v, false)
+		total += messagesPromptTokens(v)
 	}
 	if v, ok := parsed["input"]; ok {
 		total += approximateTokenCount(v)
@@ -526,7 +526,14 @@ func estimatePromptTokens(parsed map[string]any) int {
 func estimateBillingPromptTokens(parsed map[string]any) int {
 	total := 0
 	if v, ok := parsed["messages"]; ok {
-		total += messagesPromptTokens(v, true)
+		// Billing MUST stay a guaranteed upper bound (len(bytes) >= tokens for any
+		// BPE tokenizer), so it keeps counting full message bytes — including a
+		// base64 image's bytes and every non-content field (role, tool_calls,
+		// name). Switching to the media-aware flat count here would DROP those
+		// fields and under-reserve for tool-calling requests. Over-reservation on a
+		// large image is safe (it is refunded after inference); the routing/ITPM
+		// estimate (estimatePromptTokens) is the media-aware one.
+		total += approximateTokenCountUpperBound(v)
 	}
 	if v, ok := parsed["input"]; ok {
 		total += approximateTokenCountUpperBound(v)
@@ -560,17 +567,15 @@ func isMediaPartType(t string) bool {
 	return false
 }
 
-// messageContentTokens estimates prompt tokens for one message's `content`,
-// counting text parts as text and each image/video part as a flat media cost
-// (never the base64 length). byteBound selects the billing upper bound (len) vs
-// the routing heuristic (len/4) for text.
-func messageContentTokens(content any, byteBound bool) int {
+// messageContentTokens estimates ROUTING prompt tokens for one message's
+// `content`, counting text parts as text (len/4) and each image/video part as a
+// flat media cost (never the base64 length). Used only for the routing/ITPM
+// estimate; billing uses approximateTokenCountUpperBound (a guaranteed upper
+// bound that intentionally still counts the base64 bytes).
+func messageContentTokens(content any) int {
 	textTokens := func(s string) int {
 		if s == "" {
 			return 0
-		}
-		if byteBound {
-			return len(s)
 		}
 		if t := len(s) / 4; t > 0 {
 			return t
@@ -599,74 +604,75 @@ func messageContentTokens(content any, byteBound bool) int {
 				total += videoPromptTokenCost
 			default:
 				if b, err := json.Marshal(pm); err == nil {
-					if byteBound {
-						total += len(b)
-					} else {
-						total += len(b) / 4
-					}
+					total += len(b) / 4
 				}
 			}
 		}
 		return total
 	default:
-		if byteBound {
-			return approximateTokenCountUpperBound(content)
-		}
 		return approximateTokenCount(content)
 	}
 }
 
-// messagesPromptTokens sums media-aware content tokens across a messages array.
-// Falls back to the byte heuristic when messages isn't the standard array shape.
-func messagesPromptTokens(messages any, byteBound bool) int {
+// messagesPromptTokens sums media-aware routing content tokens across a messages
+// array. Falls back to the len/4 heuristic when messages isn't the standard
+// array shape.
+func messagesPromptTokens(messages any) int {
 	arr, ok := messages.([]any)
 	if !ok {
-		if byteBound {
-			return approximateTokenCountUpperBound(messages)
-		}
 		return approximateTokenCount(messages)
 	}
 	total := 0
 	for _, m := range arr {
 		mm, ok := m.(map[string]any)
 		if !ok {
-			if byteBound {
-				total += approximateTokenCountUpperBound(m)
-			} else {
-				total += approximateTokenCount(m)
-			}
+			total += approximateTokenCount(m)
 			continue
 		}
 		total += 4 // small per-message framing (role + delimiters)
-		total += messageContentTokens(mm["content"], byteBound)
+		total += messageContentTokens(mm["content"])
 	}
 	return total
 }
 
-// detectMediaRequirement reports whether the request carries image/video input in
-// any message's content parts. The coordinator sees plaintext at this point
-// (sealedTransport decrypts before the handler), so this drives the vision
-// routing gate and the fail-fast "no vision-capable provider" response.
-func detectMediaRequirement(parsed map[string]any) bool {
-	messages, ok := parsed["messages"].([]any)
+// contentPartsHaveMedia reports whether a `content` value (a content-part array)
+// carries any image/video part.
+func contentPartsHaveMedia(content any) bool {
+	parts, ok := content.([]any)
 	if !ok {
 		return false
 	}
-	for _, m := range messages {
-		mm, ok := m.(map[string]any)
+	for _, part := range parts {
+		pm, ok := part.(map[string]any)
 		if !ok {
 			continue
 		}
-		parts, ok := mm["content"].([]any)
-		if !ok {
-			continue
+		if typ, _ := pm["type"].(string); isMediaPartType(typ) {
+			return true
 		}
-		for _, part := range parts {
-			pm, ok := part.(map[string]any)
-			if !ok {
-				continue
+	}
+	return false
+}
+
+// detectMediaRequirement reports whether the request carries image/video input.
+// The coordinator sees plaintext at this point (sealedTransport decrypts before
+// the handler), so this drives the vision routing gate and the fail-fast "no
+// vision-capable provider" response. It scans both the Chat Completions
+// `messages[].content` parts and the Responses API `input[].content` parts so a
+// media request on either surface is gated (never silently routed text-blind).
+func detectMediaRequirement(parsed map[string]any) bool {
+	if messages, ok := parsed["messages"].([]any); ok {
+		for _, m := range messages {
+			if mm, ok := m.(map[string]any); ok && contentPartsHaveMedia(mm["content"]) {
+				return true
 			}
-			if typ, _ := pm["type"].(string); isMediaPartType(typ) {
+		}
+	}
+	// Responses API: `input` may be a string (no media) or an array of items,
+	// each carrying `content` parts in the same image_url/input_image shape.
+	if input, ok := parsed["input"].([]any); ok {
+		for _, item := range input {
+			if im, ok := item.(map[string]any); ok && contentPartsHaveMedia(im["content"]) {
 				return true
 			}
 		}
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -561,7 +561,9 @@ const (
 // image or video input.
 func isMediaPartType(t string) bool {
 	switch t {
-	case "image_url", "input_image", "video_url", "input_video":
+	// OpenAI chat (image_url/video_url), OpenAI Responses (input_image/input_video),
+	// and Anthropic /v1/messages content blocks ({"type":"image"|"video","source":…}).
+	case "image_url", "input_image", "image", "video_url", "input_video", "video":
 		return true
 	}
 	return false
@@ -598,9 +600,9 @@ func messageContentTokens(content any) int {
 				if s, ok := pm["text"].(string); ok {
 					total += textTokens(s)
 				}
-			case typ == "image_url" || typ == "input_image":
+			case typ == "image_url" || typ == "input_image" || typ == "image":
 				total += imagePromptTokenCost
-			case typ == "video_url" || typ == "input_video":
+			case typ == "video_url" || typ == "input_video" || typ == "video":
 				total += videoPromptTokenCost
 			default:
 				if b, err := json.Marshal(pm); err == nil {
@@ -1251,11 +1253,24 @@ func (s *Server) handleChatCompletions(w http.ResponseWriter, r *http.Request) {
 	// when the fleet has no such provider (e.g. before the gemma fleet finishes
 	// updating to 0.6.0); the routing layer enforces the same gate per dispatch.
 	requiresVision := detectMediaRequirement(parsed)
-	if requiresVision && !s.registry.HasVisionProviderForModel(model) {
-		writeJSON(w, http.StatusServiceUnavailable, errorResponse("model_unavailable",
-			fmt.Sprintf("model %q has no vision-capable provider available for image/video input right now", publicModel),
-			withParam("model")))
-		return
+	if requiresVision {
+		// The Responses API path lowers `input` to chat messages via
+		// responsesRequestToChatCompletions, which does NOT carry image/video parts
+		// through — so a media request there would be routed and then silently
+		// stripped (image-blind). Reject it cleanly until that conversion preserves
+		// media (tracked follow-up); the console uses /v1/chat/completions for images.
+		if input != nil && len(messages) == 0 {
+			writeJSON(w, http.StatusBadRequest, errorResponse("invalid_request_error",
+				"image/video input via the Responses API is not supported yet; use /v1/chat/completions",
+				withParam("input")))
+			return
+		}
+		if !s.registry.HasVisionProviderForModel(model) {
+			writeJSON(w, http.StatusServiceUnavailable, errorResponse("model_unavailable",
+				fmt.Sprintf("model %q has no vision-capable provider available for image/video input right now", publicModel),
+				withParam("model")))
+			return
+		}
 	}
 
 	isResponsesAPI := input != nil && len(messages) == 0
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -239,6 +239,7 @@ func (s *Server) dispatchOneProvider(
 	reservedMicroUSD int64,
 	estimatedPromptTokens int,
 	requestedMaxTokens int,
+	requiresVision bool,
 	allowedProviderSerials []string,
 	isResponsesAPI bool,
 	policy selfRoutePolicy,
@@ -263,6 +264,7 @@ func (s *Server) dispatchOneProvider(
 		ConsumerLocation:       consumerLocation,
 		IsResponsesAPI:         isResponsesAPI,
 		EstimatedPromptTokens:  estimatedPromptTokens,
+		RequiresVision:         requiresVision,
 		RequestedMaxTokens:     requestedMaxTokens,
 		ReservedMicroUSD:       reservedMicroUSD,
 		BaseReservedMicroUSD:   reservedMicroUSD,
@@ -503,7 +505,7 @@ func approximateTokenCountUpperBound(v any) int {
 func estimatePromptTokens(parsed map[string]any) int {
 	total := 0
 	if v, ok := parsed["messages"]; ok {
-		total += approximateTokenCount(v)
+		total += messagesPromptTokens(v, false)
 	}
 	if v, ok := parsed["input"]; ok {
 		total += approximateTokenCount(v)
@@ -524,7 +526,7 @@ func estimatePromptTokens(parsed map[string]any) int {
 func estimateBillingPromptTokens(parsed map[string]any) int {
 	total := 0
 	if v, ok := parsed["messages"]; ok {
-		total += approximateTokenCountUpperBound(v)
+		total += messagesPromptTokens(v, true)
 	}
 	if v, ok := parsed["input"]; ok {
 		total += approximateTokenCountUpperBound(v)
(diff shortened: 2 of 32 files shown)


coordinator/api/consumer.go
```diff
@@ -538,6 +547,139 @@ func estimateBillingPromptTokens(parsed map[string]any) int {
...
+const (
+	imagePromptTokenCost = 300
+	videoPromptTokenCost = 1500
+)
+
+// isMediaPartType reports whether an OpenAI/OpenRouter content-part type denotes
+// image or video input.
+func isMediaPartType(t string) bool {
+	switch t {
+	case "image_url", "input_image", "video_url", "input_video":
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 49fdc5a8 — `isMediaPartType` now recognizes Anthropic `image`/`video` content blocks (and the media-aware token estimate counts them), so an Anthropic `/v1/messages` image request is gated to a vision provider instead of routed text-blind.

coordinator/registry/registry.go
```diff
@@ -1496,6 +1553,43 @@ func (r *Registry) providerServesCatalogModelLocked(p *Provider, model string) b
...
+// updating to 0.6.0), instead of queueing the request to a timeout.
+func (r *Registry) HasVisionProviderForModel(model string) bool {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	for _, p := range r.providers {
+		// p.Status and p.Models are guarded by p.mu (writers hold it), so the
+		// whole eligibility read must happen under the provider lock.
+		p.mu.Lock()
+		eligible := p.Status != StatusOffline && p.Status != StatusUntrusted &&
+			r.providerServesVisionModelLocked(p, model)
```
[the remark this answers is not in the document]
Gajesh2007: Correctness is held by the reservation path: `ReserveProviderEx` applies full eligibility (trust floor, challenge freshness, private-only, capacity) AND the per-build vision gate, so a vision request can never be served by an ineligible or text-only provider. `HasVisionProviderForModel` is only a best-effort fast-fail for the error message when *zero* vision providers exist; an existing-but-unroutable one just yields a normal capacity/queue path, not an image-blind answer. Resolving as precision-only.

coordinator/api/consumer.go
```diff
@@ -1103,6 +1245,19 @@ func (s *Server) handleChatCompletions(w http.ResponseWriter, r *http.Request) {
 	}
 	model, rawBody = buildModel, resolvedBody
 
+	// Vision gating: a request carrying image/video input must land on a provider
+	// advertising a vision-capable (VLM) build of this model, or the media is
+	// silently dropped and the answer is image-blind. Fail fast with a clear error
+	// when the fleet has no such provider (e.g. before the gemma fleet finishes
+	// updating to 0.6.0); the routing layer enforces the same gate per dispatch.
+	requiresVision := detectMediaRequirement(parsed)
+	if requiresVision && !s.registry.HasVisionProviderForModel(model) {
```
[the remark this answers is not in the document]
Gajesh2007: Valid for the future alias+vision case. Today `gemma-4-26b` is a concrete model (not an alias) and vision is off at launch, so this is inert. Once the 4-bit alias cutover lands and vision is enabled, alias resolution should prefer a vision-capable build for media requests. Leaving open as a tracked follow-up for that cutover.

coordinator/api/consumer.go
```diff
@@ -538,6 +547,139 @@ func estimateBillingPromptTokens(parsed map[string]any) int {
...
+			if mm, ok := m.(map[string]any); ok && contentPartsHaveMedia(mm["content"]) {
+				return true
+			}
+		}
+	}
+	// Responses API: `input` may be a string (no media) or an array of items,
+	// each carrying `content` parts in the same image_url/input_image shape.
+	if input, ok := parsed["input"].([]any); ok {
+		for _, item := range input {
+			if im, ok := item.(map[string]any); ok && contentPartsHaveMedia(im["content"]) {
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 49fdc5a8 — image/video requests on `/v1/responses` are now rejected with a clear 400 (the `responsesRequestToChatCompletions` lowering doesn't carry media parts, so routing-then-stripping would be image-blind). The console uses `/v1/chat/completions` for images. Preserving media through the conversion is a tracked follow-up.

coordinator/api/consumer.go
```diff
@@ -538,6 +547,139 @@ func estimateBillingPromptTokens(parsed map[string]any) int {
...
+	}
+	total := 0
+	for _, m := range arr {
+		mm, ok := m.(map[string]any)
+		if !ok {
+			total += approximateTokenCount(m)
+			continue
+		}
+		total += 4 // small per-message framing (role + delimiters)
+		total += messageContentTokens(mm["content"])
```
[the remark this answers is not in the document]
Gajesh2007: Real but minor: this is the ROUTING heuristic only — the billing reservation uses `approximateTokenCountUpperBound`, which counts the full message (incl. `tool_calls`), so there's no under-charge. A large `tool_calls` history slightly under-estimates routing/admission tokens; the provider does its own fit check. Leaving open as a minor refinement.

coordinator/registry/scheduler.go
```diff
@@ -320,13 +340,30 @@ func (r *Registry) selectBestCandidateLockedFull(model string, pr *PendingReques
...
+		// vision-capable build of this model. Providers reach here only if they
+		// already serve the model (snapshot ok), so a miss here means "serves it,
+		// but text-only" — counted separately so the caller can return a precise
+		// "no vision-capable provider" error rather than a busy/429. snapshot
+		// released p.mu, so re-take it for the p.Models read.
+		if pr.RequiresVision {
+			p.mu.Lock()
+			servesVision := r.providerServesVisionModelLocked(p, model)
+			p.mu.Unlock()
+			if !servesVision {
```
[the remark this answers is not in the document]
Gajesh2007: Correctness is held by the reservation gate + the fast-fail: a vision request can't be served by a text-only provider regardless of the preflight, and `HasVisionProviderForModel` clean-errors when zero vision providers exist. `QuickCapacityCheck` only affects the early 429/queue signal (it may count text providers), so at worst a slightly less precise busy signal, never a wrong route. Resolving as precision-only.

provider-swift/Sources/ProviderCore/Inference/ToolSchemaNormalization.swift
```diff
@@ -0,0 +1,80 @@
...
+            dict["properties"] != nil || dict["items"] != nil ||
+            dict["enum"] != nil || dict["description"] != nil ||
+            dict["anyOf"] != nil || dict["oneOf"] != nil || dict["allOf"] != nil
+        if dict["type"] == nil, looksLikeSchemaNode {
+            if dict["properties"] != nil {
+                dict["type"] = "object"
+            } else if dict["items"] != nil {
+                dict["type"] = "array"
+            } else {
+                dict["type"] = "string"
```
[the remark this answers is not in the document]
Gajesh2007: Acknowledged tradeoff: defaulting a typeless `anyOf`/`oneOf` node to `string` is deliberate crash-prevention for the Gemma `value['type'] | upper` template (a typeless union would still crash it); it does lose union semantics. This only affects tool calls on the 4-bit Gemma build (a later cutover), so leaving open to refine then rather than risk re-introducing the crash now.

coordinator/api/consumer.go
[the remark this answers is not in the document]
Gajesh2007: Same class as the fast-fail above — the reservation path applies the serial-allowlist / self-route constraint AND the vision gate together, so a constrained media request is never served by an out-of-set or text-only provider. The global `HasVisionProviderForModel` probe only affects pre-dispatch error precision, not routing correctness. Resolving as precision-only.

coordinator/api/consumer.go
```diff
@@ -503,7 +505,7 @@ func approximateTokenCountUpperBound(v any) int {
 func estimatePromptTokens(parsed map[string]any) int {
 	total := 0
 	if v, ok := parsed["messages"]; ok {
-		total += approximateTokenCount(v)
+		total += messagesPromptTokens(v)
 	}
 	if v, ok := parsed["input"]; ok {
 		total += approximateTokenCount(v)
```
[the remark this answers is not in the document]
Gajesh2007: Resolved by the Responses-media rejection in 49fdc5a8 — a `/v1/responses` request carrying `input_image`/`input_video` is now 400'd before token estimation, so the inflated `input` estimate no longer occurs. Non-media `input` continues to use the text estimator, which is correct.

provider-swift/Sources/ProviderCore/Coordinator/CoordinatorClient.swift
```diff
@@ -408,6 +408,23 @@ final class ReachabilityMonitor: @unchecked Sendable {
...
 public actor CoordinatorClient {
+    /// Upper bound on a single inbound WebSocket message. The coordinator sends each
+    /// inference request as ONE text frame carrying the base64 NaCl-box of the full
+    /// body; for a vision request that frame includes base64-encoded image bytes and
+    /// can be several MiB. `URLSessionWebSocketTask` defaults to a 1 MiB cap and
+    /// THROWS on any larger frame, which tears down the entire session and cancels
+    /// every unrelated in-flight request on this provider (then reconnects with
+    /// backoff). Size this comfortably above the coordinator's 16 MiB sealed-body cap
+    /// after base64 expansion (×4/3 ≈ 21.3 MiB).
+    static let maxInboundMessageBytes = 32 * 1024 * 1024
```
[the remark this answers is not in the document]
Gajesh2007: The 32 MiB provider WS cap is sized above the binding upstream limit: the coordinator caps the sealed request body at 16 MiB (`sender_encryption.go`), and the WS message is `base64(box(plaintext))` of that ≤16 MiB body ≈ ≤21 MiB < 32 MiB. The console's 4×10 MB allowance is rejected with a 400 at the 16 MiB sealed cap *before* reaching the WS, so it's not a teardown vector. The console-side cap (which should be ≤~11 MB pre-base64 to match) is a separate console follow-up. Resolving the WS-cap thread.

provider-swift/Sources/ProviderCore/Inference/ToolSchemaNormalization.swift
```diff
@@ -0,0 +1,80 @@
...
+        for key in ["anyOf", "oneOf", "allOf"] {
+            if let variants = dict[key] as? [Any] {
+                dict[key] = variants.map(injectDefaultTypes)
+            }
+        }
+
+        let looksLikeSchemaNode =
+            dict["properties"] != nil || dict["items"] != nil ||
+            dict["enum"] != nil || dict["description"] != nil ||
+            dict["anyOf"] != nil || dict["oneOf"] != nil || dict["allOf"] != nil
```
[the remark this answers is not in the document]
Gajesh2007: Valid — the pass should recurse into `additionalProperties` (it already recurses into array `items`) so a map-shaped param's inner schema gets a default `type`. Leaving open; I'll fix it in a follow-up provider build to avoid a concurrent build with other in-flight provider work.

provider-swift/Sources/ProviderCore/Inference/ToolSchemaNormalization.swift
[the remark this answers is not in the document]
Gajesh2007: Improved in this push: a typeless union now borrows the first concrete (non-null) member type (e.g. nullable number → "number") instead of a blanket "string", so union intent is preserved while the template still gets a usable type.

provider-swift/Sources/ProviderCore/Inference/ToolSchemaNormalization.swift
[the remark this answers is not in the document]
Gajesh2007: Fixed in this push: the pass now recurses into additionalProperties (and types a typeless map node as "object"), so map-shaped params no longer leave an inner schema without a type.

coordinator/api/code_attest_throttle.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (t *codeAttestThrottle) recordAttested(seKey, version string) {
+	if seKey == "" {
+		return
+	}
+	t.mu.Lock()
+	t.attested[seKey] = codeAttestRecord{at: t.now(), version: version}
+	t.mu.Unlock()
+}
```
ethenotethan, line 97: 🔵 **[INFO]** 🔒 Time-based throttling could leak timing information about device attestation patterns

💡 _Suggestion:_ Consider adding jitter to throttling intervals to prevent timing analysis

📊 _Score: 2×2 = 4 · Category: timing-side-channel_

coordinator/api/consumer.go
```diff
@@ -538,6 +547,141 @@ func estimateBillingPromptTokens(parsed map[string]any) int {
...
+	total := 0
+	for _, m := range arr {
+		mm, ok := m.(map[string]any)
+		if !ok {
+			total += approximateTokenCount(m)
+			continue
+		}
+		total += 4 // small per-message framing (role + delimiters)
+		total += messageContentTokens(mm["content"])
+	}
```
ethenotethan, line 636: 🟡 **[MEDIUM]** 🔒 Media detection relies on string matching without proper validation of content parts structure

💡 _Suggestion:_ Add strict validation of content part structure before accessing nested fields to prevent potential type confusion attacks

📊 _Score: 3×3 = 9 · Category: input-validation_

coordinator/api/model_alias_handlers.go
```diff
@@ -19,6 +19,16 @@ type aliasUpsertRequest struct {
...
 	Active        *bool  `json:"active"` // pointer so omission defaults to true
+	// Takeover lets a public alias adopt the name of an EXISTING concrete model,
+	// absorbing that same-named build as its previous_build (fallback). This is the
+	// only way to migrate a live public name (e.g. "gemma-4-26b") onto a new
+	// desired build without renaming what providers already advertise — used for
+	// the 8-bit→4-bit gemma cutover. Fail-closed: only the exact shape
+	// alias_id == previous_build == an existing concrete model id is permitted, and
+	// desired_build must still be a distinct registered build. It does NOT touch
+	// the absorbed model's catalog weight hash, so it never untrusts the providers
+	// already serving it; they converge to the desired build via desired_models.
```
ethenotethan, line 30: 🔵 **[INFO]** 🔒 Takeover functionality could potentially be abused to hijack model names

💡 _Suggestion:_ Add additional authorization checks and audit logging for takeover operations

📊 _Score: 2×2 = 4 · Category: auth-bypass_

coordinator/cmd/coordinator/main.go
```diff
@@ -524,6 +536,25 @@ func main() {
...
+		return time.Time{}
+	}
+	t, err := time.Parse(time.RFC3339, raw)
+	if err != nil {
+		logger.Error("APNS_ENFORCE_AFTER is not valid RFC3339 — leaving APNs in grace mode (no enforcement)",
+			"value", raw, "error", err)
+		return time.Time{}
+	}
+	return t
+}
```
ethenotethan, line 567: 🔵 **[INFO]** 🔒 RFC3339 timestamp parsing falls back to grace mode on malformed input without strict validation

💡 _Suggestion:_ Consider failing fast on malformed APNS_ENFORCE_AFTER values rather than silently falling back to grace mode

📊 _Score: 2×3 = 6 · Category: input-validation_

coordinator/api/consumer.go
```diff
@@ -503,7 +505,7 @@ func approximateTokenCountUpperBound(v any) int {
 func estimatePromptTokens(parsed map[string]any) int {
```
ethenotethan, line 505: 🔵 **[INFO]** ⚡ messagesPromptTokens called twice for same messages data

💡 _Suggestion:_ Cache the result of messagesPromptTokens(v) and reuse it in estimateBillingPromptTokens to avoid duplicate computation

📊 _Score: 2×3 = 6 · Category: repeated work_

coordinator/api/consumer.go
```diff
@@ -1103,6 +1247,32 @@ func (s *Server) handleChatCompletions(w http.ResponseWriter, r *http.Request) {
...
+		}
+		if !s.registry.HasVisionProviderForModel(model) {
+			writeJSON(w, http.StatusServiceUnavailable, errorResponse("model_unavailable",
+				fmt.Sprintf("model %q has no vision-capable provider available for image/video input right now", publicModel),
+				withParam("model")))
+			return
+		}
+	}
+
 	isResponsesAPI := input != nil && len(messages) == 0
```
ethenotethan, line 1276: 🟡 **[MEDIUM]** ⚡ Vision capability check blocks request processing

💡 _Suggestion:_ Move HasVisionProviderForModel check to async validation or cache results to avoid blocking the hot path

📊 _Score: 3×4 = 12 · Category: blocking I/O_

coordinator/api/consumer.go
```diff
@@ -4306,6 +4477,16 @@ func (s *Server) handleGenericInference(w http.ResponseWriter, r *http.Request,
...
 	}
 
+	// Vision gating (see handleChatCompletions): a media request must land on a
+	// vision-capable provider, or fail fast when none exists.
+	requiresVision := detectMediaRequirement(parsed)
+	if requiresVision && !s.registry.HasVisionProviderForModel(model) {
+		writeJSON(w, http.StatusServiceUnavailable, errorResponse("model_unavailable",
+			fmt.Sprintf("model %q has no vision-capable provider available for image/video input right now", publicModel),
+			withParam("model")))
+		return
```
ethenotethan, line 4487: 🟡 **[MEDIUM]** ⚡ Duplicate vision capability check in generic inference handler

💡 _Suggestion:_ Extract vision validation to shared function or middleware to avoid duplicate registry lookups

📊 _Score: 3×4 = 12 · Category: blocking I/O_

coordinator/registry/scheduler.go
```diff
@@ -320,13 +340,30 @@ func (r *Registry) selectBestCandidateLockedFull(model string, pr *PendingReques
...
+				visionRejections++
+				continue
+			}
```
ethenotethan, line 356: 🟡 **[MEDIUM]** ⚡ Vision capability check repeated for each provider in routing loop

💡 _Suggestion:_ Pre-filter providers by vision capability before the main routing loop to avoid N checks per request

📊 _Score: 2×4 = 8 · Category: repeated work_

coordinator/registry/registry.go
```diff
@@ -1496,6 +1553,43 @@ func (r *Registry) providerServesCatalogModelLocked(p *Provider, model string) b
...
+		p.mu.Unlock()
+		if eligible {
+			return true
+		}
+	}
```
ethenotethan, line 1589: 🔵 **[INFO]** ⚡ HasVisionProviderForModel iterates all providers without early exit optimization

💡 _Suggestion:_ Add early return when first vision-capable provider is found instead of checking all providers

📊 _Score: 2×3 = 6 · Category: repeated work_

coordinator/api/consumer.go
ethenotethan, line 505: 🟡 **[MEDIUM]** 🏷️ Function approximateTokenCountUpperBound accepts bare any parameter without type constraints

💡 _Suggestion:_ Add type assertion or use a more specific interface type instead of any

📊 _Score: 3×4 = 12 · Category: bare_interface_

coordinator/api/consumer.go
```diff
@@ -524,6 +526,13 @@ func estimatePromptTokens(parsed map[string]any) int {
 func estimateBillingPromptTokens(parsed map[string]any) int {
```
ethenotethan, line 526: 🟡 **[MEDIUM]** 🏷️ Function messagesPromptTokens accepts bare any parameter without type constraints

💡 _Suggestion:_ Add type assertion or use a more specific interface type instead of any

📊 _Score: 3×4 = 12 · Category: bare_interface_

coordinator/api/consumer.go
```diff
@@ -538,6 +547,141 @@ func estimateBillingPromptTokens(parsed map[string]any) int {
...
+	}
+	return false
+}
+
+// messageContentTokens estimates ROUTING prompt tokens for one message's
+// `content`, counting text parts as text (len/4) and each image/video part as a
+// flat media cost (never the base64 length). Used only for the routing/ITPM
+// estimate; billing uses approximateTokenCountUpperBound (a guaranteed upper
+// bound that intentionally still counts the base64 bytes).
+func messageContentTokens(content any) int {
```
ethenotethan, line 577: 🟡 **[MEDIUM]** 🏷️ Function messageContentTokens accepts bare any parameter without type constraints

💡 _Suggestion:_ Add type assertion or use a more specific interface type instead of any

📊 _Score: 3×4 = 12 · Category: bare_interface_

coordinator/api/consumer.go
```diff
@@ -538,6 +547,141 @@ func estimateBillingPromptTokens(parsed map[string]any) int {
...
+		}
+		total += 4 // small per-message framing (role + delimiters)
+		total += messageContentTokens(mm["content"])
+	}
+	return total
+}
+
+// contentPartsHaveMedia reports whether a `content` value (a content-part array)
+// carries any image/video part.
+func contentPartsHaveMedia(content any) bool {
```
ethenotethan, line 642: 🟡 **[MEDIUM]** 🏷️ Function contentPartsHaveMedia accepts bare any parameter without type constraints

💡 _Suggestion:_ Add type assertion or use a more specific interface type instead of any

📊 _Score: 3×4 = 12 · Category: bare_interface_

coordinator/api/consumer.go
```diff
@@ -538,6 +547,141 @@ func estimateBillingPromptTokens(parsed map[string]any) int {
...
+		if !ok {
+			continue
+		}
+		if typ, _ := pm["type"].(string); isMediaPartType(typ) {
+			return true
+		}
+	}
+	return false
+}
+
```
ethenotethan, line 658: 🟡 **[MEDIUM]** 🏷️ Function detectMediaRequirement accepts map[string]any parameter

💡 _Suggestion:_ Define a typed struct for the parsed request instead of using map[string]any

📊 _Score: 3×4 = 12 · Category: bare_interface_

coordinator/api/code_attest_throttle.go
ethenotethan, line 97: 🟡 **[MEDIUM]** 🧩 New throttle abstraction with injectable time function for testing

💡 _Suggestion:_ Consider using real time.Now() and controlling time in tests via test helpers or context deadlines rather than injecting a time function

coordinator/api/consumer.go
ethenotethan, line 636: 🔵 **[INFO]** 🧩 Media detection and token estimation logic duplicated across multiple functions

💡 _Suggestion:_ Extract shared media detection logic into a single helper function used by both detectMediaRequirement and messageContentTokens

coordinator/api/consumer.go
```diff
@@ -1103,6 +1247,32 @@ func (s *Server) handleChatCompletions(w http.ResponseWriter, r *http.Request) {
...
+				"image/video input via the Responses API is not supported yet; use /v1/chat/completions",
+				withParam("input")))
+			return
+		}
+		if !s.registry.HasVisionProviderForModel(model) {
+			writeJSON(w, http.StatusServiceUnavailable, errorResponse("model_unavailable",
+				fmt.Sprintf("model %q has no vision-capable provider available for image/video input right now", publicModel),
+				withParam("model")))
+			return
+		}
```
ethenotethan, line 1273: 🔵 **[INFO]** 🧩 Vision routing logic mixed into consumer handler

💡 _Suggestion:_ Move vision capability checking to registry layer where other routing decisions are made

coordinator/registry/registry.go
```diff
@@ -309,9 +316,9 @@ func providerSupportsPrivateTextLocked(p *Provider) bool {
 		return false
 	}
 	// v0.6.0 APNs code-identity gate — the SINGLE chokepoint, no self-route
```
ethenotethan, line 318: 🟡 **[MEDIUM]** 🧩 Complex APNs attestation policy with multiple overlapping configuration knobs

💡 _Suggestion:_ Simplify to a single configuration method that sets both configured flag and deadline atomically, reducing state combinations

📊 _Score: 3×4 = 12 · Category: over-configuration_

provider-swift/Sources/ProviderCore/Inference/ToolSchemaNormalization.swift
```diff
@@ -0,0 +1,107 @@
...
+            for variant in variants {
+                if let v = variant as? [String: Any],
+                    let t = v["type"] as? String, t != "null" {
+                    return t
+                }
+            }
+        }
+        return nil
+    }
+}
```
ethenotethan, line 116: 🟡 **[MEDIUM]** 🧩 Custom JSON schema manipulation instead of using existing JSON schema libraries

💡 _Suggestion:_ Consider using a dedicated JSON schema validation/normalization library rather than custom JSON manipulation logic

📊 _Score: 4×3 = 12 · Category: reinventing stdlib_

coordinator/api/model_alias_handlers.go
```diff
@@ -58,15 +68,37 @@ func (s *Server) handleModelAliasUpsert(w http.ResponseWriter, r *http.Request)
...
+				"takeover requires previous_build to equal alias_id (the concrete build absorbed as the alias fallback)", withParam("previous_build")))
+			return
+		}
+	} else if req.PreviousBuild == req.AliasID {
+		// No concrete model of this name exists, so there is nothing to absorb.
+		writeJSON(w, http.StatusBadRequest, errorResponse("invalid_request_error",
+			"an alias cannot reference itself", withParam("previous_build")))
 		return
 	}
 	if req.PreviousBuild != "" && req.PreviousBuild == req.DesiredBuild {
```
ethenotethan, line 104: 🟡 **[MEDIUM]** 🧩 Complex takeover logic with multiple validation branches and edge cases

💡 _Suggestion:_ Simplify takeover validation by extracting common validation patterns into helper functions

📊 _Score: 3×3 = 9 · Category: over-configuration_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 4 finding(s) (1 blocking)
- 🔵 **[INFO]** `coordinator/api/code_attest_throttle.go:1-97` — Time-based throttling could leak timing information about device attestation patterns
  - _Suggestion:_ Consider adding jitter to throttling intervals to prevent timing analysis
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:547-636` — Media detection relies on string matching without proper validation of content parts structure
  - _Suggestion:_ Add strict validation of content part structure before accessing nested fields to prevent potential type confusion attacks
- 🔵 **[INFO]** `coordinator/api/model_alias_handlers.go:19-30` — Takeover functionality could potentially be abused to hijack model names
  - _Suggestion:_ Add additional authorization checks and audit logging for takeover operations
- 🔵 **[INFO]** `coordinator/cmd/coordinator/main.go:536-556` — RFC3339 timestamp parsing falls back to grace mode on malformed input without strict validation
  - _Suggestion:_ Consider failing fast on malformed APNS_ENFORCE_AFTER values rather than silently falling back to grace mode

### Performance — 5 finding(s) (3 blocking)
- 🔵 **[INFO]** `coordinator/api/consumer.go:505` — messagesPromptTokens called twice for same messages data
  - _Suggestion:_ Cache the result of messagesPromptTokens(v) and reuse it in estimateBillingPromptTokens to avoid duplicate computation
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:1247-1276` — Vision capability check blocks request processing
  - _Suggestion:_ Move HasVisionProviderForModel check to async validation or cache results to avoid blocking the hot path
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:4477-4487` — Duplicate vision capability check in generic inference handler
  - _Suggestion:_ Extract vision validation to shared function or middleware to avoid duplicate registry lookups
- 🟡 **[MEDIUM]** `coordinator/registry/scheduler.go:340-356` — Vision capability check repeated for each provider in routing loop
  - _Suggestion:_ Pre-filter providers by vision capability before the main routing loop to avoid N checks per request
- 🔵 **[INFO]** `coordinator/registry/registry.go:1553-1589` — HasVisionProviderForModel iterates all providers without early exit optimization
  - _Suggestion:_ Add early return when first vision-capable provider is found instead of checking all providers

### Type_diligence — 5 finding(s) (5 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:505` — Function approximateTokenCountUpperBound accepts bare any parameter without type constraints
  - _Suggestion:_ Add type assertion or use a more specific interface type instead of any
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:526` — Function messagesPromptTokens accepts bare any parameter without type constraints
  - _Suggestion:_ Add type assertion or use a more specific interface type instead of any
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:577` — Function messageContentTokens accepts bare any parameter without type constraints
  - _Suggestion:_ Add type assertion or use a more specific interface type instead of any
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:642` — Function contentPartsHaveMedia accepts bare any parameter without type constraints
  - _Suggestion:_ Add type assertion or use a more specific interface type instead of any
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:658` — Function detectMediaRequirement accepts map[string]any parameter
  - _Suggestion:_ Define a typed struct for the parsed request instead of using map[string]any

### Additive_complexity — 6 finding(s) (4 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/code_attest_throttle.go:1-97` — New throttle abstraction with injectable time function for testing
  - _Suggestion:_ Consider using real time.Now() and controlling time in tests via test helpers or context deadlines rather than injecting a time function
- 🔵 **[INFO]** `coordinator/api/consumer.go:526-636` — Media detection and token estimation logic duplicated across multiple functions
  - _Suggestion:_ Extract shared media detection logic into a single helper function used by both detectMediaRequirement and messageContentTokens
- 🔵 **[INFO]** `coordinator/api/consumer.go:1247-1273` — Vision routing logic mixed into consumer handler
  - _Suggestion:_ Move vision capability checking to registry layer where other routing decisions are made
- 🟡 **[MEDIUM]** `coordinator/registry/registry.go:292-318` — Complex APNs attestation policy with multiple overlapping configuration knobs
  - _Suggestion:_ Simplify to a single configuration method that sets both configured flag and deadline atomically, reducing state combinations
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/Inference/ToolSchemaNormalization.swift:1-107` — Custom JSON schema manipulation instead of using existing JSON schema libraries
  - _Suggestion:_ Consider using a dedicated JSON schema validation/normalization library rather than custom JSON manipulation logic
- 🟡 **[MEDIUM]** `coordinator/api/model_alias_handlers.go:71-104` — Complex takeover logic with multiple validation branches and edge cases
  - _Suggestion:_ Simplify takeover validation by extracting common validation patterns into helper functions

_20 finding(s) total, 13 blocking. Verdict: REQUEST_CHANGES._

coordinator/cmd/coordinator/main.go
Gajesh2007: Good catch — fixed in 76b7e854. A non-empty but malformed APNS_ENFORCE_AFTER now returns an error and the coordinator refuses to start (rather than silently falling back to perpetual grace, which would keep un-attested providers routable forever). Unset still means grace, the intended safe default.

PR #272, opened 2026-06-03
feat: honor reasoning_effort and return accurate reasoning_tokens

Gajesh2007:
## Summary

Closes two customer-reported gaps on gpt-oss / reasoning models.

### 1. `reasoning_effort` is now honored (gpt-oss low/medium/high)
Previously the param was silently dropped — the provider decoded into the upstream `OpenAIChatCompletionRequest` shape (which doesn't model `reasoning_effort`) and always passed `additionalContext: nil` to the chat template, so the gpt-oss harmony template stayed at its `medium` default. That's why a customer saw no difference between low/high.

The provider now decodes `reasoning_effort` from the raw body (`extractReasoningEffort`) and threads it through `MultiModelBatchSchedulerEngine` into `applyChatTemplate(additionalContext:)`. The swift-transformers renderer injects each `additionalContext` key directly into the Jinja context, so the harmony template reads `reasoning_effort` and emits the matching `Reasoning: <effort>` system directive. Values pass through verbatim — the valid set is model-specific and lives in each template, so there's no hardcoded allowlist. (`minimal`/`xhigh` are GPT-5-family values, not gpt-oss, so they remain no-ops on gpt-oss by design.)

### 2. Accurate `reasoning_tokens`
Chat completions returned no reasoning breakdown; the Responses API reported **all** completion tokens as reasoning (a 2–4× overcount). The provider now tokenizes the actual reasoning content with the model tokenizer (clamped to ≤ `completion_tokens`) and surfaces it as:
- `usage.completion_tokens_details.reasoning_tokens` — chat completions (streaming, non-streaming reconstruction, and raw passthrough)
- `output_tokens_details.reasoning_tokens` — Responses API

The count rides the structured `UsageInfo` and is injected into the streaming usage SSE frame (which the coordinator forwards verbatim).

## Protocol / compatibility
`UsageInfo` gains `reasoning_tokens` in **both** `coordinator/protocol/messages.go` (Go, `omitempty`) and `provider-swift/.../Protocol/Types.swift` (Swift, `decodeIfPresent`). Old↔new provider/coordinator mixes are safe; `resolveReasoningTokens` falls back to the prior estimate for providers that predate accurate counting.

## Not changed (out of scope, deliberately)
- Billing still consumes only `prompt_tokens` + `completion_tokens`.
- Response-hash / attestation domain is unaffected (only the usage frame is rewritten, which carries no content/reasoning deltas).

## Testing
[...]

PR #304, opened 2026-06-10
fix: report reasoning_tokens separately in chat-completions streaming usage

Gajesh2007:
## Problem
Reasoning tokens were being lumped into `completion_tokens` with **no separate breakdown** for **streaming chat completions** — the most common path (chat UI, OpenAI SDK streaming). Consumers got `completion_tokens` (which includes reasoning) but no `completion_tokens_details.reasoning_tokens`.

## Root cause
The provider already counts reasoning tokens accurately and reports them via `UsageInfo.ReasoningTokens` over the WS (→ `CompleteCh`). The **non-streaming** chat path and **both Responses API** paths already splice that into the usage. But the **chat-completions streaming** handler (`handleStreamingResponseWithFirstChunk`) only forwarded the provider's terminal usage chunk verbatim — which carries `{prompt,completion,total}` but no `completion_tokens_details` — and never read `CompleteCh`. So streaming chat dropped the reasoning breakdown. (Added in #272, wired everywhere except chat streaming.)

## Fix
The streaming handler now **holds** the terminal `include_usage` chunk (empty `choices` + non-null `usage`) and re-emits it at stream end with `completion_tokens_details.reasoning_tokens` spliced in from the authoritative `UsageInfo` (`CompleteCh`) — reusing the existing `injectReasoningDetailIntoRawUsage`. OpenAI semantics preserved: `completion_tokens` stays the full output; `reasoning_tokens` is the subset.

- Single usage chunk (held + augmented, not doubled).
- No-op when there's no reasoning count or `stream_options.include_usage` wasn't requested.
- Backward compatible (an old provider reporting 0 reasoning → no detail).
- Hot-path safe: the per-chunk check early-returns without parsing for content deltas and `usage:null` chunks; only the one real usage chunk is parsed.

## Tests
- Helper unit test (`isUsageOnlyStreamChunk` / `injectReasoningIntoStreamUsageChunk`).
- Handler-level streaming test asserting `reasoning_tokens:8` in the final usage chunk, `completion_tokens` unchanged, exactly one usage chunk, content delta + `[DONE]` preserved.
- Full `api` suite green.

## Out of scope (optional follow-ups, found during investigation)
These are analytics/persistence, not the API output: usage records (`RecordUsage*`) don't persist `reasoning_tokens`; no `inference.reasoning_tokens` Datadog metric; console billing view doesn't show the reasoning split. Billing is already correct (reasoning is billed as completion — they're generated tokens).

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/304"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783670352&installation_id=133247490&pr_number=304&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F304&signature=0cb7a19f84ab6fdc44ddb22af87f7fdc50203a87218676e5bdd054ff3f18db11"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -2517,10 +2517,11 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r 
 	timer := time.NewTimer(inferenceTimeout)
 	defer timer.Stop()
 
-	// The terminal include_usage chunk lacks the reasoning breakdown; we hold it
-	// and re-emit it at stream end augmented with the provider's authoritative
-	// reasoning-token count (CompleteCh) — matching the non-streaming/Responses paths.
-	var pendingUsageChunk string
+	// The terminal include_usage chunk lacks the reasoning breakdown; we hold its
+	// parsed object and re-emit it at stream end with the provider's authoritative
+	// reasoning count (CompleteCh) spliced in — matching the non-streaming/Responses
+	// paths. Held as a parsed map so it is decoded exactly once.
+	var pendingUsage map[string]any
 
 	for {
 		select {
@@ -2555,7 +2556,11 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r 
 					// Emit the held usage chunk (stream_options.include_usage) with the
 					// reasoning breakdown spliced in from the provider's authoritative
 					// UsageInfo, so streaming reports reasoning_tokens like the rest.
-					if pendingUsageChunk != "" {
+					// This select runs once, at stream end: the provider's
+					// inferenceComplete (which populates CompleteCh) is what ends the
+					// stream, so it is effectively already buffered — the timeout is a
+					// fallback, not a hot-path wait.
+					if pendingUsage != nil {
 						var usage protocol.UsageInfo
 						select {
 						case u, uok := <-pr.CompleteCh:
@@ -2565,11 +2570,10 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r 
 						case <-time.After(2 * time.Second):
 						case <-r.Context().Done():
 						}
-						out := normalizeSSEChunk(pendingUsageChunk)
-						out = injectReasoningIntoStreamUsageChunk(out, usage)
-						out = rewriteChunkModel(out, pr)
-						fmt.Fprintf(w, "%s\n\n", out)
-						flusher.Flush()
+						if out := finalizeUsageChunk(pendingUsage, usage, pr); out != "" {
+							fmt.Fprintf(w, "%s\n\n", out)
+							flusher.Flush()
+						}
 					}
 					// Chat completions format: append SE signature + [DONE].
 					if pr.SESignature != "" {
@@ -2593,17 +2597,13 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r 
 			}
 			// Hold the terminal usage chunk (chat completions only) so we can splice
 			// in the reasoning breakdown at stream end; forwarding it inline would
-			// emit it without reasoning_tokens.
-			if !sawResponsesAPI && isUsageOnlyStreamChunk(chunk) {
-				pendingUsageChunk = chunk
-				if !timer.Stop() {
-					select {
-					case <-timer.C:
-					default:
-					}
+			// emit it without reasoning_tokens. The stream closes right after, so the
+			// running timer still guards a post-usage hang.
+			if !sawResponsesAPI {
+				if obj, isUsage := parseUsageOnlyStreamChunk(chunk); isUsage {
+					pendingUsage = obj
+					continue
 				}
-				timer.Reset(inferenceTimeout)
-				continue
 			}
 			if !sawResponsesAPI {
 				chunk = normalizeSSEChunk(chunk)
@@ -3381,43 +3381,44 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
 	obj["usage"] = usageObj
 }
 
-// isUsageOnlyStreamChunk reports whether an SSE data chunk is the terminal usage
-// chunk emitted for stream_options.include_usage: empty choices + a non-null usage
-// object, carrying the final usage and no content delta.
-func isUsageOnlyStreamChunk(chunk string) bool {
+// parseUsageOnlyStreamChunk decodes a terminal include_usage chunk (empty choices
+// + a non-null usage object, carrying the final usage and no content delta) and
+// returns the parsed object. ok is false for any other chunk. Parsing here once
+// lets the caller hold the object and finalize it at stream end without re-parsing.
+func parseUsageOnlyStreamChunk(chunk string) (obj map[string]any, ok bool) {
 	line := strings.TrimPrefix(chunk, "data: ")
+	// Cheap gate: skip the parse for content deltas and usage:null chunks.
 	if !strings.Contains(line, `"usage"`) || strings.Contains(line, `"usage":null`) {
-		return false
+		return nil, false
 	}
-	var obj map[string]any
 	if err := json.Unmarshal([]byte(line), &obj); err != nil {
-		return false
+		return nil, false
 	}
-	if u, ok := obj["usage"].(map[string]any); !ok || u == nil {
-		return false
+	if u, uok := obj["usage"].(map[string]any); !uok || u == nil {
+		return nil, false
 	}
-	choices, _ := obj["choices"].([]any)
-	return len(choices) == 0
+	if choices, _ := obj["choices"].([]any); len(choices) != 0 {
+		return nil, false
+	}
+	return obj, true
 }
 
-// injectReasoningIntoStreamUsageChunk adds completion_tokens_details.reasoning_tokens
-// (from the provider's authoritative UsageInfo) to a streaming chat-completions
-// usage chunk, so streaming reports the reasoning breakdown like the non-streaming
-// and Responses paths. No-op when there is no reasoning count or the chunk isn't
-// parseable.
-func injectReasoningIntoStreamUsageChunk(chunk string, usage protocol.UsageInfo) string {
-	if usage.ReasoningTokens <= 0 {
-		return chunk
-	}
-	line := strings.TrimPrefix(chunk, "data: ")
-	var obj map[string]any
-	if err := json.Unmarshal([]byte(line), &obj); err != nil {
-		return chunk
-	}
+// finalizeUsageChunk renders the held terminal usage chunk for chat-completions
+// streaming: it splices the provider's authoritative reasoning count into
+// completion_tokens_details (no-op when there is none), strips a null
+// system_fingerprint, and rewrites the build id to the public alias — marshalling
+// ONCE (obj is already parsed). Returns "" if it can't be marshalled.
+func finalizeUsageChunk(obj map[string]any, usage protocol.UsageInfo, pr *registry.PendingRequest) string {
 	injectReasoningDetailIntoRawUsage(obj, usage)
+	if v, present := obj["system_fingerprint"]; present && v == nil {
+		delete(obj, "system_fingerprint")
+	}
+	if pr.PublicModel != "" && pr.PublicModel != pr.Model {
+		obj["model"] = pr.PublicModel
+	}
 	b, err := json.Marshal(obj)
 	if err != nil {
-		return chunk
+		return ""
 	}
 	return "data: " + string(b)
 }
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -2517,6 +2517,11 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r 
 	timer := time.NewTimer(inferenceTimeout)
 	defer timer.Stop()
 
+	// The terminal include_usage chunk lacks the reasoning breakdown; we hold it
+	// and re-emit it at stream end augmented with the provider's authoritative
+	// reasoning-token count (CompleteCh) — matching the non-streaming/Responses paths.
+	var pendingUsageChunk string
+
 	for {
 		select {
 		case chunk, ok := <-pr.ChunkCh:
@@ -2547,6 +2552,25 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r 
 				// "response.completed" as the terminal event. Adding
 				// extra chunks would break SDK parsers.
 				if !sawResponsesAPI {
+					// Emit the held usage chunk (stream_options.include_usage) with the
+					// reasoning breakdown spliced in from the provider's authoritative
+					// UsageInfo, so streaming reports reasoning_tokens like the rest.
+					if pendingUsageChunk != "" {
+						var usage protocol.UsageInfo
+						select {
+						case u, uok := <-pr.CompleteCh:
+							if uok {
+								usage = u
+							}
+						case <-time.After(2 * time.Second):
+						case <-r.Context().Done():
+						}
+						out := normalizeSSEChunk(pendingUsageChunk)
+						out = injectReasoningIntoStreamUsageChunk(out, usage)
+						out = rewriteChunkModel(out, pr)
+						fmt.Fprintf(w, "%s\n\n", out)
+						flusher.Flush()
+					}
 					// Chat completions format: append SE signature + [DONE].
 					if pr.SESignature != "" {
 						sigEvent, _ := json.Marshal(map[string]any{
@@ -2566,6 +2590,20 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r 
 				if strings.Contains(chunk, `"response.created"`) || strings.Contains(chunk, `"response.output_text.delta"`) {
 					sawResponsesAPI = true
 				}
+			}
+			// Hold the terminal usage chunk (chat completions only) so we can splice
+			// in the reasoning breakdown at stream end; forwarding it inline would
+			// emit it without reasoning_tokens.
+			if !sawResponsesAPI && isUsageOnlyStreamChunk(chunk) {
+				pendingUsageChunk = chunk
+				if !timer.Stop() {
+					select {
+					case <-timer.C:
+					default:
+					}
+				}
+				timer.Reset(inferenceTimeout)
+				continue
 			}
 			if !sawResponsesAPI {
 				chunk = normalizeSSEChunk(chunk)
@@ -3341,6 +3379,47 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
 	details["reasoning_tokens"] = usage.ReasoningTokens
 	usageObj["completion_tokens_details"] = details
 	obj["usage"] = usageObj
+}
+
+// isUsageOnlyStreamChunk reports whether an SSE data chunk is the terminal usage
+// chunk emitted for stream_options.include_usage: empty choices + a non-null usage
+// object, carrying the final usage and no content delta.
+func isUsageOnlyStreamChunk(chunk string) bool {
+	line := strings.TrimPrefix(chunk, "data: ")
+	if !strings.Contains(line, `"usage"`) || strings.Contains(line, `"usage":null`) {
+		return false
+	}
+	var obj map[string]any
+	if err := json.Unmarshal([]byte(line), &obj); err != nil {
+		return false
+	}
+	if u, ok := obj["usage"].(map[string]any); !ok || u == nil {
+		return false
+	}
+	choices, _ := obj["choices"].([]any)
+	return len(choices) == 0
+}
+
+// injectReasoningIntoStreamUsageChunk adds completion_tokens_details.reasoning_tokens
+// (from the provider's authoritative UsageInfo) to a streaming chat-completions
+// usage chunk, so streaming reports the reasoning breakdown like the non-streaming
+// and Responses paths. No-op when there is no reasoning count or the chunk isn't
+// parseable.
+func injectReasoningIntoStreamUsageChunk(chunk string, usage protocol.UsageInfo) string {
+	if usage.ReasoningTokens <= 0 {
+		return chunk
+	}
+	line := strings.TrimPrefix(chunk, "data: ")
+	var obj map[string]any
+	if err := json.Unmarshal([]byte(line), &obj); err != nil {
+		return chunk
+	}
+	injectReasoningDetailIntoRawUsage(obj, usage)
+	b, err := json.Marshal(obj)
+	if err != nil {
+		return chunk
+	}
+	return "data: " + string(b)
 }
 
 func resolveReasoningTokens(usage protocol.UsageInfo, reasoning string) uint64 {
--- a/coordinator/api/consumer_test.go
+++ b/coordinator/api/consumer_test.go
@@ -1665,22 +1665,40 @@ func TestDetectMediaRequirementAnthropicImageBlock(t *testing.T) {
 	}
 }
 
-// TestInjectReasoningIntoStreamUsageChunk covers the helper logic directly.
-func TestInjectReasoningIntoStreamUsageChunk(t *testing.T) {
-	usageChunk := `data: {"object":"chat.completion.chunk","choices":[],"usage":{"prompt_tokens":10,"completion_tokens":50,"total_tokens":60}}`
-	if !isUsageOnlyStreamChunk(usageChunk) {
-		t.Fatal("expected the usage-only chunk to be detected")
+// TestUsageChunkParseAndFinalize covers the parse-once + finalize helpers.
+func TestUsageChunkParseAndFinalize(t *testing.T) {
+	usageChunk := `data: {"object":"chat.completion.chunk","model":"gpt-oss-20b","choices":[],"usage":{"prompt_tokens":10,"completion_tokens":50,"total_tokens":60}}`
+	pr := &registry.PendingRequest{Model: "gpt-oss-20b"}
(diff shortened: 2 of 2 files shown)


coordinator/api/consumer.go
```diff
@@ -2547,6 +2552,25 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r
...
+						case u, uok := <-pr.CompleteCh:
+							if uok {
+								usage = u
+							}
+						case <-time.After(2 * time.Second):
+						case <-r.Context().Done():
+						}
+						out := normalizeSSEChunk(pendingUsageChunk)
+						out = injectReasoningIntoStreamUsageChunk(out, usage)
+						out = rewriteChunkModel(out, pr)
```
ethenotethan: 🟡 **[MEDIUM]** ⚡ Blocking select with 2-second timeout on hot streaming path

💡 _Suggestion:_ Use non-blocking select with default case or reduce timeout to 100-200ms to avoid request latency spikes

📊 _Score: 2×4 = 8 · Category: blocking_io_

coordinator/api/consumer.go
```diff
@@ -3343,6 +3381,47 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
...
+	}
+	var obj map[string]any
+	if err := json.Unmarshal([]byte(line), &obj); err != nil {
+		return false
+	}
+	if u, ok := obj["usage"].(map[string]any); !ok || u == nil {
+		return false
+	}
+	choices, _ := obj["choices"].([]any)
+	return len(choices) == 0
```
ethenotethan: 🟡 **[MEDIUM]** ⚡ JSON parsing performed twice for every usage chunk

💡 _Suggestion:_ Parse JSON once and reuse the parsed object in both isUsageOnlyStreamChunk and injectReasoningIntoStreamUsageChunk

📊 _Score: 3×4 = 12 · Category: repeated_work_

coordinator/api/consumer.go
```diff
@@ -3343,6 +3381,47 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
...
+	var obj map[string]any
+	if err := json.Unmarshal([]byte(line), &obj); err != nil {
+		return chunk
+	}
+	injectReasoningDetailIntoRawUsage(obj, usage)
+	b, err := json.Marshal(obj)
+	if err != nil {
+		return chunk
+	}
+	return "data: " + string(b)
```
ethenotethan: 🟡 **[MEDIUM]** ⚡ Redundant JSON parsing in injectReasoningIntoStreamUsageChunk

💡 _Suggestion:_ Accept pre-parsed map[string]any parameter to avoid re-parsing JSON already parsed by isUsageOnlyStreamChunk

coordinator/api/consumer.go
```diff
@@ -3343,6 +3381,47 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
 	obj["usage"] = usageObj
 }
 
+// isUsageOnlyStreamChunk reports whether an SSE data chunk is the terminal usage
+// chunk emitted for stream_options.include_usage: empty choices + a non-null usage
+// object, carrying the final usage and no content delta.
+func isUsageOnlyStreamChunk(chunk string) bool {
+	line := strings.TrimPrefix(chunk, "data: ")
```
ethenotethan, line 3389: 🟡 **[MEDIUM]** 🏷️ Missing type assertion check for obj["usage"] cast to map[string]any

💡 _Suggestion:_ Add ok check: if u, ok := obj["usage"].(map[string]any); !ok || u == nil { return false }

📊 _Score: 3×4 = 12 · Category: missing_type_assertion_

coordinator/api/consumer.go
```diff
@@ -3343,6 +3381,47 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
...
+		return false
+	}
+	var obj map[string]any
+	if err := json.Unmarshal([]byte(line), &obj); err != nil {
+		return false
+	}
+	if u, ok := obj["usage"].(map[string]any); !ok || u == nil {
+		return false
+	}
+	choices, _ := obj["choices"].([]any)
```
ethenotethan: 🔵 **[INFO]** 🏷️ Missing type assertion check for obj["choices"] cast to []any

💡 _Suggestion:_ Add ok check: if choices, ok := obj["choices"].([]any); !ok { return false }

📊 _Score: 2×3 = 6 · Category: missing_type_assertion_

coordinator/api/consumer.go
```diff
@@ -3343,6 +3381,47 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
...
+// injectReasoningIntoStreamUsageChunk adds completion_tokens_details.reasoning_tokens
+// (from the provider's authoritative UsageInfo) to a streaming chat-completions
+// usage chunk, so streaming reports the reasoning breakdown like the non-streaming
+// and Responses paths. No-op when there is no reasoning count or the chunk isn't
+// parseable.
+func injectReasoningIntoStreamUsageChunk(chunk string, usage protocol.UsageInfo) string {
+	if usage.ReasoningTokens <= 0 {
+		return chunk
+	}
+	line := strings.TrimPrefix(chunk, "data: ")
```
ethenotethan: 🟡 **[MEDIUM]** 🏷️ Function uses map[string]any for JSON parsing without type safety

💡 _Suggestion:_ Consider defining a struct type for the expected JSON structure or add comprehensive type assertions

📊 _Score: 3×4 = 12 · Category: bare_interface_

coordinator/api/consumer.go
```diff
@@ -2547,6 +2552,25 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r
...
+								usage = u
+							}
+						case <-time.After(2 * time.Second):
+						case <-r.Context().Done():
+						}
+						out := normalizeSSEChunk(pendingUsageChunk)
+						out = injectReasoningIntoStreamUsageChunk(out, usage)
+						out = rewriteChunkModel(out, pr)
+						fmt.Fprintf(w, "%s\n\n", out)
+						flusher.Flush()
```
ethenotethan: 🟡 **[MEDIUM]** 🧩 Complex inline usage chunk handling with timeout logic duplicated

💡 _Suggestion:_ Extract this usage chunk handling into a separate method to reduce cognitive load and avoid duplicating timer reset logic

📊 _Score: 3×4 = 12 · Category: over-complexity_

coordinator/api/consumer.go
```diff
@@ -2567,6 +2591,20 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r
...
+			if !sawResponsesAPI && isUsageOnlyStreamChunk(chunk) {
+				pendingUsageChunk = chunk
+				if !timer.Stop() {
+					select {
+					case <-timer.C:
+					default:
+					}
+				}
+				timer.Reset(inferenceTimeout)
+				continue
```
ethenotethan: 🔵 **[INFO]** 🧩 Timer stop/reset pattern duplicated from above

💡 _Suggestion:_ Create a helper method for the timer stop/reset pattern used in both usage chunk handling locations

📊 _Score: 2×3 = 6 · Category: duplicate-logic_

coordinator/api/consumer.go
ethenotethan: 🔵 **[INFO]** 🧩 Manual JSON parsing when structured approach would be cleaner

💡 _Suggestion:_ Consider using a struct type for the chunk format instead of map[string]any manipulation for better type safety

📊 _Score: 2×2 = 4 · Category: reinventing-stdlib_

coordinator/api/consumer.go
ethenotethan: 🔵 **[INFO]** 🧩 JSON unmarshal/marshal pattern repeated from injectReasoningDetailIntoRawUsage

💡 _Suggestion:_ Extract common JSON manipulation pattern into a shared helper function

📊 _Score: 2×3 = 6 · Category: duplicate-logic_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Performance — 3 finding(s) (3 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:2558-2570` — Blocking select with 2-second timeout on hot streaming path
  - _Suggestion:_ Use non-blocking select with default case or reduce timeout to 100-200ms to avoid request latency spikes
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:3388-3400` — JSON parsing performed twice for every usage chunk
  - _Suggestion:_ Parse JSON once and reuse the parsed object in both isUsageOnlyStreamChunk and injectReasoningIntoStreamUsageChunk
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:3408-3422` — Redundant JSON parsing in injectReasoningIntoStreamUsageChunk
  - _Suggestion:_ Accept pre-parsed map[string]any parameter to avoid re-parsing JSON already parsed by isUsageOnlyStreamChunk

### Type_diligence — 3 finding(s) (2 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:3388` — Missing type assertion check for obj["usage"] cast to map[string]any
  - _Suggestion:_ Add ok check: if u, ok := obj["usage"].(map[string]any); !ok || u == nil { return false }
- 🔵 **[INFO]** `coordinator/api/consumer.go:3399` — Missing type assertion check for obj["choices"] cast to []any
  - _Suggestion:_ Add ok check: if choices, ok := obj["choices"].([]any); !ok { return false }
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:3412` — Function uses map[string]any for JSON parsing without type safety
  - _Suggestion:_ Consider defining a struct type for the expected JSON structure or add comprehensive type assertions

- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:2555-2572` — Complex inline usage chunk handling with timeout logic duplicated
  - _Suggestion:_ Extract this usage chunk handling into a separate method to reduce cognitive load and avoid duplicating timer reset logic
- 🔵 **[INFO]** `coordinator/api/consumer.go:2597-2606` — Timer stop/reset pattern duplicated from above
  - _Suggestion:_ Create a helper method for the timer stop/reset pattern used in both usage chunk handling locations
- 🔵 **[INFO]** `coordinator/api/consumer.go:3384-3400` — Manual JSON parsing when structured approach would be cleaner
  - _Suggestion:_ Consider using a struct type for the chunk format instead of map[string]any manipulation for better type safety
- 🔵 **[INFO]** `coordinator/api/consumer.go:3408-3422` — JSON unmarshal/marshal pattern repeated from injectReasoningDetailIntoRawUsage
  - _Suggestion:_ Extract common JSON manipulation pattern into a shared helper function

_10 finding(s) total, 6 blocking. Verdict: REQUEST_CHANGES._

coordinator/api/consumer.go
```diff
@@ -2547,6 +2553,28 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r
...
+						case u, uok := <-pr.CompleteCh:
+							if uok {
+								usage = u
+							}
+						case <-time.After(2 * time.Second):
+						case <-r.Context().Done():
+						}
+						if out := finalizeUsageChunk(pendingUsage, usage, pr); out != "" {
+							fmt.Fprintf(w, "%s\n\n", out)
+							flusher.Flush()
```
ethenotethan, line 2575: 🟡 **[MEDIUM]** ⚡ Synchronous channel read with 2-second timeout blocks streaming response completion

💡 _Suggestion:_ Use non-blocking select with default case or move usage finalization to background goroutine to avoid blocking the response stream

📊 _Score: 3×4 = 12 · Category: Blocking I/O on hot paths_

coordinator/api/consumer.go
```diff
@@ -3343,6 +3381,48 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
...
+	if err := json.Unmarshal([]byte(line), &obj); err != nil {
+		return nil, false
+	}
+	if u, uok := obj["usage"].(map[string]any); !uok || u == nil {
+		return nil, false
+	}
+	if choices, _ := obj["choices"].([]any); len(choices) != 0 {
+		return nil, false
+	}
+	return obj, true
```
ethenotethan, line 3403: 🟡 **[MEDIUM]** ⚡ JSON parsing performed on every chunk to detect usage-only chunks

💡 _Suggestion:_ Use string-based heuristics first (check for 'choices':[] and 'usage':{) before expensive JSON unmarshaling

coordinator/api/consumer.go
```diff
@@ -3343,6 +3381,48 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
...
+		delete(obj, "system_fingerprint")
+	}
+	if pr.PublicModel != "" && pr.PublicModel != pr.Model {
+		obj["model"] = pr.PublicModel
+	}
+	b, err := json.Marshal(obj)
+	if err != nil {
+		return ""
+	}
+	return "data: " + string(b)
```
ethenotethan, line 3423: 🔵 **[INFO]** ⚡ JSON marshaling allocates new byte slice for every usage chunk finalization

💡 _Suggestion:_ Pre-allocate buffer or use streaming JSON encoder to reduce garbage collection pressure

📊 _Score: 2×3 = 6 · Category: Unbounded allocations_

coordinator/api/consumer.go
```diff
@@ -3343,6 +3381,48 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
 	obj["usage"] = usageObj
 }
 
+// parseUsageOnlyStreamChunk decodes a terminal include_usage chunk (empty choices
+// + a non-null usage object, carrying the final usage and no content delta) and
+// returns the parsed object. ok is false for any other chunk. Parsing here once
+// lets the caller hold the object and finalize it at stream end without re-parsing.
+func parseUsageOnlyStreamChunk(chunk string) (obj map[string]any, ok bool) {
```
ethenotethan, line 3388: 🟡 **[MEDIUM]** 🏷️ Missing type assertion check for obj["usage"] cast to map[string]any

💡 _Suggestion:_ Add ok check: if u, uok := obj["usage"].(map[string]any); !uok || u == nil { return nil, false }

📊 _Score: 3×4 = 12 · Category: missing_type_assertion_

coordinator/api/consumer.go
```diff
@@ -3343,6 +3381,48 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
...
+	line := strings.TrimPrefix(chunk, "data: ")
+	// Cheap gate: skip the parse for content deltas and usage:null chunks.
+	if !strings.Contains(line, `"usage"`) || strings.Contains(line, `"usage":null`) {
```
ethenotethan, line 3391: 🔵 **[INFO]** 🏷️ Missing type assertion check for obj["choices"] cast to []any

💡 _Suggestion:_ Add ok check: if choices, cok := obj["choices"].([]any); !cok || len(choices) != 0 { return nil, false }

📊 _Score: 2×3 = 6 · Category: missing_type_assertion_

coordinator/api/consumer.go
```diff
@@ -2547,6 +2553,28 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r
...
+								usage = u
+							}
+						case <-time.After(2 * time.Second):
+						case <-r.Context().Done():
+						}
+						if out := finalizeUsageChunk(pendingUsage, usage, pr); out != "" {
+							fmt.Fprintf(w, "%s\n\n", out)
+							flusher.Flush()
+						}
+					}
```
ethenotethan, line 2577: 🟡 **[MEDIUM]** 🧩 Complex nested select with timeout for reading CompleteCh adds cognitive load

💡 _Suggestion:_ Extract this logic into a helper function like `readUsageWithTimeout(pr *registry.PendingRequest, timeout time.Duration) protocol.UsageInfo` to simplify the main handler flow

📊 _Score: 3×4 = 12 · Category: over-complexity_

coordinator/api/consumer.go
ethenotethan, line 3403: 🔵 **[INFO]** 🧩 parseUsageOnlyStreamChunk duplicates JSON parsing pattern used elsewhere

💡 _Suggestion:_ Consider if this parsing logic could reuse existing JSON unmarshaling helpers or be consolidated with similar parsing code

📊 _Score: 2×3 = 6 · Category: duplicate-logic_

coordinator/api/consumer.go
```diff
@@ -3343,6 +3381,48 @@ func injectReasoningDetailIntoRawUsage(obj map[string]any, usage protocol.UsageI
...
+	}
+	if pr.PublicModel != "" && pr.PublicModel != pr.Model {
+		obj["model"] = pr.PublicModel
+	}
+	b, err := json.Marshal(obj)
+	if err != nil {
+		return ""
+	}
+	return "data: " + string(b)
+}
```
ethenotethan, line 3424: 🔵 **[INFO]** 🧩 finalizeUsageChunk does multiple unrelated operations in one function

💡 _Suggestion:_ Split into separate functions: one for reasoning injection, one for fingerprint cleanup, one for model aliasing - each with a single responsibility

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:2556-2575` — Synchronous channel read with 2-second timeout blocks streaming response completion
  - _Suggestion:_ Use non-blocking select with default case or move usage finalization to background goroutine to avoid blocking the response stream
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:3387-3403` — JSON parsing performed on every chunk to detect usage-only chunks
  - _Suggestion:_ Use string-based heuristics first (check for 'choices':[] and 'usage':{) before expensive JSON unmarshaling
- 🔵 **[INFO]** `coordinator/api/consumer.go:3411-3423` — JSON marshaling allocates new byte slice for every usage chunk finalization
  - _Suggestion:_ Pre-allocate buffer or use streaming JSON encoder to reduce garbage collection pressure

### Type_diligence — 2 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:3388` — Missing type assertion check for obj["usage"] cast to map[string]any
  - _Suggestion:_ Add ok check: if u, uok := obj["usage"].(map[string]any); !uok || u == nil { return nil, false }
- 🔵 **[INFO]** `coordinator/api/consumer.go:3391` — Missing type assertion check for obj["choices"] cast to []any
  - _Suggestion:_ Add ok check: if choices, cok := obj["choices"].([]any); !cok || len(choices) != 0 { return nil, false }

### Additive_complexity — 3 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:2556-2577` — Complex nested select with timeout for reading CompleteCh adds cognitive load
  - _Suggestion:_ Extract this logic into a helper function like `readUsageWithTimeout(pr *registry.PendingRequest, timeout time.Duration) protocol.UsageInfo` to simplify the main handler flow
- 🔵 **[INFO]** `coordinator/api/consumer.go:3384-3403` — parseUsageOnlyStreamChunk duplicates JSON parsing pattern used elsewhere
  - _Suggestion:_ Consider if this parsing logic could reuse existing JSON unmarshaling helpers or be consolidated with similar parsing code
- 🔵 **[INFO]** `coordinator/api/consumer.go:3411-3424` — finalizeUsageChunk does multiple unrelated operations in one function
  - _Suggestion:_ Split into separate functions: one for reasoning injection, one for fingerprint cleanup, one for model aliasing - each with a single responsibility

_8 finding(s) total, 4 blocking. Verdict: REQUEST_CHANGES._

coordinator/api/consumer.go
Gajesh2007: Kept on purpose, but it is no longer on the hot path. After b281e71e the terminal usage chunk is held as a parsed object and the CompleteCh value is almost always already present by stream end, so the select returns immediately. The 2s bound only matters if a provider never sends `inferenceComplete` — it stops that from hanging the response forever. Dropping it reintroduces an unbounded wait.

coordinator/api/consumer.go
Gajesh2007: Fixed in b281e71e — `parseUsageOnlyStreamChunk` unmarshals once and returns the `map[string]any`, which we hold and hand straight to `finalizeUsageChunk`. No detect-then-reparse anymore.

coordinator/api/consumer.go
Gajesh2007: Fixed in b281e71e — injection now runs on the already-parsed object inside `finalizeUsageChunk`, and we marshal exactly once at the end. No separate parse in the inject path.

coordinator/api/consumer.go
Gajesh2007: Already covered — `parseUsageOnlyStreamChunk` does exactly your suggested check at consumer.go:3397: `if u, uok := obj["usage"].(map[string]any); !uok || u == nil { return nil, false }`. The line you flagged was the pre-refactor string-scan path, which no longer exists after b281e71e.

coordinator/api/consumer.go
Gajesh2007: This one is intentional, not a missing check. Detection uses `choices, _ := obj["choices"].([]any); len(choices) != 0` (consumer.go:3400): a failed assertion yields a nil slice → len 0 → correctly treated as a usage-only chunk. The suggested `if _, ok := ...; !ok { return false }` would reject a terminal usage chunk that omits `choices` entirely (a valid shape under `stream_options.include_usage`), so we would drop the usage. Keeping the len-based check.

coordinator/api/consumer.go
Gajesh2007: By design here. This is a passthrough finalizer for the provider terminal usage chunk — it must preserve unknown/forward-compat fields verbatim and only splice reasoning + rewrite the public model alias. A typed struct would silently drop any field we do not model. `map[string]any` is the established convention on this passthrough path (same as the Responses handler).

coordinator/api/consumer.go
Gajesh2007: Addressed in b281e71e — the inline block is extracted into `parseUsageOnlyStreamChunk` + `finalizeUsageChunk`, and the handler just holds the parsed object and finalizes once at stream end. Added unit coverage in `TestUsageChunkParseAndFinalize`.

coordinator/api/consumer.go
Gajesh2007: Fixed in b281e71e — removed the duplicated stop/reset in the capture branch. The usage chunk is the last thing before stream close, and the already-running timer still bounds a post-usage hang, so the extra reset was dead code.

coordinator/api/consumer.go
Gajesh2007: Same as the type-safety thread — intentional. This path forwards the provider usage object and must not drop forward-compat fields, so we parse into `map[string]any` rather than a fixed struct. It now parses exactly once (b281e71e).

coordinator/api/consumer.go
Gajesh2007: Fixed in b281e71e — consolidated to a single unmarshal (`parseUsageOnlyStreamChunk`) and a single marshal (`finalizeUsageChunk`). No repeated round-trip.

coordinator/api/consumer.go
```diff
@@ -2567,6 +2595,16 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r
...
 				}
 			}
+			// Hold the terminal usage chunk (chat completions only) so we can splice
+			// in the reasoning breakdown at stream end; forwarding it inline would
+			// emit it without reasoning_tokens. The stream closes right after, so the
+			// running timer still guards a post-usage hang.
+			if !sawResponsesAPI {
+				if obj, isUsage := parseUsageOnlyStreamChunk(chunk); isUsage {
+					pendingUsage = obj
+					continue
```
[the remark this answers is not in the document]
Gajesh2007: Valid — fixed in follow-up #305. The dedup in this PR dropped the timer reset on the usage-hold `continue` (the two original copies guarded different paths). #305 moves a single `Stop()`/`Reset()` to the top of the chunk-received branch so both the forward and hold paths re-arm the idle window. Harmless in practice (`inferenceTimeout = 600s`), but the fix is correct.

Gajesh2007:
Both Codex findings addressed in follow-up #305:

1. **Usage-only first chunk** (the L2503-2511 note) — a zero-delta completion makes the `include_usage` frame the very first chunk; #305 routes `firstChunk` through `parseUsageOnlyStreamChunk` so it is held + reasoning-spliced instead of emitted raw. Regression test `TestStreamingChatUsageOnlyFirstChunk` added. (Practical impact ~nil since no reasoning text is streamed in that case, but it closes the gap.)
2. **Timer reset on the hold path** (L2605) — fixed as noted inline.

go test ./api/ green.

PR #317, opened 2026-06-11
security: cap inference request bodies, close earnings IDOR, clamp negative billing tokens

anupsv:
Coordinator request/billing-path security hardening, split out from a larger working tree (PR 2 of 2).

## SEC-006 — unbounded request bodies
`handleChatCompletions` and `handleGenericInference` now wrap the body in a 64 MiB `http.MaxBytesReader` (→ 413 on overflow); the HTTP server gets `ReadHeaderTimeout` + `MaxHeaderBytes`. The cap is sized to admit the console's vision payloads (4 × 10 MB images ≈ 53 MiB base64).

## H5 — unauthenticated earnings IDOR
`GET /v1/provider/earnings` and `/v1/provider/node-earnings` were unauthenticated and returned any party's financial data by query param. Now both require auth and scope to the authenticated account: `handleProviderEarnings` ignores the client `?wallet=` param; `handleNodeEarnings` filters detail rows to the caller and uses a new account-scoped lifetime summary (`GetProviderEarningsSummaryForAccount`) so a re-linked, multi-account `provider_key` cannot leak another account's aggregate earnings.

## H6 — negative-token balance minting
Provider-reported usage tokens are clamped `>=0` at ingestion, cost is clamped `>=0` in pricing, settlement refund is capped `<= reserved`, and `Debit`/`Credit` reject negative amounts in both stores (`ErrNegativeAmount`). Closes a path where negative `completion_tokens` from a serving provider could over-refund (mint) a service account's balance.

## Tests
- `body_limit_test.go` (413), `earnings_authz_test.go` (anon 401, cross-account 403, multi-account no-leak, owner 200), `negative_usage_billing_test.go` (no-mint), `store/negative_amount_test.go` (negative Debit/Credit rejected)
- All revert-verified (fail without the fix); both independent reviewers (correctness + Go security) passed.
- `go build` / `vet` / `gofmt` clean; `api` + `payments` + `store` suites green.

---
[...]

PR #319, opened 2026-06-11
security: close model-substitution fail-open (H3) + tighten code-attest reuse staleness (H2)

anupsv:
Stacked on #317 (base = `security/coordinator-body-earnings-billing`). Closes two attestation/routing hardening items from the security review.

## H3 — model-substitution fail-open (registry)

`registry.modelAllowedByCatalogLocked` previously fail-open'd on an **empty reported weight hash**:

```go
return entry.WeightHash == "" || model.WeightHash == "" || model.WeightHash == entry.WeightHash
```

A provider could bypass a *pinned* catalog entry by reporting no hash at all. The fix removes the `model.WeightHash == ""` clause: an empty reported hash no longer bypasses a pinned entry.

**Why this is safe (verified against `provider-swift/Sources`):** an honest Swift provider *eagerly* computes a per-model weight hash from disk at startup for every model it can serve (and refreshes it on load). An empty hash therefore means the model isn't on disk (can't be served anyway) or the provider is dodging swap detection. Behavior table — only the empty-on-pinned case changes:

| catalog entry | reported | before | after |
|---|---|---|---|
| unpinned (`""`) | anything | allow | allow |
| pinned `H` | `""` | **allow (fail-open)** | **reject** |
| pinned `H` | `H` | allow | allow |
| pinned `H` | `X` | reject | reject |

**Availability:** a `false` here is a routing *deroute*, not an untrust — fully recoverable the instant the provider reports a real hash; requests queue (120s) rather than hard-fail. The **mismatch** clause was already enforced unflagged, and a *separate* untrust path already fires on a non-empty mismatch — so a healthy fleet already reports correct non-empty hashes. No flag/grace needed.

**Scope honesty:** this is a *self-reported, defense-in-depth* layer. A determined attacker can report the publicly-known expected hash while serving different weights; the real cryptographic proof of the loaded model is the Secure-Enclave-signed attestation (`activeModelHash`), not this plaintext field. The code comment says exactly this.

## H2 — code-attestation reuse staleness (throttle)

- Pin `reuseWindow` 30m → 20m (== `pushCooldown`). With `reuseWindow > pushCooldown` there was a band where a reconnect could neither reuse nor push (cooldown not elapsed) — derouted until cooldown. Pinning them equal closes that gap and tightens staleness.
- Correct the doc/comments that falsely implied the self-reported `version` proves the binary is unchanged. Reuse only sets the `CodeAttested` bit; routing independently requires this connection's own live SE challenge (`ChallengeVerifiedSIP`), so a stale reuse never routes a connection on its own. This is hardening + comment-correctness, not a critical-bypass fix.

No new APNs pushes: spacing is governed solely by the unchanged `pushCooldown` (≤3/hr/device); a shorter `reuseWindow` only causes more *reuse misses*, each still rate-limited.

## Tests (both revert-verified — fail when the fix is reverted)

- `registry/catalog_hash_test.go`: empty-on-pinned rejected, match allowed, mismatch rejected, unpinned allows empty, not-in-catalog denied.
- `api/code_attest_throttle_test.go`: reuse honored in-window; denied on version mismatch / different device / past window; plus the `reuseWindow <= pushCooldown` invariant.
- Fixtures `registerBuildsProvider` / `connectAndPrepareProvider` now report the seeded weight hash (`testHash`) — modelling honest provider behavior — so the 3 alias/list-models tests stay green under the stricter gate (all 14 call sites enumerated; unpinned-catalog callers unaffected).

`go build ./...`, `go test ./api/ ./registry/`, `gofmt`, `go vet` all green. Reviewed by an independent Go security auditor and a general-purpose reviewer (both PASS).

### Non-blocking follow-up (pre-existing, not introduced here)
`modelAllowedByCatalogLocked` compares hashes with exact `==` while `MergeProviderModels` uses `strings.EqualFold`. Harmless today (honest sources produce identical casing); worth unifying later.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/319"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783790866&installation_id=133247490&pr_number=319&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F319&signature=cb570a31921d40462ecdd0cca5af142bd4578dbef3a1114eca0e2182d0a1db7a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


coordinator/api/code_attest_throttle_test.go
```diff
@@ -0,0 +1,48 @@
...
+		t.Error("reuse must expire after the window")
+	}
+
+	// The tightening invariant: reuse staleness never exceeds one push cooldown,
+	// so a reconnect that can no longer reuse can always push a fresh challenge.
+	if th.reuseWindow > th.pushCooldown {
+		t.Errorf("reuseWindow (%v) must be <= pushCooldown (%v) to bound staleness without a deroute gap",
+			th.reuseWindow, th.pushCooldown)
+	}
+}
```
ethenotethan: 🔵 **[INFO]** 🧩 New test file for single invariant could be inline test case

💡 _Suggestion:_ Consider adding this as a test case in existing code_attest_throttle tests rather than a separate file for one invariant check

coordinator/registry/catalog_hash_test.go
```diff
@@ -0,0 +1,46 @@
...
+		{"mismatched hash on a pinned model is rejected", protocol.ModelInfo{ID: pinned, WeightHash: "WRONG"}, false},
+		{"unpinned model allows any (incl. empty) hash", protocol.ModelInfo{ID: unpinned, WeightHash: ""}, true},
+		{"model not in catalog is denied", protocol.ModelInfo{ID: "not-in-catalog", WeightHash: "x"}, false},
+	}
+	for _, tc := range cases {
+		if got := reg.modelAllowedByCatalogLocked(tc.model); got != tc.want {
+			t.Errorf("%s: allowed=%v, want %v", tc.name, got, tc.want)
+		}
+	}
+}
```
ethenotethan, line 44: 🔵 **[INFO]** 🧩 New test file for single function could be inline test case

💡 _Suggestion:_ Consider adding this as a test case in existing registry tests rather than a separate file for one function test

ethenotethan (commented):

**Verdict:** COMMENT

### Performance — ✅ No issues found

### Additive_complexity — 2 finding(s)
- 🔵 **[INFO]** `coordinator/api/code_attest_throttle_test.go:1-48` — New test file for single invariant could be inline test case
  - _Suggestion:_ Consider adding this as a test case in existing code_attest_throttle tests rather than a separate file for one invariant check
- 🔵 **[INFO]** `coordinator/registry/catalog_hash_test.go:1-46` — New test file for single function could be inline test case
  - _Suggestion:_ Consider adding this as a test case in existing registry tests rather than a separate file for one function test

_2 finding(s) total, 0 blocking. Verdict: COMMENT._

coordinator/api/code_attest_throttle.go
```diff
@@ -47,7 +50,7 @@ func newCodeAttestThrottle() *codeAttestThrottle {
 	return &codeAttestThrottle{
 		attested:     make(map[string]codeAttestRecord),
 		lastPush:     make(map[string]time.Time),
-		reuseWindow:  30 * time.Minute,
+		reuseWindow:  20 * time.Minute, // pinned to pushCooldown — bound reuse staleness, no deroute gap
```
ethenotethan: 🟡 **[MEDIUM]** 🔒 Reduced reuse window from 30m to 20m may increase attestation staleness risk

💡 _Suggestion:_ Consider if 20 minute window provides sufficient security margin for attestation freshness

📊 _Score: 3×4 = 12 · Category: timing-side-channel_

ethenotethan: 🔵 **[INFO]** 🧩 Magic number duplication: reuseWindow hardcoded to match pushCooldown

💡 _Suggestion:_ Define reuseWindow as pushCooldown instead of duplicating the 20-minute value

coordinator/registry/catalog_hash_test.go
ethenotethan, line 44: 🔵 **[INFO]** 🧩 Test duplicates existing registry test patterns without shared helpers

💡 _Suggestion:_ Extract common test setup (registry creation, catalog setting) into shared helper functions

📊 _Score: 1×2 = 2 · Category: duplicate logic_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 1 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/code_attest_throttle.go:53` — Reduced reuse window from 30m to 20m may increase attestation staleness risk
  - _Suggestion:_ Consider if 20 minute window provides sufficient security margin for attestation freshness

### Performance — ✅ No issues found

### Additive_complexity — 2 finding(s)
- 🔵 **[INFO]** `coordinator/api/code_attest_throttle.go:53` — Magic number duplication: reuseWindow hardcoded to match pushCooldown
  - _Suggestion:_ Define reuseWindow as pushCooldown instead of duplicating the 20-minute value
- 🔵 **[INFO]** `coordinator/registry/catalog_hash_test.go:1-46` — Test duplicates existing registry test patterns without shared helpers
  - _Suggestion:_ Extract common test setup (registry creation, catalog setting) into shared helper functions

coordinator/api/code_attest_throttle_test.go
anupsv: Good catch — done in 514f9f95. There was already an established home for these unit tests: `TestCodeAttestThrottleBudgetAndReuse` in `code_attestation_integration_test.go` drives `newCodeAttestThrottle()` directly and already covers version-gating + window-expiry + the push budget. I folded the only genuinely-new assertions — the `reuseWindow <= pushCooldown` invariant (the actual H2 regression guard) and SE-key-scoped reuse — into that test and deleted the standalone file.

coordinator/registry/catalog_hash_test.go
anupsv: Keeping this as a focused file. The registry package's convention is already granular themed hash-test files — `model_hash_race_test.go` (race semantics) and `model_hash_refresh_test.go` (refresh/heal) are two separate files, both about per-model weight hashes. `catalog_hash_test.go` (the `modelAllowedByCatalogLocked` filter logic) is a third granular file in that same cluster, which matches both the existing layout and the repo's documented preference for small single-responsibility files. No existing test unit-tests this function, so there's no sibling to merge into (unlike the throttle case above).

coordinator/registry/catalog_hash_test.go
```diff
@@ -0,0 +1,46 @@
+package registry
+
+// Regression test for H3 (SEC-007 model-substitution fail-open): an empty
```
Gajesh2007: can we clean up and compress these comments

Gajesh2007 (approved):

coordinator/registry/catalog_hash_test.go
anupsv: Done in feb8ff11. Dropped the internal tracker IDs (H2/H3/SEC-007) — they aren't referenceable from the public repo — and rewrote each comment to plainly say what the code does and why, compressing the verbose blocks. Swept all of PR #319's files to confirm no `H#`/`SEC-#`/`DAR-#` references remain.

coordinator/api/code_attest_throttle.go
ethenotethan: 🟡 **[MEDIUM]** 🔒 Reduced reuse window from 30m to 20m may increase attestation staleness risk

💡 _Suggestion:_ Consider if 20 minutes provides sufficient security margin for attestation freshness in high-churn scenarios

📊 _Score: 3×4 = 12 · Category: timing-side-channel_

ethenotethan: 🔵 **[INFO]** 🧩 Magic number duplication: reuseWindow hardcoded to match pushCooldown

💡 _Suggestion:_ Define reuseWindow as pushCooldown directly or use a constant to avoid drift

coordinator/registry/catalog_hash_test.go
```diff
@@ -0,0 +1,44 @@
...
+		{"mismatched hash on a pinned model is rejected", protocol.ModelInfo{ID: pinned, WeightHash: "WRONG"}, false},
+		{"unpinned model allows any (incl. empty) hash", protocol.ModelInfo{ID: unpinned, WeightHash: ""}, true},
+		{"model not in catalog is denied", protocol.ModelInfo{ID: "not-in-catalog", WeightHash: "x"}, false},
+	}
+	for _, tc := range cases {
+		if got := reg.modelAllowedByCatalogLocked(tc.model); got != tc.want {
+			t.Errorf("%s: allowed=%v, want %v", tc.name, got, tc.want)
+		}
+	}
+}
```
ethenotethan, line 44: 🔵 **[INFO]** 🧩 Test file recreates registry setup pattern seen in other test files

💡 _Suggestion:_ Consider extracting common test registry setup to a shared helper function

📊 _Score: 1×2 = 2 · Category: duplicate logic_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 1 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/code_attest_throttle.go:53` — Reduced reuse window from 30m to 20m may increase attestation staleness risk
  - _Suggestion:_ Consider if 20 minutes provides sufficient security margin for attestation freshness in high-churn scenarios

### Performance — ✅ No issues found

### Additive_complexity — 2 finding(s)
- 🔵 **[INFO]** `coordinator/api/code_attest_throttle.go:53` — Magic number duplication: reuseWindow hardcoded to match pushCooldown
  - _Suggestion:_ Define reuseWindow as pushCooldown directly or use a constant to avoid drift
- 🔵 **[INFO]** `coordinator/registry/catalog_hash_test.go:1-44` — Test file recreates registry setup pattern seen in other test files
  - _Suggestion:_ Consider extracting common test registry setup to a shared helper function

coordinator/api/code_attest_throttle.go
anupsv: The direction is inverted: 30m→20m *reduces* staleness (attestations are re-verified sooner), it doesn't increase it. 20m isn't arbitrary — it's the **floor**, pinned to `pushCooldown`. Going shorter would open a deroute gap: a reconnect's reuse would expire before a fresh push is allowed (`now-lastAttested ≥ reuseWindow` while `now-lastPush < pushCooldown`), leaving the device unable to attest until cooldown elapses. Going longer just adds staleness. So `reuseWindow == pushCooldown` is the minimum-staleness point with no availability gap.

High-churn is handled by reuse itself: between pushes (capped at ≤3/hr/device by `pushCooldown`) reconnects reuse the recent attestation, so churn never exhausts the APNs budget. And code-attest reuse only sets the `CodeAttested` bit — routing independently requires this connection's own live SE challenge (`ChallengeVerifiedSIP`), so a stale reuse never routes a connection on its own. b9dbb504 makes the floor structural (both derived from one const).

anupsv: Done in b9dbb504 — both `reuseWindow` and `pushCooldown` now derive from a single `codeAttestPushCooldown` const, so the "pinned to pushCooldown" relationship is structural and can't drift. Good call.

coordinator/registry/catalog_hash_test.go
anupsv: The setup here is `New(testLogger())` + a test-specific `SetModelCatalog(...)`. `testLogger()` is already the package's shared helper, and the catalog is intentionally per-test (each case pins a different catalog), so there's no shared state left to extract — a helper would just wrap one standard constructor call. (Same focused-file rationale as the thread above, which was approved.)

ethenotethan (commented):

**Verdict:** COMMENT

### Performance — ✅ No issues found

### Additive_complexity — 3 finding(s)
- 🔵 **[INFO]** `coordinator/api/code_attest_throttle.go:49-55` — Const definition with verbose comment duplicates logic already in constructor
  - _Suggestion:_ Move the comment to the constructor where both values are actually set, or simplify to just document the rate limit purpose
- 🔵 **[INFO]** `coordinator/api/code_attest_throttle.go:57-64` — Constructor comment repeats const comment with slight variations
  - _Suggestion:_ Consolidate the documentation - either in the const or constructor, not both with overlapping explanations
- 🔵 **[INFO]** `coordinator/registry/catalog_hash_test.go:3-6` — Multi-line comment in unusual location explains test purpose redundantly
  - _Suggestion:_ Move explanation to test function docstring or simplify to standard single-line comment

_3 finding(s) total, 0 blocking. Verdict: COMMENT._

coordinator/api/code_attest_throttle.go
```diff
@@ -43,12 +46,20 @@ type codeAttestRecord struct {
 	version string
 }
 
+// codeAttestPushCooldown bounds APNs pushes to one device (<= 3/hour, within
+// Apple's background-push budget) and is reused verbatim as the reuse-window
+// length. Pinning the two equal is the safe floor: a shorter reuse window would
+// open a deroute gap (reuse expires before a fresh push is allowed), while a
+// longer one only adds attestation staleness. Deriving both from this single
+// const keeps them from drifting apart.
+const codeAttestPushCooldown = 20 * time.Minute
```
ethenotethan, line 55: 🔵 **[INFO]** 🧩 Const definition with verbose comment duplicates logic already in constructor

💡 _Suggestion:_ Move the comment to the constructor where both values are actually set, or simplify to just document the rate limit purpose

coordinator/api/code_attest_throttle.go
```diff
@@ -43,12 +46,20 @@ type codeAttestRecord struct {
...
 func newCodeAttestThrottle() *codeAttestThrottle {
 	return &codeAttestThrottle{
 		attested:     make(map[string]codeAttestRecord),
 		lastPush:     make(map[string]time.Time),
-		reuseWindow:  30 * time.Minute,
-		pushCooldown: 20 * time.Minute, // <= 3 pushes/hour/device (APNs background budget)
+		reuseWindow:  codeAttestPushCooldown, // pinned to pushCooldown — no deroute gap, bounded staleness
+		pushCooldown: codeAttestPushCooldown,
 		maxAttempts:  3,
 		now:          time.Now,
```
ethenotethan, line 64: 🔵 **[INFO]** 🧩 Constructor comment repeats const comment with slight variations

💡 _Suggestion:_ Consolidate the documentation - either in the const or constructor, not both with overlapping explanations

📊 _Score: 1×2 = 2 · Category: duplicate logic_

coordinator/registry/catalog_hash_test.go
```diff
@@ -0,0 +1,44 @@
+package registry
+
+// Regression test for the catalog weight-hash filter: an empty provider-reported
+// weight hash must not bypass a pinned catalog entry. An honest provider hashes
+// every servable model from disk, so an empty hash is never a valid "cold model"
+// signal — it's a missing model or a swap-detection dodge. Fails without the fix
```
ethenotethan, line 6: 🔵 **[INFO]** 🧩 Multi-line comment in unusual location explains test purpose redundantly

💡 _Suggestion:_ Move explanation to test function docstring or simplify to standard single-line comment

📊 _Score: 1×2 = 2 · Category: over-abstraction_

PR #321, opened 2026-06-11
fix(coordinator): unblock the gemma 4bit takeover migration — challenge alibi, load-fail cooldown, rollback pre-positioning

Gajesh2007:
Pre-flight fixes for today's `gemma-4-26b` → `gemma-4-26b-qat-4bit` alias-takeover migration, from two independent code audits of the shipped hot-swap machinery. **The migration must not be flipped until this is deployed.**

## Why these are blockers

1. **Challenge alibi (the big one).** After a hard-swap, the retired build stays GPU-resident — and remains the provider's *active* model — after leaving the advertised set. The bare `active_model_hash` membership check only consulted **advertised** models, so every swapped provider would be **hard-untrusted (all models, non-recoverable until process reconnect)** at its next 5-minute challenge. Fleet-wide self-deroute, mid-migration. The check now also accepts an active hash matching a **per-model-validated** reported hash (a catalog-registered build the provider legitimately has loaded). Reviewers confirmed **no net security weakening**: a tampered hash still matches nothing and still untrusts; the only new-pass state is the legitimate retired-resident case. Regression tests in both directions.
2. **Load-failure cool-down for generic failures.** A build that disk-verifies but can't GPU-load (metallib/kernel) previously kept attracting 100% of alias traffic as repeated 500s with no cool-down and no alias fallback. `"model load failed"` now cools the provider-model pair, making the desired build unroutable so resolution falls back to `previous`. (Canary remains the primary safety; this is the backstop.)
3. **Takeover rollback pre-positioning.** A takeover alias can never flip back to its own name (`desired_build == alias_id` is rejected) — the documented revert was structurally impossible. `scripts/preposition-rollback-build.sh` server-side-copies the old weights to a distinct id (R2 copy + manifest rewrite + register + promote) so revert becomes a normal alias flip. Reviewer-found schema bugs in the register step fixed (explicit registry args; `/v1/models` doesn't carry registerable fields).
4. **Runbook corrections**: same-name takeover is a single atomic POST (the documented two-step 400s); revert + Step-7 retire sequences for takeover aliases (deprecate-first, gated on residency drain to protect the alibi); monitoring keys incl. `/v1/models/capacity` name-decay.

## Tests
`TestChallengeRetiredResidentBuildHashDoesNotUntrust`, `TestChallengeRetiredAlibiDoesNotWeakenTamperCheck`, `TestIsModelLoadFailure`; full api+registry suites green. Two-reviewer gate: independent review PASS (script schema findings fixed in ce9bd71d); codex review in flight.

## Deploy order (human)
Merge → deploy coordinator → promote qat → run pre-positioning script → canary the build on one owned machine → single takeover POST at low traffic.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/321"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783800188&installation_id=133247490&pr_number=321&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F321&signature=31581b799ca3bded40c9b2da5b88fd4c589e64070564f0c59b209bc5cbfcf482"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/api/model_hash_race_test.go
+++ b/coordinator/api/model_hash_race_test.go
@@ -65,6 +65,23 @@ func challengeExchangeWithCatalog(
 	activeModelHash string,
 ) registry.ProviderStatus {
 	t.Helper()
+	return challengeExchangeAdvertising(t, catalog, []protocol.ModelInfo{
+		{ID: "model-gemma", SizeBytes: 1000, ModelType: "chat", Quantization: "8bit", WeightHash: gemmaHash},
+		{ID: "model-gptoss", SizeBytes: 1000, ModelType: "chat", Quantization: "4bit", WeightHash: gptOSSHash},
+	}, modelHashes, activeModelHash)
+}
+
+// challengeExchangeAdvertising additionally takes the ADVERTISED model set, so
+// tests can shape the post-hot-swap state where a still-resident build reports
+// a hash without being advertised any more.
+func challengeExchangeAdvertising(
+	t *testing.T,
+	catalog []registry.CatalogEntry,
+	advertised []protocol.ModelInfo,
+	modelHashes map[string]string,
+	activeModelHash string,
+) registry.ProviderStatus {
+	t.Helper()
 	logger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelError}))
 	st := store.NewMemory(store.Config{AdminKey: "test-key"})
 	reg := registry.New(logger)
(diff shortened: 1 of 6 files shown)


Gajesh2007:
## Security review of the attestation alibi — NO NET WEAKENING (two independent traces)

The one change that weakens a check on paper — accepting an `active_model_hash` that matches a per-model-validated `ModelHashes` entry — was reviewed adversarially under a fully-malicious-provider threat model. Verdict: **no real integrity guarantee is reduced.**

Why: the coordinator **never recomputes a provider's served-weight hash** (no inference-path or challenge-path consumer hashes GPU bytes). The check only ever compared SE-signed *self-reported* strings against the public catalog constant — a consistency check, not a served-bytes==catalog-bytes proof. A malicious enrolled provider could **already** pass the old check while serving tampered weights, simply by reporting an advertised model's (publicly-readable) catalog hash as active. The change widens the accepted set only to "active == a real catalog hash of a registered build the provider names in `ModelHashes`" — values already supplyable via the advertised path. A forged hash matches neither loop and still hard-untrusts (`TestChallengeRetiredAlibiDoesNotWeakenTamperCheck`). The new-trusted inputs are exactly the legitimate retired-but-resident registered build.

Residual caveat (unchanged by this PR, pre-existing): the whole model-swap check is advisory against a *malicious* provider — served-content identity is self-reported, integrity rests on admission/vetting + APNs code-identity, not crypto. This commit does not touch that boundary.

Reviews: independent full-diff review PASS (rollback-script register-schema bug + Step-7 takeover-retire gap found → fixed in ce9bd71d); focused security tie-breaker PASS (above). Codex bot wedged on infra (non-gating).

PR #374, opened 2026-06-16
fix(provider): bound Gemma-4 checkpoint-capture to stop the Metal live-resource (499000) leak

Gajesh2007:
## Summary

Gemma-4 (`gemma-4-26b-qat-4bit`) crash-loops on busy providers with
`[metal::malloc] Resource limit (499000) exceeded` (≈ every 15 min on an M3
Ultra with free RAM). This is a **live Metal buffer-COUNT leak**, not a
bytes/cache problem. This PR bounds the checkpoint KV-capture pipeline so the
live resource count plateaus instead of climbing to the ceiling.

Provider-only change. The `libs/mlx-swift*` submodules are untouched.

Target: **0.6.12**.

## Root cause (verified)

Gemma-4 is heterogeneous (sliding `[8,256]` + full `[2,512]`), so it uses the
`.checkpoint` prefix-cache tier, which is **on by default** (opt out with
`DARKBLOOM_PREFIX_CACHE=0`):

- `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:1029-1066` — `.checkpoint` tier setup
- `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:1185-1188` — prefix cache default-ON / opt-out

The capture hook, fired per checkpoint boundary, spawned an **unbounded**
detached Task:

- `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:1115-1124` (production) and `:821-825` (test seam)

```swift
scheduler.onCheckpointCapture = { prefixTokens, length, caches in
    let box = SendableKVCaches(caches)
    Task { await mgr.store(tokens: prefixTokens, checkpointLength: length, caches: box) }  // <-- UNBOUNDED
}
```

The hook is fired synchronously on the engine queue with a freshly-extracted,
per-layer KV snapshot:

- `libs/mlx-swift-lm/Libraries/MLXLMCommon/ContinuousBatching/Scheduler.swift:396-401` — `caches = pp.ppBatch.promptCache.map { $0.extractBatched(0) }`

Each spawned `Task` **pins a full per-layer KV snapshot (live `MLXArray`s,
≈ layers × 2 Metal buffers) until `mgr.store` runs**. `PrefixCacheManager` is a
single actor that serializes `store` behind AES-GCM serialize + fsync and
against every `lookup`/`materialize`:

- `provider-swift/Sources/ProviderCore/KVCache/PrefixCacheManager.swift:752` — `func store(...)`

Under sustained traffic the actor falls behind, the capture Tasks queue, their
live snapshots accumulate, and the **live Metal resource COUNT** (independent of
bytes — see `Memory.numResources` /  `Memory.resourceLimit` in
`libs/mlx-swift/Source/MLX/Memory.swift:199-222`) climbs to the
`iogpu.rsrc_limit` ceiling (~499000) → crash.

## Why the prior mlx/mlx-c "count-aware" fixes can't fix this

Those fixes trim the **cache** (recycled, freed buffers) by making the trim
count-aware. But the snapshots held by the queued capture Tasks are **LIVE**,
not cached — they are still referenced, so the allocator cannot reclaim them no
matter how aggressively the cache is trimmed. The idle-gated
`Memory.clearCache()` also never fires under sustained load. The only fix is to
**bound the number of live snapshots retained in flight**, which is what this PR
does.

## The fix (provider-only)

1. **Backpressure / bounded capture pipeline (primary).** New
   `CheckpointCapturePipeline` (`BatchScheduler.swift` → new
   `CheckpointCapturePipeline.swift`): a bounded, single-consumer `AsyncStream`
   (`bufferingNewest(max-in-flight)`) feeding one consumer Task that calls
   `mgr.store`. `submit` is synchronous/non-blocking and **drops the surplus
   snapshot on full** (released immediately, so its Metal buffers free). This
   caps retained live snapshots to a small constant — `max-in-flight + 1`
   (buffer + the one in the consumer) — regardless of how far behind the
   manager actor falls. Default `max-in-flight = 2`, env-tunable via
   `DARKBLOOM_KV_CAPTURE_MAX_INFLIGHT`. Dropping is safe: capture is best-effort
   cache-warming, so a drop only costs a future cache miss, never correctness.

2. **Admission gate under resource pressure (secondary, cheap).** Before
   enqueuing, skip capture when the live Metal resource count is high —
   `Memory.numResources > 0.7 × Memory.resourceLimit` (guards
   `resourceLimit == 0`). This stops feeding the leak exactly when it matters,
   and matches the engine's existing `[rsrc]` 70%-pressure threshold.

3. **Teardown drains/cancels.** `stopCurrentEngine` (and the superseded-load
   bailouts) finish the stream and cancel the consumer, releasing any retained
   snapshots so nothing leaks across model swaps. Ordered before the manager
   flush so an in-flight `store` can't race the index flush / accountant
   deregister.

The decode path, KV-cache classes, and the manager's persistence logic are
unchanged; the only behavior change is bounding capture. Both capture-hook
sites (production builder + test seam) share one `makeCheckpointCaptureWiring`
helper.

## Tests

`swift build` passes; new + existing suites are green against `master`
(`v0.6.11`) submodule pins:

- **New** `CheckpointCapturePipelineTests` (4 tests):
  - `capturePipelineBoundsInFlightSnapshots` — floods the pipeline against a
    deliberately-stalled consumer and asserts the number of retained snapshots
    never exceeds the small constant bound (vs all N if unbounded), that
    overflow is **dropped not queued**, and that nothing leaks after drain.
  - `capturePipelineCapacityFloorIsOne` — capacity clamps to ≥ 1.
  - `captureAdmissionGateThreshold` — 70% threshold math incl. the
    `resourceLimit == 0` guard.
  - `captureMaxInFlightIsSmallAndPositive` — default cap + env knob.
- Regression suites: `BatchScheduler`, `PrefixCache`, `Continuous`, `Hybrid`,
  `KVCache` — **226 tests pass** (live model-download tests skip without
  `DARKBLOOM_LIVE_MLX_TESTS=1`).

## Validating in prod

Run a provider on Gemma-4 with the resource-count diagnostic (default-on; this
just makes the cadence explicit):

```
DARKBLOOM_MLX_RESOURCE_DEBUG=1 darkbloom serve
```

Watch the `[rsrc]` log line (subsystem `dev.darkbloom.provider`, category
`rsrc`). Before this change the live `resources` count climbs monotonically
toward 499000 under sustained traffic; after it should **plateau** well below
the limit. Optionally tune `DARKBLOOM_KV_CAPTURE_MAX_INFLIGHT` (default 2).

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/374"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784231935&installation_id=133247490&pr_number=374&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F374&signature=c31180d5314632a4d2fa896e71e56fb92d6a40e4d6f5c5a2c779d51586d70d88"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
@@ -83,6 +83,12 @@ public actor BatchScheduler {
     var checkpointBoundaries: [Int] = []
     /// Per-layer cache class/window signature for restore validation.
     var checkpointLayerSignatures: [CheckpointLayerSignature] = []
+    /// Bounded, single-consumer pipeline that drains checkpoint KV snapshots
+    /// into `checkpointManager`. Caps the number of live KV snapshots retained
+    /// in flight so a busy manager can't drive the live Metal buffer count to
+    /// the 499000 ceiling (the Gemma-4 leak). Built with the engine; shut down
+    /// in `stopCurrentEngine`. nil ⇒ feature off / no checkpoint manager.
+    var capturePipeline: CheckpointCapturePipeline<CheckpointCapture>?
 
     /// Engine-tier owner (EncryptedPrefixCachePersistence) for pure-
     /// attention models. Non-nil only when the model classifies as `.engine` AND
@@ -240,6 +246,7 @@ public actor BatchScheduler {
         self.checkpointBoundaries = build.checkpointBoundaries
         self.checkpointLayerSignatures = build.checkpointLayerSignatures
         self.engineTierOwner = build.engineTierOwner
+        self.capturePipeline = build.capturePipeline
         await engine.start()
         // Final epoch check after start() — start can suspend too.
         // Identity-checked cleanup — only nil self.engine if it's
@@ -251,6 +258,7 @@ public actor BatchScheduler {
             if self.checkpointBoundaries == build.checkpointBoundaries { self.checkpointBoundaries = [] }
             if self.checkpointLayerSignatures == build.checkpointLayerSignatures { self.checkpointLayerSignatures = [] }
             if self.engineTierOwner === build.engineTierOwner { self.engineTierOwner = nil }
+            if self.capturePipeline === build.capturePipeline { self.capturePipeline?.shutdown(); self.capturePipeline = nil }
             await engine.stop()
             return
         }
@@ -294,6 +302,7 @@ public actor BatchScheduler {
             if self.checkpointBoundaries == build.checkpointBoundaries { self.checkpointBoundaries = [] }
             if self.checkpointLayerSignatures == build.checkpointLayerSignatures { self.checkpointLayerSignatures = [] }
             if self.engineTierOwner === build.engineTierOwner { self.engineTierOwner = nil }
+            if self.capturePipeline === build.capturePipeline { self.capturePipeline?.shutdown(); self.capturePipeline = nil }
             await engine.stop()
             return
         }
@@ -333,6 +342,7 @@ public actor BatchScheduler {
                     if self.checkpointBoundaries == build.checkpointBoundaries { self.checkpointBoundaries = [] }
                     if self.checkpointLayerSignatures == build.checkpointLayerSignatures { self.checkpointLayerSignatures = [] }
                     if self.engineTierOwner === build.engineTierOwner { self.engineTierOwner = nil }
+                    if self.capturePipeline === build.capturePipeline { self.capturePipeline?.shutdown(); self.capturePipeline = nil }
                     await engine.stop()
                     return
                 }
@@ -818,11 +828,12 @@ public actor BatchScheduler {
             )
         } ?? []
         engine?.core.scheduler.checkpointBoundaries = boundaries
-        engine?.core.scheduler.onCheckpointCapture = { prefixTokens, length, caches in
-            let box = SendableKVCaches(caches)
-            // TB-016 sub-feature B: test seam also RAM-only (no eager flush).
-            Task { await mgr.store(tokens: prefixTokens, checkpointLength: length, caches: box) }
-        }
+        // Same bounded + admission-gated wiring production uses, so the test
+        // seam exercises the real backpressure path (not the old unbounded one).
+        self.capturePipeline?.shutdown()
+        let wiring = Self.makeCheckpointCaptureWiring(manager: mgr)
+        self.capturePipeline = wiring.pipeline
+        engine?.core.scheduler.onCheckpointCapture = wiring.hook
     }
 
     /// TEST SEAM: drive the real `finalizeRestore` fallback path without a live
@@ -956,6 +967,9 @@ public actor BatchScheduler {
         let checkpointBoundaries: [Int]
         let checkpointLayerSignatures: [CheckpointLayerSignature]
         let engineTierOwner: EncryptedPrefixCachePersistence?
+        /// Bounded capture pipeline (non-nil iff `checkpointManager` is). The
+        /// caller stores it on the actor and shuts it down at teardown.
+        let capturePipeline: CheckpointCapturePipeline<CheckpointCapture>?
     }
 
     private static func makeBatchedEngine(
@@ -1001,6 +1015,7 @@ public actor BatchScheduler {
 
             var enginePrefixCache: PrefixCache? = nil
             var checkpointManager: PrefixCacheManager? = nil
+            var capturePipeline: CheckpointCapturePipeline<CheckpointCapture>? = nil
             var boundaries: [Int] = []
             var checkpointLayerSignatures: [CheckpointLayerSignature] = []
             // Capture the engine-tier owner for accountant registration.
@@ -1112,16 +1127,14 @@ public actor BatchScheduler {
             // when a manager exists, so .engine/.none models are untouched.
             if let mgr = checkpointManager {
                 scheduler.checkpointBoundaries = boundaries
-                scheduler.onCheckpointCapture = { prefixTokens, length, caches in
-                    let box = SendableKVCaches(caches)
-                    // TB-016 sub-feature B: capture = RAM-ONLY. Store to RAM
-                    // (fast); 2nd-use promotion handles SSD persistence when the
-                    // prefix is re-accessed (RAM-first admission stops the write
-                    // storm). No eager flushToSSD.
-                    Task {
-                        await mgr.store(tokens: prefixTokens, checkpointLength: length, caches: box)
-                    }
-                }
+                // Bound the capture pipeline: at most `max-in-flight` live KV
+                // snapshots retained while the manager actor is busy (crypto +
+                // fsync), dropping the surplus rather than queuing it. This is
+                // the fix for the Gemma-4 Metal live-resource (499000) leak — the
+                // old `Task { await mgr.store(...) }` per boundary was unbounded.
+                let wiring = Self.makeCheckpointCaptureWiring(manager: mgr)
+                capturePipeline = wiring.pipeline
+                scheduler.onCheckpointCapture = wiring.hook
             }
             return EngineBuild(
                 engine: BatchedEngine(
@@ -1139,7 +1152,8 @@ public actor BatchScheduler {
                 checkpointManager: checkpointManager,
                 checkpointBoundaries: boundaries,
                 checkpointLayerSignatures: checkpointLayerSignatures,
-                engineTierOwner: engineTierOwner
+                engineTierOwner: engineTierOwner,
+                capturePipeline: capturePipeline
             )
         }
     }
@@ -1930,6 +1944,13 @@ public actor BatchScheduler {
         self.engine = nil
         modelContainer = nil
         tokenizer = nil
+        // Drain the bounded capture pipeline: the engine is stopped (no more
+        // capture hooks fire), so finish the stream and cancel the consumer.
+        // This releases any retained KV snapshots and stops an in-flight
+        // `mgr.store` from racing the index flush / accountant deregister below,
+        // and guarantees no snapshot (or consumer Task) leaks across a swap.
+        capturePipeline?.shutdown()
+        capturePipeline = nil
         // Persist any coalesced index writes before dropping the manager, so
         // checkpoints written since the last coalesced save survive restart.
         if let mgr = checkpointManager {
--- a/provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
@@ -89,6 +89,21 @@ final class CheckpointCapturePipeline<Payload: Sendable>: @unchecked Sendable {
         self.continuation = continuation
         self.consumer = Task {
             for await payload in stream {
+                // Honor shutdown before consuming. `shutdown()` calls
+                // `finish()` + `cancel()`, but `finish()` still delivers
+                // already-buffered payloads to this loop and an AsyncStream
+                // `for await` does NOT observe Task cancellation on its own.
+                // Without this guard we would `store(...)` old-model KV
+                // snapshots after a model swap has begun — pinning large live
+                // Metal buffers that must not overlap the next engine.
+                //
+                // We `continue` rather than `break`: the buffered remainder must
+                // still be pulled so each payload is released as it drains (the
+                // retained `continuation` keeps the stream's internal buffer —
+                // and anything stranded in it — alive). Skipping `consume` drops
+                // the captures (best-effort) without storing them; the finished
+                // stream then terminates the loop once its buffer empties.
+                if Task.isCancelled { continue }
                 await consume(payload)
             }
         }
--- a/provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
@@ -1,0 +1,233 @@
+// Copyright © 2026 Eigen Labs.
+//
+// Bounded, single-consumer pipeline for best-effort checkpoint KV capture.
+//
+// Background — the Gemma-4 `[metal::malloc] Resource limit (499000) exceeded`
+// crash. Heterogeneous models (Gemma-4: sliding [8,256] + full [2,512]) use the
+// `.checkpoint` prefix-cache tier. At each checkpoint boundary the engine hands
+// the provider a freshly-extracted, per-layer KV snapshot (live MLXArrays,
+// ≈ layers × 2 Metal buffers). The previous wiring spawned an UNBOUNDED
+// `Task { await mgr.store(...) }` per boundary:
+//
+//     scheduler.onCheckpointCapture = { prefixTokens, length, caches in
+//         let box = SendableKVCaches(caches)
+//         Task { await mgr.store(...) }   // <-- one detached Task per boundary
+//     }
+//
+// `PrefixCacheManager` is a single actor that serializes `store` behind
+// AES-GCM serialize + fsync and against every lookup/materialize. Under
+// sustained traffic the actor falls behind, the spawned Tasks queue, and each
+// queued Task keeps its live KV snapshot pinned until it runs. The live Metal
+// resource COUNT (not bytes/cache) climbs to the `iogpu.rsrc_limit` ceiling
+// (~499000) → crash. The mlx/mlx-c "count-aware" cache trims can only reclaim
+// CACHED buffers; these snapshots are LIVE, so trimming cannot help and the
+// idle-gated `Memory.clearCache()` never fires under load.
+//
+// This pipeline caps the number of snapshots retained in flight. Capture is
+// best-effort cache-warming: when the bounded buffer is full the surplus
+// snapshot is DROPPED (and released, freeing its Metal buffers) rather than
+// queued. Dropping a snapshot only costs a future cache miss, never
+// correctness.
+
+import Foundation
+import MLX
+import MLXLMCommon
+
+/// One checkpoint capture: the prefix tokens it covers, the checkpoint length,
+/// and the boxed per-layer KV snapshot to hand to `PrefixCacheManager.store`.
+/// `SendableKVCaches` is the ownership-transfer box; this struct is the
+/// production payload for ``CheckpointCapturePipeline``.
+struct CheckpointCapture: Sendable {
+    let tokens: [Int]
+    let length: Int
+    let caches: SendableKVCaches
+}
+
+/// Bounded, single-consumer delivery pipeline.
+///
+/// A small `AsyncStream` (buffering policy `.bufferingNewest(capacity)`) feeds a
+/// single long-lived consumer Task. `submit(_:)` is synchronous, non-blocking,
+/// and `@Sendable`, so it is safe to call from the engine queue inside
+/// `Scheduler.onCheckpointCapture`. At most `capacity` payloads sit in the
+/// buffer and the consumer holds at most one more while it `await`s, so the
+/// pipeline retains at most `capacity + 1` payloads — regardless of how far
+/// behind the consumer falls. (Counting the single payload being handed to
+/// `submit` during an evicting enqueue, at most `capacity + 2` are live for an
+/// instant.) Surplus payloads are dropped (and released) rather than queued —
+/// this small constant bound is what caps the live Metal buffer count and
+/// fixes the 499000 leak.
+///
+/// Generic over `Payload` so the bounding/drop/teardown logic is unit-testable
+/// without MLX (the production instantiation is `Payload == CheckpointCapture`).
+final class CheckpointCapturePipeline<Payload: Sendable>: @unchecked Sendable {
+    /// Maximum number of payloads buffered before the surplus is dropped.
+    let capacity: Int
+
+    private let continuation: AsyncStream<Payload>.Continuation
+    private let consumer: Task<Void, Never>
+
+    // Lightweight counters (lock-guarded; read from tests / future telemetry).
+    private let statsLock = NSLock()
+    private var _accepted = 0
+    private var _dropped = 0
+
+    /// - Parameters:
+    ///   - capacity: max buffered payloads (clamped to ≥ 1). Small by design
+    ///     (1–2): this is the hard cap on live KV snapshots retained in flight.
+    ///   - consume: the (async) sink. Invoked serially by the single consumer
+    ///     Task, one payload at a time.
+    init(
+        capacity: Int,
+        consume: @escaping @Sendable (Payload) async -> Void
+    ) {
+        let cap = max(1, capacity)
+        self.capacity = cap
+        let (stream, continuation) = AsyncStream.makeStream(
+            of: Payload.self,
+            bufferingPolicy: .bufferingNewest(cap)
+        )
+        self.continuation = continuation
+        self.consumer = Task {
+            for await payload in stream {
+                await consume(payload)
+            }
+        }
+    }
+
+    /// Enqueue a payload. Synchronous, non-blocking, `@Sendable`.
+    ///
+    /// Returns `true` when the payload was buffered without eviction, `false`
+    /// when the buffer was already full (an overflow occurred and the surplus
+    /// payload was dropped + released) or the pipeline has been shut down.
+    @discardableResult
+    func submit(_ payload: Payload) -> Bool {
+        switch continuation.yield(payload) {
+        case .enqueued:
+            statsLock.lock(); _accepted += 1; statsLock.unlock()
+            return true
+        case .dropped:
+            // `.bufferingNewest` keeps the newest `capacity` payloads and
+            // returns the evicted (oldest) one here; letting it go out of scope
+            // releases its KV snapshot so the Metal buffers free immediately.
+            statsLock.lock(); _dropped += 1; statsLock.unlock()
+            return false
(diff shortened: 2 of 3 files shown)


provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
```diff
@@ -0,0 +1,233 @@
...
+/// buffer and the consumer holds at most one more while it `await`s, so the
+/// pipeline retains at most `capacity + 1` payloads — regardless of how far
+/// behind the consumer falls. (Counting the single payload being handed to
+/// `submit` during an evicting enqueue, at most `capacity + 2` are live for an
+/// instant.) Surplus payloads are dropped (and released) rather than queued —
+/// this small constant bound is what caps the live Metal buffer count and
+/// fixes the 499000 leak.
+///
+/// Generic over `Payload` so the bounding/drop/teardown logic is unit-testable
+/// without MLX (the production instantiation is `Payload == CheckpointCapture`).
```
ethenotethan, line 61: 🔵 **[INFO]** ⚡ NSLock allocation on every pipeline instance for lightweight counters

💡 _Suggestion:_ Consider using atomic operations (OSAtomicIncrement32 or std::atomic equivalent) for simple counters instead of NSLock to reduce allocation overhead

📊 _Score: 2×3 = 6 · Category: Unbounded allocations_

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
```diff
@@ -0,0 +1,233 @@
...
+            return false
+        case .terminated:
+            return false
+        @unknown default:
+            return false
+        }
+    }
+
+    /// Count of `submit` calls that buffered without eviction.
+    var acceptedCount: Int { statsLock.lock(); defer { statsLock.unlock() }; return _accepted }
```
ethenotethan, line 137: 🟡 **[MEDIUM]** ⚡ Lock acquisition in getter methods called from hot paths

💡 _Suggestion:_ Cache counter values or use lock-free atomic operations since these are simple integer counters accessed frequently

📊 _Score: 2×4 = 8 · Category: Blocking I/O on hot paths_

provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
```diff
@@ -251,6 +258,7 @@ public actor BatchScheduler {
             if self.checkpointBoundaries == build.checkpointBoundaries { self.checkpointBoundaries = [] }
```
ethenotethan, line 258: 🟡 **[MEDIUM]** ⚡ Redundant pipeline shutdown calls in multiple cleanup paths

💡 _Suggestion:_ Add a shutdown flag to make shutdown() idempotent and avoid redundant work in cleanup paths

📊 _Score: 3×3 = 9 · Category: Repeated work_

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
```diff
@@ -0,0 +1,233 @@
...
+                tokens: capture.tokens,
+                checkpointLength: capture.length,
+                caches: capture.caches
+            )
+        }
+        prefixCacheLogger.info(
+            "checkpoint capture pipeline: max-in-flight \(maxInFlight) (drop-on-full), admission-gate at \(Int(captureResourcePressureFraction * 100))% of Metal resource limit")
+        let hook: CheckpointCaptureHook = { [pipeline] prefixTokens, length, caches in
+            // 1. Admission gate. Returning here lets the `caches` argument go
+            //    out of scope → its Metal buffers free immediately.
```
ethenotethan, line 233: 🟡 **[MEDIUM]** ⚡ captureResourcePressureHigh() called on every checkpoint capture

💡 _Suggestion:_ Cache the resource pressure check result for a short duration (e.g., 100ms) since MLX.Memory counters don't change rapidly

provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
ethenotethan, line 258: 🔵 **[INFO]** 🧩 Identical cleanup pattern repeated 4 times with same identity checks

💡 _Suggestion:_ Extract a helper method `cleanupBuildArtifacts(build:)` to eliminate the repeated identity-checked cleanup pattern

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
ethenotethan, line 61: 🔵 **[INFO]** 🧩 Generic pipeline class when only one concrete type is used

💡 _Suggestion:_ Consider making this concrete for CheckpointCapture since no other payload types are anticipated

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
```diff
@@ -0,0 +1,233 @@
...
+final class CheckpointCapturePipeline<Payload: Sendable>: @unchecked Sendable {
+    /// Maximum number of payloads buffered before the surplus is dropped.
+    let capacity: Int
+
+    private let continuation: AsyncStream<Payload>.Continuation
+    private let consumer: Task<Void, Never>
+
+    // Lightweight counters (lock-guarded; read from tests / future telemetry).
+    private let statsLock = NSLock()
+    private var _accepted = 0
```
ethenotethan, line 71: 🔵 **[INFO]** 🧩 Lock-guarded counters for simple statistics that could be atomic

💡 _Suggestion:_ Use OSAtomicIncrement64 or similar atomic operations instead of NSLock for simple counters

📊 _Score: 1×2 = 2 · Category: over-configuration_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:61` — NSLock allocation on every pipeline instance for lightweight counters
  - _Suggestion:_ Consider using atomic operations (OSAtomicIncrement32 or std::atomic equivalent) for simple counters instead of NSLock to reduce allocation overhead
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:120-122` — Lock acquisition in getter methods called from hot paths
  - _Suggestion:_ Cache counter values or use lock-free atomic operations since these are simple integer counters accessed frequently
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:258` — Redundant pipeline shutdown calls in multiple cleanup paths
  - _Suggestion:_ Add a shutdown flag to make shutdown() idempotent and avoid redundant work in cleanup paths
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:218` — captureResourcePressureHigh() called on every checkpoint capture
  - _Suggestion:_ Cache the resource pressure check result for a short duration (e.g., 100ms) since MLX.Memory counters don't change rapidly

- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:258` — Identical cleanup pattern repeated 4 times with same identity checks
  - _Suggestion:_ Extract a helper method `cleanupBuildArtifacts(build:)` to eliminate the repeated identity-checked cleanup pattern
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:61` — Generic pipeline class when only one concrete type is used
  - _Suggestion:_ Consider making this concrete for CheckpointCapture since no other payload types are anticipated
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:69-71` — Lock-guarded counters for simple statistics that could be atomic
  - _Suggestion:_ Use OSAtomicIncrement64 or similar atomic operations instead of NSLock for simple counters
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:1185-1188` — Complex capture pipeline construction logic embedded in engine builder
  - _Suggestion:_ Move the capture pipeline factory logic to a dedicated builder class or the CheckpointCapturePipeline itself

_8 finding(s) total, 4 blocking. Verdict: REQUEST_CHANGES._

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
ethenotethan: :white_check_mark: Looks addressed in `a776a51` — The code now uses atomic operations with NSLock for counter access instead of lock acquisition in getter methods.. Resolving.

Gajesh2007:
Addressed the cancellation/teardown review point in a776a512: `shutdown()` already called `finish()` + `cancel()`, but an AsyncStream `for await` doesn't observe Task cancellation on its own and `finish()` still delivers buffered payloads — so the consumer could still `store(...)` old-model KV snapshots after a model swap began. The loop now checks `Task.isCancelled` and **drain-and-drops** the buffered remainder (keeps pulling so the retained continuation doesn't strand it alive, but skips `consume`). Added a regression test (`capturePipelineDropsBufferedCapturesOnShutdown`) proving only the in-flight capture runs after shutdown and nothing leaks; full `capturePipeline` suite green.

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
```diff
@@ -0,0 +1,248 @@
...
+        capacity: Int,
+        consume: @escaping @Sendable (Payload) async -> Void
+    ) {
+        let cap = max(1, capacity)
+        self.capacity = cap
+        let (stream, continuation) = AsyncStream.makeStream(
+            of: Payload.self,
+            bufferingPolicy: .bufferingNewest(cap)
+        )
+        self.continuation = continuation
```
ethenotethan, line 89: 🟡 **[MEDIUM]** ⚡ AsyncStream with bufferingNewest policy may cause memory pressure under high load

💡 _Suggestion:_ Consider adding telemetry to monitor buffer utilization and dropped payloads to validate the capacity sizing in production

📊 _Score: 3×4 = 12 · Category: Unbounded allocations_

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
```diff
@@ -0,0 +1,248 @@
...
+    /// Maximum number of payloads buffered before the surplus is dropped.
+    let capacity: Int
+
+    private let continuation: AsyncStream<Payload>.Continuation
+    private let consumer: Task<Void, Never>
+
+    // Lightweight counters (lock-guarded; read from tests / future telemetry).
+    private let statsLock = NSLock()
+    private var _accepted = 0
+    private var _dropped = 0
```
ethenotethan, line 72: 🔵 **[INFO]** ⚡ NSLock usage in stats tracking could cause contention on high-frequency submit calls

💡 _Suggestion:_ Replace NSLock with atomic operations (OSAtomicIncrement64 or similar) for simple counters to eliminate lock contention

📊 _Score: 2×3 = 6 · Category: Blocking I/O on hot paths_

provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
```diff
@@ -1930,6 +1944,13 @@ public actor BatchScheduler {
         self.engine = nil
         modelContainer = nil
         tokenizer = nil
+        // Drain the bounded capture pipeline: the engine is stopped (no more
+        // capture hooks fire), so finish the stream and cancel the consumer.
+        // This releases any retained KV snapshots and stops an in-flight
+        // `mgr.store` from racing the index flush / accountant deregister below,
+        // and guarantees no snapshot (or consumer Task) leaks across a swap.
+        capturePipeline?.shutdown()
```
ethenotethan, line 1952: 🔵 **[INFO]** ⚡ Synchronous shutdown and drain operations in stopCurrentEngine may block actor

💡 _Suggestion:_ Consider making capturePipeline shutdown asynchronous or use a detached task to avoid blocking the actor during engine teardown

📊 _Score: 2×2 = 4 · Category: Blocking I/O on hot paths_

provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
ethenotethan, line 258: 🔵 **[INFO]** 🧩 Identical cleanup pattern repeated 4 times with same identity checks

💡 _Suggestion:_ Extract a helper method `cleanupBuildArtifacts(build:)` to eliminate the repeated identity-check cleanup pattern

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
```diff
@@ -0,0 +1,248 @@
...
+/// buffer and the consumer holds at most one more while it `await`s, so the
+/// pipeline retains at most `capacity + 1` payloads — regardless of how far
+/// behind the consumer falls. (Counting the single payload being handed to
+/// `submit` during an evicting enqueue, at most `capacity + 2` are live for an
+/// instant.) Surplus payloads are dropped (and released) rather than queued —
+/// this small constant bound is what caps the live Metal buffer count and
+/// fixes the 499000 leak.
+///
+/// Generic over `Payload` so the bounding/drop/teardown logic is unit-testable
+/// without MLX (the production instantiation is `Payload == CheckpointCapture`).
```
ethenotethan, line 61: 🔵 **[INFO]** 🧩 Generic type parameter adds complexity for single concrete use

💡 _Suggestion:_ Remove generic `<Payload: Sendable>` and hardcode `CheckpointCapture` since there's only one production use case

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
ethenotethan, line 72: 🔵 **[INFO]** 🧩 Lightweight counters with lock for simple integers

💡 _Suggestion:_ Use atomic operations or remove counters entirely if only used for testing

📊 _Score: 1×2 = 2 · Category: over-abstraction_

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
```diff
@@ -0,0 +1,248 @@
...
+        await consumer.value
+    }
+
+    deinit {
+        // Belt-and-suspenders: if the owner dropped us without `shutdown()`,
+        // make sure the stream terminates and the consumer Task ends.
+        continuation.finish()
+        consumer.cancel()
+    }
+}
```
ethenotethan, line 162: 🔵 **[INFO]** 🧩 Environment variable tuning for internal buffer size

💡 _Suggestion:_ Remove env var tuning - the default of 2 is already conservative and this adds unnecessary configuration surface

📊 _Score: 1×2 = 2 · Category: over-configuration_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:86-89` — AsyncStream with bufferingNewest policy may cause memory pressure under high load
  - _Suggestion:_ Consider adding telemetry to monitor buffer utilization and dropped payloads to validate the capacity sizing in production
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:70-72` — NSLock usage in stats tracking could cause contention on high-frequency submit calls
  - _Suggestion:_ Replace NSLock with atomic operations (OSAtomicIncrement64 or similar) for simple counters to eliminate lock contention
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:1945-1952` — Synchronous shutdown and drain operations in stopCurrentEngine may block actor
  - _Suggestion:_ Consider making capturePipeline shutdown asynchronous or use a detached task to avoid blocking the actor during engine teardown

### Additive_complexity — 4 finding(s)
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:258` — Identical cleanup pattern repeated 4 times with same identity checks
  - _Suggestion:_ Extract a helper method `cleanupBuildArtifacts(build:)` to eliminate the repeated identity-check cleanup pattern
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:61` — Generic type parameter adds complexity for single concrete use
  - _Suggestion:_ Remove generic `<Payload: Sendable>` and hardcode `CheckpointCapture` since there's only one production use case
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:70-72` — Lightweight counters with lock for simple integers
  - _Suggestion:_ Use atomic operations or remove counters entirely if only used for testing
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:158-162` — Environment variable tuning for internal buffer size
  - _Suggestion:_ Remove env var tuning - the default of 2 is already conservative and this adds unnecessary configuration surface

_7 finding(s) total, 1 blocking. Verdict: REQUEST_CHANGES._

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
ethenotethan, line 89: 🟡 **[MEDIUM]** ⚡ AsyncStream with bufferingNewest policy may still accumulate tasks under high contention

💡 _Suggestion:_ Consider adding explicit task cancellation tracking or using a semaphore-based approach to guarantee bounded concurrency

📊 _Score: 3×4 = 12 · Category: Unbounded goroutines_

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
```diff
@@ -0,0 +1,248 @@
...
+                // still be pulled so each payload is released as it drains (the
+                // retained `continuation` keeps the stream's internal buffer —
+                // and anything stranded in it — alive). Skipping `consume` drops
+                // the captures (best-effort) without storing them; the finished
+                // stream then terminates the loop once its buffer empties.
+                if Task.isCancelled { continue }
+                await consume(payload)
+            }
+        }
+    }
```
ethenotethan, line 110: 🔵 **[INFO]** ⚡ Consumer task processes payloads serially which could create backpressure

💡 _Suggestion:_ Consider parallel processing of payloads if the consume function is I/O bound, or document that this is intentionally serial

📊 _Score: 2×3 = 6 · Category: Blocking I/O on hot paths_

provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
ethenotethan, line 1952: 🔵 **[INFO]** ⚡ Shutdown called multiple times in different code paths without idempotency check

💡 _Suggestion:_ Add explicit idempotency check in shutdown() method or ensure callers coordinate to avoid redundant shutdown calls

📊 _Score: 2×2 = 4 · Category: Repeated work_

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
ethenotethan, line 61: 🔵 **[INFO]** 🧩 Generic pipeline class when only one concrete type is used

💡 _Suggestion:_ Consider making this non-generic since it's only used with CheckpointCapture. The generic abstraction adds complexity without clear benefit.

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
ethenotethan, line 72: 🔵 **[INFO]** 🧩 Lightweight counters with manual locking for simple statistics

💡 _Suggestion:_ Consider using atomic operations or removing these counters entirely if they're only for testing/debugging. The manual locking adds complexity for simple increment operations.

provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift
ethenotethan, line 258: 🟡 **[MEDIUM]** 🧩 Identical cleanup pattern repeated 4 times in the file

💡 _Suggestion:_ Extract the identity-checked cleanup pattern into a helper method to reduce duplication and maintenance burden.

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
ethenotethan, line 162: 🔵 **[INFO]** 🧩 Environment variable parsing with fallback for simple integer

💡 _Suggestion:_ Consider using a simpler approach or removing the environment variable entirely since this is an internal tuning parameter with a reasonable default.

provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift
```diff
@@ -0,0 +1,248 @@
...
+                //
+                // We `continue` rather than `break`: the buffered remainder must
+                // still be pulled so each payload is released as it drains (the
+                // retained `continuation` keeps the stream's internal buffer —
+                // and anything stranded in it — alive). Skipping `consume` drops
+                // the captures (best-effort) without storing them; the finished
+                // stream then terminates the loop once its buffer empties.
+                if Task.isCancelled { continue }
+                await consume(payload)
+            }
```
ethenotethan, line 108: 🔵 **[INFO]** 🧩 Task cancellation check in data processing loop

💡 _Suggestion:_ Consider handling cancellation at the Task level rather than checking it inside the processing loop, which mixes cancellation concerns with data processing logic.

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:86-89` — AsyncStream with bufferingNewest policy may still accumulate tasks under high contention
  - _Suggestion:_ Consider adding explicit task cancellation tracking or using a semaphore-based approach to guarantee bounded concurrency
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:95-110` — Consumer task processes payloads serially which could create backpressure
  - _Suggestion:_ Consider parallel processing of payloads if the consume function is I/O bound, or document that this is intentionally serial
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:1948-1952` — Shutdown called multiple times in different code paths without idempotency check
  - _Suggestion:_ Add explicit idempotency check in shutdown() method or ensure callers coordinate to avoid redundant shutdown calls

### Additive_complexity — 5 finding(s) (1 blocking)
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:61` — Generic pipeline class when only one concrete type is used
  - _Suggestion:_ Consider making this non-generic since it's only used with CheckpointCapture. The generic abstraction adds complexity without clear benefit.
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:70-72` — Lightweight counters with manual locking for simple statistics
  - _Suggestion:_ Consider using atomic operations or removing these counters entirely if they're only for testing/debugging. The manual locking adds complexity for simple increment operations.
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/Inference/BatchScheduler.swift:258` — Identical cleanup pattern repeated 4 times in the file
  - _Suggestion:_ Extract the identity-checked cleanup pattern into a helper method to reduce duplication and maintenance burden.
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:156-162` — Environment variable parsing with fallback for simple integer
  - _Suggestion:_ Consider using a simpler approach or removing the environment variable entirely since this is an internal tuning parameter with a reasonable default.
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCore/Inference/CheckpointCapturePipeline.swift:95-108` — Task cancellation check in data processing loop
  - _Suggestion:_ Consider handling cancellation at the Task level rather than checking it inside the processing loop, which mixes cancellation concerns with data processing logic.

_8 finding(s) total, 2 blocking. Verdict: REQUEST_CHANGES._

PR #382, opened 2026-06-17
feat: per-operation energy tracking (DAR-324)

Gajesh2007:
## Summary

Adds **per-operation energy tracking** so we can understand provider unit economics: measure real SoC energy on each Apple-Silicon provider via the **sudoless IOReport** API, attribute it to the operation the engine was doing (idle / prefill / decode / model-load), report it on the heartbeat, and **persist a per-provider time-series** in the coordinator for analysis.

The headline numbers it produces per model × chip: **J/decode-token**, **J/prefill-token**, **idle-floor watts**, and a per-subsystem split (CPU/GPU/ANE/DRAM) — enough to compute Wh/op and (with a $/kWh) $/request.

Refs **DAR-324**.

## Architecture — before / after

**Before** — power was a static estimate; no measured energy anywhere:

```mermaid
flowchart LR
  P["Provider (darkbloom)"] -->|"heartbeat: stats, system_metrics, backend_capacity"| RH["Registry.Heartbeat"]
  RH --> PS["Provider state"]
  PS --> STATS["GET /v1/stats"]
  STATS -->|"EstimateMachineWatts (static chip table)"| EST["active_power_watts<br/>(estimated)"]
```

**After** — measured energy is sampled, attributed per operation, reported, and persisted:

```mermaid
flowchart LR
  subgraph Provider["Provider (darkbloom)"]
    IOR["IOReportSampler<br/>/usr/lib/libIOReport.dylib"] --> ACC["EnergyAccountant<br/>EnergyLedger (phase bucketing)"]
    TOK["live decode frames +<br/>prefill tokens + inferenceActive"] --> ACC
    ACC --> HB["heartbeat<br/>+ power_watts<br/>+ energy ledger"]
    CLI["darkbloom energy CLI"] --> IOR
  end
  HB -->|WebSocket| RH["Registry.Heartbeat"]
  RH --> PS["Provider.Energy +<br/>SystemMetrics.PowerWatts"]
  PS --> STATS["GET /v1/stats"]
  STATS -->|"measured power, else estimate"| OUT["active_power_watts<br/>(measured)"]
  PS --> PE["persistEnergyThrottled (~60s)"]
  PE --> DB[("provider_energy<br/>time-series table")]
  DB --> AN["analysis:<br/>J/token, Wh/op, $/req"]
```

## How attribution works

IOReport's "Energy Model" group exposes **cumulative** joule counters (CPU/GPU/ANE/DRAM). We diff them each ~500ms tick and bucket the energy using **live** signals:

- **Classification:** `loading` (a backend slot is `reloading`) → load bucket; `inferenceActive` (the engine is serving) → active; otherwise idle — and idle is only counted as *warm* when a model is resident (machine idle with no model is excluded so it can't bias the floor).
- **Idle floor** is measured from sustained quiet ticks (stability-filtered against transient spikes).
- **Active ticks:** energy above the idle baseline (clamped to the tick's measured joules, so buckets never exceed measured energy) is bucketed by whether output tokens advanced this tick — ticks with **live** decode-token progress (≈ one per streamed content frame) are decode; active ticks with no output are prefill.
- **Per-token energy** is then a direct ratio — `decode_joules / decode_tokens`, `prefill_joules / prefill_tokens` — robust to continuous-batching timing.

> The first revision used a regression over completion-time token totals; PR review (independent verifier + GitHub Codex) correctly found those totals only update at request completion, so every streaming tick looked "idle." This revision drives classification + bucketing from live per-tick signals instead.

> Note: IOReport reports the same energy at nested levels (per-DVFS-state `*DTL*` → per-core → cluster → rollup); we read only the top rollup channels to avoid triple-counting. Channel names differ M1–M4 (`PACC`/`EACC`) vs M5 (`MCPU`).

## What changed

**Provider (Swift)** — new `ProviderCore/Hardware/` module:
- `IOReportSampler` — persistent IOReport subscription, per-subsystem joule deltas, graceful fallback when unavailable.
- `EnergyLedger` — idle/prefill/decode/load buckets + the regression + per-token coefficients.
- `EnergyAccountant` — 500ms accounting loop; snapshot attached to the heartbeat.
- `darkbloom energy` — live sudoless power readout.
- Wiring: prefill-token counter at request completion (`ProviderLoop`), `SystemMetrics.powerWatts` + `energy` on the heartbeat (`CoordinatorClient`, codec, `Messages`).

**Protocol** — `SystemMetrics.power_watts` + `EnergyLedger` (`messages.go` ↔ `Types.swift`/`Messages.swift`), all `omitempty`/encode-when-nonzero so legacy providers stay byte-compatible. Symmetry tests both sides.

**Coordinator** — `Registry.Heartbeat` ingests the ledger onto `Provider.Energy`; `persistEnergyThrottled` writes a snapshot/minute; `/v1/stats` prefers measured power over `EstimateMachineWatts` (same observed→estimate precedence as decode TPS).

**Store** — `ProviderEnergyRecord` + `RecordProviderEnergy`/`ProviderEnergySince` (interface, memory, postgres) + `provider_energy` table migration + round-trip test.

**Docs** — AGENTS.md + CLAUDE.md now require a before/after Mermaid diagram on every PR (this PR follows it).

## Verification

- **Swift:** full `swift build` ✅; `swift test` ✅ (`energyLedgerSnapshotSymmetry`, omitempty regression). Sampler + ledger validated on an M4 Max (live watts track load; regression recovers known coefficients).
- **Go:** `go build ./...` ✅, `gofmt` clean, `go test ./protocol/ ./store/ ./registry/ ./api/` ✅ (incl. energy round-trip).

## Addressed review feedback (commit 67137ae4)

- **P1** live token deltas — classify from `inferenceActive` + incremental streaming decode tokens; replaced the regression with phase bucketing.
- **P2** model-load activity wired (`loading` from `reloading` slot state + `noteModelLoad` on the rising edge) → `load_joules`/`model_loads` populate.
- **P2** skip warm-idle accounting when no model is resident.
- **P2** clamp the idle baseline to the tick's measured joules (conservative buckets).
- **P2** clamp untrusted `power_watts` (`MaxReasonablePowerWatts`) in `/v1/stats` and the heartbeat path.
- **P2** start the accountant at session start, not first heartbeat.
- Postgres `RecordProviderEnergy` returns its error; in-memory series bounded; per-session `ProviderID` + idle-floor caveats documented.

## Caveats / follow-ups

- **SoC vs wall power:** IOReport measures on-die energy, not wall-socket draw. The measured value augments (and falls back to) the existing wall estimate; a SoC→wall calibration (smart-plug ground truth) is the remaining open item.
- `loading` flag is stubbed `false` for v1 (load energy folds via the idle-stability filter); a precise model-load signal is a clean P2.
- Per-operation numbers populate once a provider runs a model under load → heartbeat → DB. `darkbloom energy` shows live machine power immediately.
- Out of scope (per issue): $/kWh sourcing, watts-per-token routing, console dashboards.

## Build note

`libs/mlx-swift*` are git submodules not auto-populated in fresh worktrees; run `git submodule update --init` before `swift build`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -224,6 +224,14 @@ Keep the codebase modular, never monolithic.
 - One file/component should do one thing. If a file mixes several concerns or grows past a few hundred lines, that's a signal to split it.
 - **At the end of every large piece of work, do a refactor pass to make it modular before calling it done.** Extract helpers/types/hooks into focused files, delete dead code, and keep the public entry point thin. The refactor must be behavior-preserving — build, lint, and tests stay green.
 
+## Pull Requests
+
+Every pull request **must** include a **before/after architecture diagram** — two Mermaid diagrams in the PR description, one showing the relevant flow/architecture *before* the change and one *after*. This is mandatory for every PR, no exceptions; it makes the structural impact of the change reviewable at a glance.
+
+- Use fenced ` ```mermaid ` blocks (GitHub renders them natively).
+- Scope the diagrams to the subsystem(s) the PR touches — the data flow, call path, or component wiring that changed — not the whole system.
+- For a purely additive change, the "before" diagram still shows the prior state so the delta is explicit.
+
 ## Formatting
 
 A pre-commit hook in `.githooks/pre-commit` checks staged files only. It is enabled via:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -260,6 +260,14 @@ Each reviewer should:
 
 Only proceed to the next objective after both reviewers pass. If either flags issues, fix them before moving on.
 
+## Pull Requests
+
+Every pull request **must** include a **before/after architecture diagram** — two Mermaid diagrams in the PR description, one showing the relevant flow/architecture *before* the change and one *after*. This is mandatory for every PR, no exceptions; it makes the structural impact of the change reviewable at a glance.
+
+- Use fenced ` ```mermaid ` blocks (GitHub renders them natively).
+- Scope the diagrams to the subsystem(s) the PR touches — the data flow, call path, or component wiring that changed — not the whole system.
+- For a purely additive change, the "before" diagram still shows the prior state so the delta is explicit.
+
 ## Git Hooks
 
 Hooks live in `.githooks/` and are enabled via `git config core.hooksPath .githooks` (already set for this repo).
--- a/coordinator/api/stats.go
+++ b/coordinator/api/stats.go
@@ -110,13 +110,15 @@ func (s *Server) handleStats(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		// Prefer the provider's measured SoC power (IOReport) when reported and
-		// within a sane bound; fall back to the static chip estimate otherwise.
-		// The bound stops a bogus/malicious heartbeat from inflating the public
-		// figure. Same observed→estimate precedence pattern used for decode TPS.
-		if pw := p.SystemMetrics.PowerWatts; pw > 0 && pw <= registry.MaxReasonablePowerWatts {
+		// plausible for THIS chip; fall back to the static estimate otherwise.
+		// Bounding to ~1.5x the chip's own estimate (rather than a loose global
+		// cap) stops a bogus/malicious heartbeat from inflating the public figure
+		// by an order of magnitude. Same observed→estimate precedence as decode TPS.
+		est := registry.EstimateMachineWatts(p.Hardware.ChipFamily, p.Hardware.ChipTier, p.Hardware.GPUCores)
+		if pw := p.SystemMetrics.PowerWatts; pw > 0 && pw <= est*1.5 {
 			activePowerWatts += pw
 		} else {
-			activePowerWatts += registry.EstimateMachineWatts(p.Hardware.ChipFamily, p.Hardware.ChipTier, p.Hardware.GPUCores)
+			activePowerWatts += est
 		}
 		totalRequests += p.Stats.RequestsServed
 		totalTokensGen += p.Stats.TokensGenerated
--- a/coordinator/api/stats.go
+++ b/coordinator/api/stats.go
@@ -109,10 +109,11 @@ func (s *Server) handleStats(w http.ResponseWriter, r *http.Request) {
 		if p.PrivateOnly {
 			return
 		}
-		// Prefer the provider's measured SoC power (IOReport) when reported;
-		// fall back to the static chip estimate otherwise. Same observed→estimate
-		// precedence pattern used for decode TPS.
-		if pw := p.SystemMetrics.PowerWatts; pw > 0 {
+		// Prefer the provider's measured SoC power (IOReport) when reported and
+		// within a sane bound; fall back to the static chip estimate otherwise.
+		// The bound stops a bogus/malicious heartbeat from inflating the public
+		// figure. Same observed→estimate precedence pattern used for decode TPS.
+		if pw := p.SystemMetrics.PowerWatts; pw > 0 && pw <= registry.MaxReasonablePowerWatts {
 			activePowerWatts += pw
 		} else {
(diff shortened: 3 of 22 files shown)


Gajesh2007:
Pushed `67137ae4` addressing all review findings (independent verifier + Codex):

- **P1 — live token deltas:** classification now uses live `inferenceActive` + an incremental streaming decode counter (≈ one per content frame) instead of completion-time totals. Replaced the regression with **phase bucketing** (bucket joules / bucket tokens). Re-validated with a realistic completion-timed harness: idle floor `8.00W`, `J/decode 0.750`, `J/prefill 0.0977` — all within tolerance; machine-idle-with-no-model correctly excluded.
- **P2 — model load:** `loading` derived from the `reloading` slot state; `noteModelLoad()` called on the rising edge → `load_joules`/`model_loads` populate.
- **P2 — idle when no model resident:** warm-idle accounting + floor learning now gated on `modelResident`.
- **P2 — conservative buckets:** idle baseline clamped to the tick's measured joules.
- **P2 — untrusted power:** `power_watts` clamped to `MaxReasonablePowerWatts` in `/v1/stats` and the heartbeat path.
- **P2 — sampling window:** accountant starts at session start, not first heartbeat.
- Postgres `RecordProviderEnergy` returns its error; in-memory series bounded; per-session `ProviderID` + idle-floor caveats documented.

Verification: `go build ./...`, `gofmt`, `go test ./protocol/ ./store/ ./registry/` and the `/v1/stats` power tests pass; full `swift build` + `swift test` (energy symmetry) green.

@codex review

Gajesh2007:
Pushed `6517d0aa` addressing the re-review (all three were valid):

- **Load state (CoordinatorClient.swift):** correct — the provider only emits `running`/`idle` slot states, so the `reloading` check was dead and `load_joules`/`model_loads` never populated. Now mirror `ProviderLoop.isLoadingAny` (the authoritative load critical-section flag, covering cold loads **and** coordinator preloads) to a shared atomic the accountant reads.
- **Off-model subsystem totals (EnergyLedger.swift):** correct — off-model ticks are now excluded from the CPU/GPU/ANE/DRAM totals as well, so the subsystem split matches the operation buckets.
- **MRU-model tagging (registry.go):** correct — the ledger is process-wide (IOReport measures whole-SoC energy, not splittable per concurrently-resident model). Documented that `model` is the active model at snapshot for context only, not a per-model attribution key; attribution is per-operation, which is the feature's intent.

Re-validated: realistic ledger harness, `go build`/`gofmt`/`go test`, full `swift build` + `swift test` all green.

@codex review

Gajesh2007:
Pushed `c24248ff` addressing the latest review. All five were valid:

- **Preserve decode attribution during preloads** + **treat token progress as active work** (EnergyLedger): token progress now forces "active" classification. A preload overlapping an already-serving slot credits `decode_joules` (not the whole tick to `load`), and the first ticks of a short request — before the capacity-derived `inferenceActive` refreshes — are attributed to decode/prefill instead of warm idle. (Verified with new harness cases: `stale=Y preload=Y`.)
- **Tighten the provider power clamp** (stats.go): accept measured power only up to ~1.5× the chip's own `EstimateMachineWatts` (was a loose 2000 W global), so a provider can't report ~1999 W and inflate `active_power_watts` by an order of magnitude.
- **Credit prefill tokens on canceled prefill work** (ProviderLoop): on a cancel-before-content early return, credit a `promptTokenFloor` so prefill_joules keeps a matching denominator (no `j_per_prefill_token` skew in cancellation-heavy runs).
- **Don't tag cumulative energy with only the MRU model** (registry): the ledger is process-wide (whole-SoC energy can't be split per concurrently-resident model), so `model` is left empty when >1 model is resident rather than mislabeled.
- The earlier "exclude off-model energy from subsystem totals" thread is already resolved in `6517d0aa` (off ticks return before subsystem accrual).

Re-validated: ledger harness (idle 8.00 W, J/decode 0.750, J/prefill 0.0977 + preload/stale edges), `go build`/`gofmt`/`go test`, full `swift build` + `swift test` all green.

PR #388, opened 2026-06-17
fix(provider): only sample cold prefills for observed_prefill_tps (no cache-hit overflow)

Gajesh2007:
## Root cause
`observed_prefill_tps` was sampled whenever `prefillSeconds > 0`. On a prefix-cache HIT the engine restores KV and emits the first token in micro/nanoseconds, so the admitted→first-token window collapses toward ~0 while `prefilledTokens` stays > 0 → **tps explodes to billions**, which the coordinator clamps to 5000 and which corrupts the routing-v2 TTFT estimate (and undermines the hard gate).

## Fix (`BatchScheduler+EngineBridge.swift`)
Sample only a genuine **cold** prefill and reject implausible measurements:
1. `restoredPrefixTokens == 0` (skip cache hits entirely — unrepresentative window)
2. `prefilledTokens > 0`
3. `prefillSeconds >= minPrefillWindowSeconds` (1ms floor — kills the near-zero-window artifact)
4. `tps` finite and `<= maxPlausiblePrefillTps` (8000; above the coordinator's 5000 clamp so legit fast cold prefills still register)

## Tests
`prefillEwmaSamplesColdPrefillOnly` (cold updates EWMA; cache-restore records nothing) and `prefillEwmaRejectsNearZeroWindow` (50µs window → no sample). `swift build` succeeds; `swift test --filter prefillEwma` passes. (GPU/metallib tests are env-gated and unrelated to this CPU-side bookkeeping change.)

Pairs with the coordinator stopgap #387. Ships in **provider 0.6.14**.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/388"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784273556&installation_id=133247490&pr_number=388&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F388&signature=cfa2213e713902f1d220d87aea64527fbe723ef5668d4926600347f5d603fe89"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/provider-swift/Sources/ProviderCore/Inference/BatchScheduler+EngineBridge.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/BatchScheduler+EngineBridge.swift
@@ -22,6 +22,26 @@ import MLXLMCommon
 
 extension BatchScheduler {
 
+    // MARK: - Prefill-EWMA sampling bounds
+    //
+    // `observed_prefill_tps` feeds the coordinator's routing-v2 TTFT estimate, so
+    // a single bogus sample skews routing for later cold prompts. We therefore
+    // sample ONLY a genuine cold prefill and drop implausible measurements.
+
+    /// Minimum admitted→first-token window (seconds) for a prefill sample to
+    /// count. A prefix-cache HIT restores the KV and emits the first token almost
+    /// instantly, collapsing the window toward ~0; dividing a positive prompt
+    /// length by a near-zero denominator produced the "billions of tok/s" readings
+    /// the coordinator clamped to 5000. 1ms is far below any real cold prefill yet
+    /// rejects that near-zero-window artifact.
+    static let minPrefillWindowSeconds = 0.001
+
+    /// Upper plausibility bound (tok/s) for a prefill sample. Set above the
+    /// coordinator's 5000 clamp so a legitimately fast cold prefill still
+    /// registers, while clearly impossible rates are dropped (defense in depth
+    /// behind the cold-only + window-floor guards).
+    static let maxPlausiblePrefillTps = 8000.0
+
     // MARK: - Bridge runner (called from `submit` in the main file)
 
     /// Spin up the per-request stream consumer that translates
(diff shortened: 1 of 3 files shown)


Gajesh2007:
Addressed the Codex P1 (engine-tier cache hits sampled as cold) in `58e26e7e`.

**Root cause (confirmed):** `restoredPrefixTokens` is only written on the **checkpoint** path (`BatchScheduler.swift:519`). The `.engine` tier (pure-attention models, in-GPU block `PrefixCache`) restores a matched prefix inside MLX and never sets it, so the `restoredPrefixTokens == 0` cold-guard misreads an engine-tier hit as a cold prefill and records `fullPrompt / collapsed-window`. A moderately-inflated value (below the coordinator's 5000 range) slips past the #387 backstop, re-poisoning the EWMA.

**Fix (B — surgical):** gate sampling additionally on `!enginePrefixCacheActive`, a computed property derived from `engineTierOwner` (single source of truth — non-nil exactly when an engine-tier prefix cache is active). Pure-attention models therefore skip prefill sampling entirely and fall back to the coordinator's TTFT heuristic. **Checkpoint-tier models (Gemma-4, GPT-OSS) are unaffected** — they set `restoredPrefixTokens` on a hit, so the existing cold-only guard already excludes their restores. Blast radius is just the niche pure-attention models.

Added a regression test (`prefillEwmaSkipsEngineTierCache`): engine-tier active ⇒ no sample even for a window that looks like a clean cold prefill. `BatchSchedulerBudgetTests` all pass locally.

**Follow-up (A — proper fix, tracked):** have the engine report reused tokens per request (an `mlx-swift-lm` change) so cold engine prefills can be sampled too instead of skipped.

PR #389, opened 2026-06-17
fix(routing): relaunch safety — debounce cold-dispatch kick, real-free cold-load admission, safe-by-default flags

Gajesh2007:
## Routing v2 relaunch safety

Follow-up to the routing-v2 production incident (write-lock storm + OOM cold loads). Stacked on #387 (prefill-overflow gate); base is `fix/prefill-tps-coordinator`. Together the stack is the **Tier-1 relaunch posture**: prefill gate fixed, cold-dispatch off, warm-pool observe-only — plus the safety fixes so Tier-2 (turning cold-dispatch back on) is safe later.

### Changes

**1. Debounce the cold-dispatch kick (write-lock storm)** — `coordinator/api/cold_dispatch.go`, `server.go`
`kickColdDispatch` fired `TriggerModelSwaps` (registry **write** lock) on every enqueue → O(request volume) lock acquisitions → reader starvation (`/v1/stats` 503s, stale heartbeats). New `coldKickState` coalesces into single-flight + min-interval: a burst of N enqueues collapses to one swap pass. Interval via `EIGENINFERENCE_COLD_DISPATCH_MIN_INTERVAL` (default 1s). Idempotent `TriggerModelSwaps` means dropped kicks lose nothing.

**2. Cold-load admission uses real free memory, not total (OOM loads)** — `coordinator/registry/scheduler.go`
The cold-load branch of `freeMemoryAdmits` checked `weights + kv + 4 <= TOTAL` (assuming the provider evicts everything), while the provider's authoritative `ModelLoadAdmission` gate checks against **real free** memory. Coordinator over-admitted → `insufficient memory to load model` → jetsam OOM → WS `read_error` disconnects. Now: `realFree = total − gpuMemoryActiveGB − 4` and require `weights + kv + 3 (activation reserve) <= realFree`. GPU *cache* is reclaimable so it is not subtracted.

> **Strictly safer:** at the branch `required = modelSize + kv`, so new admit ⇔ `modelSize+kv+7+gpuActive ≤ total` vs old `modelSize+kv+4 ≤ total`. For all `gpuActive ≥ 0` the new set is a subset of the old — it never admits a load the old check rejected, so it cannot over-admit relative to today.

TODO left in code: the clean long-term fix is for the provider to report `freeForLoadGb` in `BackendCapacity` (single source of truth) — intentionally not done here (no protocol change).

**3. Safe-by-default flags** — `coordinator/api/cold_dispatch.go`, `coordinator/registry/config.go`
[...]

PR #390, opened 2026-06-17
feat: provider reports free_for_load_gb; coordinator uses it for cold-load admission

Gajesh2007:
## Provider-reported `free_for_load_gb` — the single source of truth for cold-load admission

This is the proper "fix A" for the cold-load memory mismatch that drove the routing-v2 meltdown (and the residual Codex flagged on #389 P2). Instead of the coordinator re-deriving "can this provider load model X?" from gpu/total memory figures, the **provider reports the answer it already computes** via its authoritative `ModelLoadAdmission` gate.

### Why
The coordinator's re-derivation can't see the 90% unified-memory cap, the OS baseline, the activation/min-KV load headroom, or eviction of idle models. So it both:
- **over-admits** cold loads the provider then OOM-rejects (the meltdown's `insufficient memory to load model` → jetsam → WS `read_error`), and
- **under-admits** when idle-but-evictable models are resident (Codex #389 P2 #1).

There is no way to perfectly mirror the provider's gate from the coordinator — so report it.

### Provider (Swift)
- `ModelLoadAdmission.maxLoadableWeightGb(...)` — pure helper: max additional model **weight** (GB) loadable now, net of the load reserve (which enforces the 90% cap) + load headroom (activation 3 GiB + min serveable KV 1 GiB), clamped to real OS-available memory, treating current MLX usage as **reclaimable** (idle models are LRU-evicted on a cold load).
- `BackendCapacity.free_for_load_gb` populated in `updateAggregateCapacity` with the **same** `loadReserve`/headroom the load gate uses. **Invariant test:** for an idle box, the reported value equals the largest weight the provider's own `canLoad()` accepts.
- Backward-compatible: additive field, decode tolerates absence (defaults to 0).

### Coordinator (Go)
- `protocol.BackendCapacity.FreeForLoadGB *float64` (`omitempty`). Pointer ⇒ legacy provider that doesn't report it is `nil`.
- `freeMemoryAdmits` cold branch: if reported, **admit iff `modelSizeGB <= freeForLoadGb`** (exact mirror of the provider gate); else fall back to the existing heuristic.
- `clampBackendCapacity` nils out-of-range values (NaN/Inf/neg/absurd) → fall back rather than mis-admit; a legitimate `0` is preserved.

### How this behaves on a 24 GB machine (asked during review)
Production headroom = activation 3 + min-KV 1 = **4 GiB**; reserve = `max(configReserve 4, physical − 90% cap)` = **4 GiB**. With the OS reporting ~21–22 GB available on an idle 24 GB box:

`free_for_load_gb ≈ min(24, sysAvail) − 4 (reserve) − 4 (headroom) ≈ 13–14 GB`

So a 24 GB box advertises ~13–14 GB of loadable weight and the coordinator routes **gpt-oss-20b (~13.5 GB)** to it *only when it genuinely has the room*, and never sends it **gemma-26b (~14–31 GB)**. The machine reports its real, momentary capacity (it already does this for its own load gate — see `gptOssLoadsOn24GB`); the coordinator now respects it instead of guessing from total RAM (which would send a 24 GB box a 14 GB model: `14+4 ≤ 24` → admit → provider OOM/reject). It's self-tuning per machine — no hardcoded RAM tiers.

### Tests
- Swift: `maxLoadableWeightReclaimsIdleModels`, `…ClampsToSystemAvailable`, `…FloorsAtZero`, `…MirrorsCanLoadWhenIdle`, `backendCapacityRoundTripsFreeForLoad`, `…DecodesLegacyWithoutFreeForLoad`. `swift build --build-tests` clean; selected suites pass.
- Go: `freeMemoryAdmits` reported-value admit/reject, reported-0 rejects, legacy nil falls back, clamp validation. `go build ./...` + `go test ./...` green; gofmt clean.

### Note on stacking / reconciliation with #389
Branched off `master` (per request), so the fallback heuristic here is master's `modelSizeGB + kv + 4 ≤ total`. **#389** independently improves that same fallback to the evictable-idle real-free check. They touch the same cold branch, so when both land I'll reconcile to: **reported value (this PR) as the fast path + #389's evictable heuristic as the fallback.** No behavioral conflict — they converge.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/390"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784303033&installation_id=133247490&pr_number=390&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F390&signature=8c65eb62dc4c605c64f73ef33539eca6d6b545cbf28960a1db37b8f718f2ea76"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/protocol/messages.go
+++ b/coordinator/protocol/messages.go
@@ -232,6 +232,14 @@ type BackendCapacity struct {
 	GPUMemoryPeakGB   float64               `json:"gpu_memory_peak_gb"`   // Metal peak memory
 	GPUMemoryCacheGB  float64               `json:"gpu_memory_cache_gb"`  // Metal cache memory (reclaimable)
 	TotalMemoryGB     float64               `json:"total_memory_gb"`      // total system/GPU memory
+	// FreeForLoadGB is the max additional model-WEIGHT footprint (GB) the
+	// provider can load right now: net of the 90% unified-memory cap, OS/operator
+	// reserve, and activation+min-KV load headroom, clamped to real OS-available
+	// memory, and treating idle resident models as evictable. It is the single
+	// source of truth for cold-load admission (the coordinator no longer
+	// re-derives free memory). A pointer so a legacy provider that doesn't report
+	// it is nil (→ coordinator falls back to the total-memory heuristic).
+	FreeForLoadGB *float64 `json:"free_for_load_gb,omitempty"`
 }
 
 // SystemMetrics contains live resource utilization reported by a provider.
--- a/coordinator/registry/clamp_test.go
+++ b/coordinator/registry/clamp_test.go
@@ -84,6 +84,31 @@ func TestClampBackendCapacityMaliciousValues(t *testing.T) {
 	}
 }
 
+func TestClampBackendCapacityFreeForLoad(t *testing.T) {
+	logger := slog.New(slog.NewTextHandler(os.Stderr, nil))
+
+	// Out-of-range reported values are nilled (→ cold-load gate falls back to the
+	// heuristic) rather than trusted.
+	for _, bad := range []float64{math.NaN(), math.Inf(1), -3, 1e9} {
+		v := bad
+		bc := &protocol.BackendCapacity{TotalMemoryGB: 64, FreeForLoadGB: &v}
+		clampBackendCapacity(logger, "p1", bc)
+		if bc.FreeForLoadGB != nil {
+			t.Errorf("FreeForLoadGB=%v should be nilled, got %v", bad, *bc.FreeForLoadGB)
+		}
+	}
+
+	// A legitimate value (including 0) is preserved.
+	for _, ok := range []float64{0, 9, 128} {
+		v := ok
+		bc := &protocol.BackendCapacity{TotalMemoryGB: 64, FreeForLoadGB: &v}
+		clampBackendCapacity(logger, "p1", bc)
+		if bc.FreeForLoadGB == nil || *bc.FreeForLoadGB != ok {
+			t.Errorf("FreeForLoadGB=%v should be preserved, got %v", ok, bc.FreeForLoadGB)
+		}
+	}
+}
+
 func TestClampBackendCapacityReasonableValues(t *testing.T) {
 	// Realistic values should pass through unchanged.
 	logger := slog.New(slog.NewTextHandler(os.Stderr, nil))
--- a/coordinator/registry/cold_dispatch.go
+++ b/coordinator/registry/cold_dispatch.go
@@ -127,6 +127,14 @@ func (r *Registry) coldSpillProviderEligibleLocked(p *Provider, model string, tr
 		if !modelFitsHardware(entry.MinRAMGB, entry.SizeGB, float64(p.Hardware.MemoryGB)) {
 			return false
 		}
+		// Live free-capacity gate (shared helper): keep cold-spill in sync with the
+		// load planner. If the provider reports it cannot fit this model, don't spill
+		// the request into its queue — the planner (modelLoadCandidatePendingLocked)
+		// would refuse the load and the request would wait out the 120s queue timeout
+		// instead of failing fast (#390).
+		if admit, reported := reportedFreeForLoadAdmits(entry.SizeGB, backendFreeForLoadGB(p.BackendCapacity)); reported && !admit {
+			return false
+		}
 	}
 
 	return true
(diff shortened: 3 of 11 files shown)


Gajesh2007:
Addressed the Codex review in `037bcc51`.

**P2 #3 (over-reporting evictable memory) — fixed.** You're right: `MLX.GPU.activeMemory` is global and also covers the **local inference endpoint** (ProviderLoop runs `startLocalEndpoint`; local streams are tracked by `localReservations`, not `modelSlots`), so treating all MLX usage as reclaimable could advertise an actively-served model's weights as free. Fixed by gating the eviction add-back on **`hasInflightWork`** — the comprehensive signal that covers *both* coordinator-inflight (`inflightTasks`/`requestToModel`) **and** local streams (`localReservations`): assume idle models are evictable only when nothing is in flight; otherwise reclaim nothing (conservative — never over-reports).

**P2 #2 (planner ignored reported capacity) — fixed.** `modelLoadCandidatePendingLocked` (warm-pool / queue-before-shed cold dispatch) only checked static `modelFitsHardware`. Now it also skips any candidate whose catalog weight exceeds the provider's reported `free_for_load_gb`, mirroring the direct `freeMemoryAdmits` gate, so the planner can't send a `load_model` the provider then rejects (which would leave queued requests to time out). Legacy providers (nil) fall through to the static gate. Added `TestModelLoadCandidateRespectsFreeForLoad`.

**P2 #1 (request KV not in the cold gate) — by design, left as-is.** `free_for_load_gb` deliberately answers "can the model LOAD and serve a *modest* request" (the provider reserves min serveable KV in its headroom); per-request KV is a runtime concern handled by `GlobalKVCacheBudget` once loaded. Folding in the coordinator's *generic* KV estimate would **break edge-fit cold loads** — e.g. on a 24GB box gpt-oss uses the entire `free_for_load_gb` budget for weights, so `weights + anyKV > free_for_load_gb` → it'd never cold-load. The clean enhancement (a per-request/KV-aware load budget reported by the provider) is a tracked follow-up; for now a large-context cold request that doesn't fit is rejected at runtime and reroutes.

Go suite + Swift build green; gofmt clean.

Gajesh2007:
Second pass — all 4 new findings addressed in `be346d1e`, via one shared helper so the paths can't drift again.

**P2 #4 (units/padding mismatch) — fixed.** You're right that `free_for_load_gb` is padded GiB (provider's `canLoad` uses `estimatedMemoryGb` = on-disk ×1.2 / 2^30) while the coordinator compared raw `catalogSizeGB` (`TotalSizeBytes/1e9`). Added `coldLoadCatalogGBToMemGiB = 1.2 × (1e9/2^30) ≈ 1.1176` and normalize catalog GB → padded GiB before comparing, so the gate exactly mirrors `canLoad` (near-threshold models no longer admitted-then-503'd). Conservative by construction. New test `TestFreeMemoryAdmitsColdLoadNormalizesCatalogSize`.

**P2 #1 (warm-pool) + P2 #2 (cold-spill) — fixed.** Extracted a single `reportedFreeForLoadAdmits()` helper and applied it in **all four** cold-load paths: direct admission, the swap planner, `warmPoolCandidateLocked`, and `coldSpillProviderEligibleLocked`. Previously only the first two were gated, so the warm pool could issue a `load_model` the provider rejects, and cold-spill would queue a request the planner then refuses (sit until the 120s timeout). Now they're consistent — one helper, no drift.

**P2 #3 (outstanding KV) — fixed.** The heartbeat `free_for_load_gb` now subtracts `kvBudget.outstandingReservedBytes()`, exactly as the real load gate (`availableMemoryGb`) does, so it can't advertise reserved-but-unallocated bytes as loadable. `maxLoadableWeightGb` gained an `outstandingReservationBytes` param. New test `maxLoadableWeightSubtractsOutstandingReservations`.

Go suite (17 pkgs) + Swift build green; gofmt clean. The four cold-load paths now share one gate and the comparison is apples-to-apples with the provider's `canLoad`.

PR #395, opened 2026-06-17
feat(provider): only discover/advertise coordinator-catalog models

Gajesh2007:
## Summary

Make the Darkbloom provider treat the coordinator model catalog as the single source of truth for which local weights can be discovered, advertised, served, or diagnosed. Non-catalog weights are invisible to the network and to the operator-facing CLI.

### Before: any local MLX weight directory could be advertised

```mermaid
sequenceDiagram
    participant U as Operator
    participant P as darkbloom provider
    participant C as Coordinator
    participant R as Registry/Scheduler

    U->>P: darkbloom start --all
    P->>P: scan ~/.cache/lm-studio/Models
    P->>P: scan ~/.lmstudio/models
    P->>C: register with every discovered model
    C->>R: expose unsupported IDs as available capacity
    R->>C: route consumer requests to unsupported weights
```

### After: only coordinator-catalog models are discovered or advertised

```mermaid
sequenceDiagram
    participant U as Operator
    participant P as darkbloom provider
    participant C as Coordinator
    participant R as Registry/Scheduler

    U->>P: darkbloom start --all
    P->>C: GET /v1/models/catalog
    C-->>P: supported model IDs
    P->>P: scan local directories
    P->>P: filter discovered models against catalog
    P->>C: register only catalog-matching models
    C->>R: expose only supported IDs as available capacity
    R->>C: route consumer requests to supported weights only
```

### Provider (Swift)
- `loadRuntimeSnapshot` fetches `GET /v1/models/catalog` at startup and caches it to `~/.cache/darkbloom/catalog.json`. Offline commands fall back to the cache; if none exists, the catalog is treated as empty (fail closed).
- New `catalogFilteredModels()` and updated `advertisedModels()` filter every local model through the coordinator catalog before applying `enabled_models` or `--model`.
- `darkbloom start --all` now means "all downloaded models that are in the catalog."
- `darkbloom start --model <non-catalog-id>` fails with a clear error for explicit user invocations; the launchd `--foreground` path filters stale plist flags and logs warnings instead of refusing to start.
- `darkbloom doctor` diagnoses only catalog models and warns when the catalog cannot be fetched.
- `darkbloom models catalog` no longer shows a "Local only" section; off-catalog weights get a one-line "will not be served" note.
- `ProviderLoop.applyVerifiedPrefetch` refuses to advertise a prefetched build that is not in the catalog.
- `ProviderLoopConfig` carries the startup catalog snapshot.

### Coordinator (Go)
- Provider registration filters `msg.Models` to catalog-known IDs and logs any dropped non-catalog models as a defense-in-depth backstop.

### Tests & docs
- Added `CatalogFilterTests.swift` covering `catalogFilteredModels` and `advertisedModels`.
- Added `TestProviderRegistrationFiltersNonCatalogModels`.
- Updated `docs/provider/cli-reference.md` and `docs/provider/quickstart.md`.

## Verification

- `cd coordinator && go test ./... -count=1` ✅
- `cd provider-swift && swift build` ✅
- `cd provider-swift && swift test --filter "CLIDispatchTests|CatalogFilterTests|ModelPrefetchCoordinatorTests"` ✅
- Full `swift test` exits non-zero due to a pre-existing `MLX error: Failed to load the default metallib` crash in a live-model test, unrelated to these changes.

Closes the feedback where providers "actively try to use" models that the coordinator does not support.

### DIFF
--- a/coordinator/api/provider.go
+++ b/coordinator/api/provider.go
@@ -133,13 +133,20 @@ func (s *Server) handleProviderWS(w http.ResponseWriter, r *http.Request) {
 }
 
 // filterRegistrationModels returns the subset of provider-reported models that
-// are in the coordinator catalog, plus a list of any dropped IDs. A nil catalog
-// (dev/test) disables the gate and keeps all models.
+// are in the coordinator catalog or are a previous/retired member of an active
+// alias lineage, plus a list of any dropped IDs. A nil catalog (dev/test)
+// disables the gate and keeps all models.
+//
+// Retired alias builds are intentionally kept: a provider that was offline
+// through the end of an alias rollout may only advertise a retired build.
+// DesiredModelsForProvider uses that retired lineage to recognize the provider
+// as part of the alias fleet and push the current desired build. Dropping the
+// retired build here would strand that provider with no desired_models update.
 func (s *Server) filterRegistrationModels(models []protocol.ModelInfo) ([]protocol.ModelInfo, []string) {
 	var filtered []protocol.ModelInfo
 	var dropped []string
 	for _, m := range models {
-		if s.registry.IsModelInCatalog(m.ID) {
+		if s.registry.IsModelInCatalog(m.ID) || s.registry.IsAliasLineageBuild(m.ID) {
 			filtered = append(filtered, m)
 		} else {
 			dropped = append(dropped, m.ID)
--- a/coordinator/api/provider.go
+++ b/coordinator/api/provider.go
@@ -132,6 +132,22 @@ func (s *Server) handleProviderWS(w http.ResponseWriter, r *http.Request) {
 	s.providerReadLoop(r.Context(), conn, providerID, acmeResult, r)
 }
 
+// filterRegistrationModels returns the subset of provider-reported models that
+// are in the coordinator catalog, plus a list of any dropped IDs. A nil catalog
+// (dev/test) disables the gate and keeps all models.
+func (s *Server) filterRegistrationModels(models []protocol.ModelInfo) ([]protocol.ModelInfo, []string) {
+	var filtered []protocol.ModelInfo
+	var dropped []string
+	for _, m := range models {
+		if s.registry.IsModelInCatalog(m.ID) {
+			filtered = append(filtered, m)
+		} else {
+			dropped = append(dropped, m.ID)
+		}
+	}
+	return filtered, dropped
+}
+
 // providerReadLoop reads messages from the provider WebSocket and dispatches
 // them. It runs until the connection closes or the context is cancelled.
 func (s *Server) providerReadLoop(ctx context.Context, conn *websocket.Conn, providerID string, acmeResult *ACMEVerificationResult, r *http.Request) {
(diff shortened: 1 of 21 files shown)


Gajesh2007:
Addressed the Codex review feedback in commit `4b8d727a`:

1. **Fetch catalog from overridden coordinator** — `loadRuntimeSnapshot` now accepts an optional `coordinatorURLOverride` and `Start` passes `--coordinator-url` through so the catalog source and connect target stay consistent.

2. **Preserve explicit `--model` filters after catalog drops** — the launchd/foreground path now keys on the original `--model` list; if every pinned ID is filtered out, the provider advertises nothing instead of falling through to `--all`/default.

3. **Don't gate verified prefetches on a stale catalog** — removed the startup-snapshot catalog check from `applyVerifiedPrefetch`. `CatalogModelPrefetcher` already resolves against the live coordinator catalog, and `models_update` is the authoritative backstop.

4. **Keep retired alias builds for resync** — `filterRegistrationModels` now keeps models that are in the active catalog or are previous/retired members of an active alias lineage, so providers offline through a retirement still receive `desired_models`.

5. **Allow `--local` mode without a catalog** — local/direct mode now falls back to local model selection when no catalog is available, preserving offline/airgapped use. When a catalog is available it still filters normally.

Added tests for retired-alias registration preservation and `localAdvertisedModels`. Coordinator `go test ./...` and targeted Swift tests (`CatalogFilterTests`, `ModelPrefetchCoordinatorTests`, `CLIDispatchTests`) pass.

Gajesh2007:
Addressed the second round of Codex feedback in commit `8c04679c`:

1. **Doctor diagnoses from unfiltered local scan** — `RuntimeSnapshot` now carries both `models` (memory-filtered/loadable) and `allModels` (unfiltered). `darkbloom doctor` builds fit alternatives from the catalog-filtered set but diagnoses the target from the unfiltered scan, so too-large supported models are still flagged.

2. **Preserve alias lineage in foreground `--model` filters** — `catalogAllowedIDs()` includes active catalog models plus alias `previous_build` / `retired_builds`. Both `validateModelSelections` and `catalogFilteredOverrides` now use this set, so stale launchd plists containing retired/previous builds are kept until the coordinator can push a `desired_models` update.

3. **`darkbloom models list --all` shows every local weight** — `--all` now bypasses the coordinator catalog and `enabled_models` filters, using the unfiltered local scan so operators can inventory and remove unsupported downloads.

4. **`darkbloom benchmark` works offline** — when no catalog is available, benchmark falls back to local model selection instead of returning an empty list.

5. **Scope catalog cache to coordinator URL** — `CatalogCache.path(for:)` derives a per-coordinator filename from the normalized base URL. A transient fetch failure on one coordinator no longer falls back to a cache populated by another.

Added tests for `catalogAllowedIDs` and coordinator-scoped cache paths. Coordinator `go test ./...` and targeted Swift tests (`CLIDispatchTests`, `CatalogFilterTests`, `ModelPrefetchCoordinatorTests`) pass.

Gajesh2007:
Addressed the third round of Codex feedback in commit `85e5190b`:

1. **Doctor uses `--coordinator` override for catalog** — `DoctorCommand` now passes the override into `loadRuntimeSnapshot`, so `darkbloom doctor --coordinator <url>` fetches and caches the catalog from that coordinator.

2. **Validate daemon selections before installing launchd** — `launchDaemon` now fails early instead of installing a plist that is guaranteed to crash-loop: explicit `--model` IDs are checked against `catalogAllowedIDs()`, and `--all` is resolved to the actual catalog-supported model set before `LaunchAgent.installAndStart` is called.

Coordinator `go test ./...` and targeted Swift tests (`CLIDispatchTests`, `CatalogFilterTests`, `ModelPrefetchCoordinatorTests`) pass.

PR #399, opened 2026-06-18
fix(coordinator): fenced code-attestation reuse across provider version bumps

anupsv:
## Problem

Providers intermittently sit **hardware-trusted but not routed** — `status`/`doctor` show `trust=hardware / online`, yet the coordinator deroutes them as not code-attested. Operators report it **happens on every update** and **clears after a terminal reinstall**.

### Root cause

Code-identity attestation (`Provider.CodeAttested`) is a separate routing gate from trust level. With enforcement on, the single private-text chokepoint deroutes any provider that is `!CodeAttested`:

```go
// coordinator/registry/registry.go:465
if r.codeAttestationEnforcedLocked() && !p.CodeAttested { return false }
```

`CodeAttested` is re-armed cheaply (no APNs push) by the reuse cache — but that cache is **version-gated**:

```go
// coordinator/api/code_attest_throttle.go (reuseAttestation)
if !ok || r.version != version || t.now().Sub(r.at) >= t.reuseWindow { return false }
```

Every `darkbloom update` bumps `ProviderCore.version`, so the post-update reconnect **misses reuse on the version mismatch** and must complete a fresh, budget-throttled APNs code-identity round-trip. Meanwhile a reconnect starts `CodeAttested=false` (fresh per-connection `Provider`), so there is a guaranteed window on **every update** where the provider is hardware-trusted but derouted — and it stays derouted until the round-trip lands (or indefinitely for a box that can't promptly re-acquire its APNs token). A terminal reinstall lays down a healthy bundle whose next round-trip succeeds, which is why a reinstall "fixes" it.

This is purely the **version gate** forcing the re-attest — the self-updater preserves the signed `.app` bundle, provisioning profile, and entitlements correctly; the binary identity is intact across an update.

## Fix

Permit code-attestation reuse **across a version change** for the same device — but only behind binary-identity fences proven on the *live* connection, so it can't become a downgrade/bad-build hole.

- New `reuseAttestationCrossVersion(seKey, token)` in the throttle: same as `reuseAttestation` minus the version match (still requires same SE key, fresh-within-window, same APNs token).
- `codeAttestLoop` invokes it **only when all hold**:
  - `AttestationResult.Valid` (valid registration attestation),
  - `RuntimeVerified` && `RuntimeManifestChecked` (runtime-manifest verification),
  - `ChallengeVerifiedSIP` (SIP-verified live challenge),
  - same APNs token (token rotation still fails closed),
  - version **at/above `MIN_PROVIDER_VERSION`** (`semverLess` floor — closes the downgrade hole).
- New `reused_cross_version` metric so the path is observable.

### Why the fences are required (not just dropping the version match)

`NodeKeyPair` is regenerated per provider startup, so a post-update reconnect always presents a **new** X25519 encryption key. The SE key + APNs token alone prove the device and push target, but **not** that the current binary is a legitimate, non-downgraded build. The version match in `reuseAttestation` was loosely standing in for that. Dropping it without the runtime/manifest/SIP/min-version fences would let a downgraded or disallowed binary re-arm `CodeAttested` from a prior proof without a fresh challenge. The fences re-establish binary identity from the live connection's own verification state; same-version reuse and token-rotation fail-closed are unchanged.

## Tests

- `TestCrossVersionReuseAboveFloorSameTokenReuses` — fenced, same-token, above-floor version bump **reuses** (no push); `CodeAttested` set so routing continues through the update.
- `TestCrossVersionReuseBelowFloorForcesChallenge` — a below-`MIN_PROVIDER_VERSION` downgrade **forces a real challenge** (downgrade hole closed).
- `TestCrossVersionReuseTokenChangeForcesChallenge` — a rotated APNs token **fails closed** across versions.
- `TestCrossVersionReuseUnfencedForcesChallenge` — a version bump without the runtime/SIP fences **forces a real challenge**.
- `TestSeededWrongVersionRowForcesRealChallenge` retained (now documents the unfenced fall-through; same-version `reuseAttestation` is unchanged).

`go build ./...`, `go vet ./api/`, and the full `coordinator/api` suite pass.

## Scope / what this does NOT change

- Coordinator-only; no protocol or Swift provider changes.
- Same-version reuse, token-rotation fail-closed, and the enforcement gate itself are untouched.
- This shrinks the forced-re-attest window to ~zero for **healthy** updates. It does not address a provider that genuinely cannot complete the APNs round-trip post-relaunch (e.g. fails to re-acquire its APNs token) — if logs show a stuck provider hitting `no_token`/`timeout`, that is a separate provider-side relaunch fix, tracked independently.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/399"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784409429&installation_id=133247490&pr_number=399&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F399&signature=27fd372c45325cf936a9a9dc7596c45a9ac9c30aba6a38604fa6b4e19b4c694d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/registry/registry.go
+++ b/coordinator/registry/registry.go
@@ -654,6 +654,45 @@ func (p *Provider) SetCodeAttested(v bool) {
 	p.CodeAttested = v
 }
 
+// CodeIdentityState is the live code-identity snapshot a reuse decision evaluates.
+// APNsDeviceToken/Version can rotate concurrently (maybeRearmCodeAttest), so the
+// decision must use these, not values captured earlier.
+type CodeIdentityState struct {
+	APNsDeviceToken        string
+	Version                string
+	SEPublicKey            string
+	AttestationValid       bool
+	RuntimeVerified        bool
+	RuntimeManifestChecked bool
+	ChallengeVerifiedSIP   bool
+}
+
+// GrantCodeAttestedIf runs `decide` against the live state and sets
+// CodeAttested=true iff it returns true — atomically under the provider lock, so a
+// concurrent token rotation can't interleave between the decision and the grant
+// (closes the rotation TOCTOU). `decide` must not take this provider's lock; it
+// may take others (e.g. the throttle) — lock order is always provider → throttle.
+func (p *Provider) GrantCodeAttestedIf(decide func(CodeIdentityState) bool) bool {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	st := CodeIdentityState{
+		APNsDeviceToken:        p.APNsDeviceToken,
+		Version:                p.Version,
+		AttestationValid:       p.AttestationResult != nil && p.AttestationResult.Valid,
+		RuntimeVerified:        p.RuntimeVerified,
+		RuntimeManifestChecked: p.RuntimeManifestChecked,
+		ChallengeVerifiedSIP:   p.ChallengeVerifiedSIP,
+	}
+	if p.AttestationResult != nil {
+		st.SEPublicKey = p.AttestationResult.PublicKey
+	}
+	if !decide(st) {
+		return false
+	}
+	p.CodeAttested = true
+	return true
+}
+
 // GetCodeAttested reports whether this connection passed code-identity
 // attestation (thread-safe).
 func (p *Provider) GetCodeAttested() bool {
--- a/coordinator/api/code_attest_throttle.go
+++ b/coordinator/api/code_attest_throttle.go
@@ -161,6 +161,26 @@ func (t *codeAttestThrottle) reuseAttestation(seKey, version, token string) bool
 	return r.token == "" || r.token == token
 }
 
+// reuseAttestationCrossVersion is reuseAttestation without the version match —
+// it rides a recent proof across a version bump (which every update causes) so a
+// healthy update isn't forced into a fresh push. Dropping the version alone is
+// unsafe (SE key + token prove the device, not that the binary is legitimate), so
+// the caller (codeAttestLoop) only invokes it behind live binary-identity fences.
+// The token match is STRICT here: both stored and current token must be non-empty
+// and equal (unlike reuseAttestation's legacy empty-token leniency).
+func (t *codeAttestThrottle) reuseAttestationCrossVersion(seKey, token string) bool {
+	if seKey == "" || token == "" {
+		return false
+	}
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	r, ok := t.attested[seKey]
+	if !ok || t.now().Sub(r.at) >= t.reuseWindow {
+		return false
+	}
+	return r.token != "" && r.token == token
+}
+
 // pushCooldown returns the per-device push budget for the active delivery mode.
 func (t *codeAttestThrottle) pushCooldown(alert bool) time.Duration {
 	if alert {
--- a/coordinator/api/code_attest_w5fix2_test.go
+++ b/coordinator/api/code_attest_w5fix2_test.go
@@ -362,9 +362,13 @@ func TestSeededStalePersistedRowForcesRealChallenge(t *testing.T) {
 	}
 }
 
-// TestSeededWrongVersionRowForcesRealChallenge proves the version gate survives
-// persistence: a seeded row for a DIFFERENT binary version must not be reused; the
-// connection must run a real challenge.
+// TestSeededWrongVersionRowForcesRealChallenge proves the SAME-VERSION reuse gate
+// survives persistence AND that an unfenced cross-version reconnect still forces a
+// real challenge: a seeded row for a DIFFERENT binary version is not reusable via
+// reuseAttestation, and codeAttestLoop falls through to a push because this
+// provider does not satisfy the cross-version fences (RuntimeVerified /
+// RuntimeManifestChecked / ChallengeVerifiedSIP are unset). The fenced
+// cross-version reuse path is covered by TestCrossVersionReuse* below.
 func TestSeededWrongVersionRowForcesRealChallenge(t *testing.T) {
 	logger := quietLogger()
 	st := store.NewMemory(store.Config{})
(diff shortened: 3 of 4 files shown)


anupsv:
## Review findings addressed (force-pushed)

Three issues raised on the first revision, all verified against source and fixed:

**1. (High) SIP fence raced — snapshotted once before the loop.** The cross-version fences (notably `ChallengeVerifiedSIP`) were read once at `codeAttestLoop` entry, but they're armed *concurrently* by the challenge/registration paths on a fresh connection — so on a real post-update reconnect they're usually still false at entry, the cross-version reuse is skipped, and the provider falls into the full push loop anyway (the fix would silently not fire). Extracted `tryCrossVersionReuse`, which **re-reads the fences live under the lock**, and call it **both up front and on every loop iteration** before deciding to push — so once a concurrent challenge flips `ChallengeVerifiedSIP=true`, the next iteration reuses instead of burning the push budget. New test `TestCrossVersionReuseSIPArmedMidLoopReuses` pins this (SIP armed mid-loop → attests after exactly one push, no completed round-trip); passes under `-race`.

**2. (High) Empty version passed the min-version floor.** `version == ""` short-circuited `aboveMinVersion` to true; since `RegisterMessage.version` is optional, an unversioned provider must NOT satisfy a configured `MIN_PROVIDER_VERSION`. Changed to `s.minProviderVersion == "" || (version != "" && !semverLess(version, s.minProviderVersion))`. New test `TestCrossVersionReuseEmptyVersionForcesChallenge`.

**3. (Medium) Cross-version reuse accepted empty stored token.** `reuseAttestationCrossVersion` returned `r.token == "" || r.token == token`, contradicting the "same token" fence (old DB rows can have `apns_token=''`). Now requires **both stored and current token non-empty and equal** (`seKey == "" || token == ""` rejected up front; `r.token != "" && r.token == token`). Same-version reuse keeps its legacy empty-token leniency (to avoid a fleet-wide re-push on deploy) — only cross-version is tightened.

`go build ./...`, `go vet ./api/`, full `coordinator/api` suite, and the cross-version tests under `-race` all pass.

anupsv:
## Addressed: token-rotation TOCTOU in cross-version reuse (force-pushed)

Confirmed the race against source and fixed it. The hole: `tryCrossVersionReuse` evaluated the fences/token captured at loop start, then granted `CodeAttested` after dropping the provider lock. A concurrent heartbeat token rotation (`maybeRearmCodeAttest`) publishes the new token + clears `CodeAttested` under the lock, then invalidates the old reuse row — so an in-flight old loop could reuse the **old-token** proof in that window, set `CodeAttested=true`, and the rearm loop could exit without proving the new token. That breaks the "token rotation forces a real challenge" guarantee.

**Fix — make the decision and the grant atomic under the provider lock against the LIVE state:**

- New `Provider.GrantCodeAttestedIf(decide func(CodeIdentityState) bool) bool` (`registry/registry.go`): snapshots the live `APNsDeviceToken`, `Version`, SE key, and fences under `p.mu`, runs `decide`, and sets `CodeAttested=true` **iff** it returns true — all in one lock acquisition. `tryCrossVersionReuse` now does its fence check + `reuseAttestationCrossVersion(liveSEKey, liveToken)` + grant inside that callback, using the **live** token, never a captured one.
- Lock order is always provider → throttle (throttle methods never touch a provider), so calling the throttle reuse-check from inside the provider-locked decision is deadlock-safe; verified `maybeRearmCodeAttest` releases `p.mu` before `invalidateReuse`, so the two never deadlock and the rotation is serialized against the grant.

**Why this is airtight:** a rotation either (a) wins the lock first → the grant reads the new token, the old-token row no longer matches → reuse fails → challenge forced; or (b) the grant runs first on the old token → the rotation's locked section then overwrites `CodeAttested=false` and invalidates the row. Either way a rotated token can never end up `CodeAttested` via a stale-token proof.

**Test:** `TestCrossVersionReuseUsesLiveTokenNotCaptured` — reuse record bound to `oldtok`, live provider token already `newtok`: `tryCrossVersionReuse` refuses to grant, and the loop forces a real challenge. All 7 cross-version tests pass under `-race`; full `coordinator/api` + `registry` suites green.

Gajesh2007:
@codex review this

coordinator/registry/registry.go
```diff
@@ -654,6 +654,49 @@ func (p *Provider) SetCodeAttested(v bool) {
...
+	ChallengeVerifiedSIP   bool
+}
+
+// GrantCodeAttestedIf evaluates `decide` against the provider's LIVE
+// code-identity state while holding the provider lock, and sets CodeAttested=true
+// iff it returns true — all atomically. This closes the rotation TOCTOU: a token
+// rotation that flips CodeAttested=false and updates APNsDeviceToken cannot
+// interleave between the reuse decision and the grant, because both happen under
+// the same lock acquisition. `decide` MUST NOT touch this provider's lock (it may
+// take other locks, e.g. the code-attest throttle, since the lock order is always
```
Gajesh2007: compress down the comments

anupsv:
## Response to threat-model review

Thanks — all four pre-merge items verified against source. Three were already closed by existing invariants; one (forensic logging) is now added.

**Rec 2 — `code_attest_throttle.go` lock-order (deadlock risk): ✅ verified safe.** Grepped the whole file: it acquires only its own `t.mu` and never references `provider.mu`, `SetCodeAttested`, or any registry/provider state. So the declared order (provider → throttle) cannot invert. `maybeRearmCodeAttest` also releases `p.mu` *before* calling `invalidateReuse`, so the rotation path never holds both locks either. No deadlock.

**Rec 3 — does the reuse row require a *verified* SE-signed challenge (not just an observed pair)? ✅ verified — already closed.** `recordAttested` has exactly one caller: `handleCodeAttestationResponse` (`provider.go:872`), reached only after (a) the nonce matches a challenge *the coordinator* pushed to this device (`:857`) and (b) `attestation.VerifyChallengeSignature` succeeds against the **registration-bound** SE key — "never a key supplied in the response" (`:863-869`). So a reuse row can only exist after a genuine SE signature over a coordinator-issued nonce. The registration-spoofing / reuse-oracle angle is structurally excluded: a provider registering with a victim's `(token, SE pubkey)` cannot produce a `recordAttested` row without the victim's hardware SE signing, and `tryCrossVersionReuse` additionally requires `ChallengeVerifiedSIP` on *this* connection — also gated on a verified signature (`verifyChallengeResponse:1252` verifies nonce + registered-key match + SE P-256 signature before `ChallengeVerifiedSIP` is set at `:1676`).

**Rec 1 — stale X25519 key after cross-version reuse: ✅ verified safe.** The X25519 rotation is not trusted via reuse. Dispatch encrypts to the provider's **live** `PublicKey` read at encrypt time (`dispatch.go:701/710/732`), not a value captured when the request was queued; `DrainQueuedRequestsForProvider` just re-runs `ReserveProviderEx` + normal dispatch, so the drain seals to the current key. Re-registration replaces `PublicKey` on the provider, and reuse only fires after `ChallengeVerifiedSIP` is set by a challenge round-trip *to this connection's current key*. There is no path that dispatches with a pre-reconnect encryption key.

**Rec 4 — log which fence blocked reuse: ✅ added.** `tryCrossVersionReuse` now emits a debug log with a `blocked_by` field naming the first failing fence (`attestation_invalid` / `runtime_unverified` / `runtime_manifest_unchecked` / `sip_challenge_pending` / `below_min_version` / `no_fresh_same_token_proof`), so a stuck reuse loop is diagnosable without source. Behavior-neutral; all 7 cross-version tests still pass under `-race`.

**T-036 / T-033 / T-037** assessments match my reading: this sets only `CodeAttested` (not `TrustHardware` — the MDM/MDA chain is untouched), and the atomic `GrantCodeAttestedIf` collapses the rotation TOCTOU. SEC-034/SEC-035 are pre-existing and untouched by this PR.

Force-pushed (`c6ace64e`).

coordinator/registry/registry.go
anupsv: Done — compressed the doc comments on `CodeIdentityState`, `GrantCodeAttestedIf`, and `tryCrossVersionReuse` (the latter had two overlapping blocks; merged into one). Behavior unchanged; build + cross-version tests still green. Force-pushed.

PR #409, opened 2026-06-19
fix(gpt-oss): reliability fixes — Jinja 500s, long-prompt cancels, load failures, partial-success accounting (DAR-329/330/331/332)

Gajesh2007:
## Summary

Fixes the four post-deploy gpt-oss reliability issues identified after `0908ecb3` (#401/#402/#406 + soft-TTFT). Together these address the live failure mix verified from prod telemetry (post-deploy era): ~10% provider 500s (Jinja), ~12% long-prompt client cancellations, residual load/dispatch failures, and ~4.6% partial-success with no dashboard visibility. All changes are additive and backward-compatible; the long-prompt routing preference is flag-gated and default-OFF.

## Before / After

A single gpt-oss request walks the same lifecycle in both diagrams; the four DAR issues are the failure points (red) on the left and the fixes (blue) on the right.

### Before (master `0908ecb3`)

```mermaid
flowchart TD
    A[gpt-oss request] --> B{Route / admit}
    B --> C[Provider selected<br/>long prompt: no special routing]
    C --> D{Model resident?}
    D -- "no - OOM on load" --> E[load_model fails<br/>pending held full 2 min<br/>approx queue timeout]:::fail
    E --> X1[DAR-331: dispatch_exhausted /<br/>insufficient memory<br/>no reject telemetry]:::fail
    D -- yes --> F[Build Jinja chat template]
    F -- "null in tool args/schema" --> X2[DAR-329: cannot convert<br/>Optional value to Jinja<br/>HTTP 500 provider_error ~10%]:::fail
    F -- ok --> G[Prefill long prompt]
    G --> H{First token before<br/>client gives up?}
    H -- "no - client disconnects" --> X3[DAR-330: client_gone / cancelled ~12%<br/>not sliced by prompt size or phase]:::fail
    H -- yes --> I[Stream then provider completes]
    I -- "client gone after commit" --> X4[DAR-332: partial_success ~4.6%<br/>invisible vs success on dashboards]:::fail
    I -- "client connected" --> J([success]):::ok
    classDef fail fill:#3b0d0d,stroke:#f1707a,color:#ffd7d7;
    classDef ok fill:#0d3b1e,stroke:#5fd08a,color:#d7ffe4;
```

### After (this PR)

```mermaid
flowchart TD
    A[gpt-oss request] --> B{Route / admit}
    B --> C[Fastest-prefill warm provider<br/>long-prompt preference - flag-gated]:::fix
    C --> D{Model resident?}
[...]

PR #422, opened 2026-06-20
fix(gpt-oss): strip inbound Harmony <|channel|> tags + normalized inference error_reason telemetry — DAR-341

Gajesh2007:
## Summary

Fixes the dominant remaining gpt-oss error: `gpt-oss-20b` (Harmony) returns **HTTP 500** on multi-turn / tool-calling requests because its `chat_template.jinja` **deliberately `raise_exception`s** when an assistant message's `content` (or `thinking`) still contains raw Harmony channel framing:

```
TemplateException: You have passed a message containing <|channel|> tags in the content field.
Instead of doing this, you should pass analysis messages ... in the 'thinking' field,
and final messages ... in the 'content' field.
```

~35% of gpt-oss **tool** requests vs ~0.1% non-tool. Failover doesn't help (`attempts=1`) — it's content-triggered. Our own provider output is already tag-free; the framing is **inbound** (a client/upstream replays a prior assistant turn with raw `<|channel|>analysis<|message|>…<|end|><|channel|>final<|message|>…`). We render it verbatim and the guard aborts.

This is a **distinct, previously-unfixed mode** from DAR-329 / #409 (the swift-jinja null-bridge crash). Both collapsed to the same opaque `"(Jinja.TemplateException error 1.)"` 500, which is why this one stayed masked — so this PR also closes the telemetry gap that hid it.

Linear: **DAR-341**

## Root cause

Harmony's replay contract: for prior (non-last) assistant turns at inference (`add_generation_prompt=true`) the template **drops analysis/CoT** and re-renders only `<|channel|>final<|message|>` + the answer, with **no** channel tags allowed in `content`/`thinking`. So the fix is to normalize inbound replayed assistant strings to that contract **before** `applyChatTemplate`.

### Provider (Swift) — the fix
- **New shared sanitizer** `ProviderCoreFoundation/HarmonySanitization.swift` → `stripHarmonyChannelFraming(fromAssistantContent:)`: unwraps `<|channel|>final<|message|>…` to the final answer and drops `<|channel|>analysis<|message|>…<|end|>` blocks. **Conservative** — strings without `<|channel|>` are returned byte-identical.
- **One implementation, every chokepoint**: wired into `sanitizeJinjaMessages` (covers stream / `/apply-template` / prompt recount), `BatchScheduler.submit`, and the scan-time `TemplateRenderCheck` self-check — preserving the "renders in the self-check == renders at request time" invariant.

### Telemetry gap (provider + coordinator)
- Normalized **`error_reason`** classified on the **provider** from the real `Error` at the catch site — `jinja_channel_tags`, `jinja_null_bridge`, `jinja_template`, `model_load` — and plumbed `OutboundMessage → codec → wire`. (The wire previously only carried the lossy `localizedDescription` = `"(Jinja.TemplateException error 1.)"`.) Logs the **offending message index + role only**, never content (E2E privacy).
- **Coordinator** persists `error_reason` on `inference_routes` (in-memory + Postgres + idempotent `ADD COLUMN IF NOT EXISTS` migration). Provider value wins (**whitelisted** so raw provider text can't leak into the DB), else derived from existing class/code, else `unknown`.
- **Datadog** `inference.error{reason,model}` counter so the modes are alertable, not log-archaeology.

### Version
- Provider `0.6.16 → 0.6.17`.

## Testing

- **Coordinator**: `go build`, `go vet`, `gofmt -l`, `go test ./protocol ./store ./api` — all clean.
- **Provider**: full `swift build` clean; unit tests green — Harmony stripping (8) + render fixtures (6) + error classifier (incl. privacy assertions) + protocol round-trip (`error_reason` omitempty).
- **Live, real `mlx-community/gpt-oss-20b-MXFP4-Q8`** (`DARKBLOOM_LIVE_MLX_TESTS=1`):
  ```
  [live] plain:         completion_tokens=34
  [live] null-tool:     completion_tokens=51
  [live] harmony-replay: completion_tokens=128  content="It is sunny, as…"
  ```
  The exact 500 scenario (raw `<|channel|>` framing in a replayed assistant turn) now renders + generates, and the model correctly received the stripped final answer as context. ✅

## Acceptance criteria
- [x] Multi-turn request with `<|channel|>` framing in assistant `content` renders + serves (no 500).
- [x] New live test passes against the real gpt-oss model.
- [ ] gpt-oss tool-call 500 rate falls to ~baseline **after the fleet converges to v0.6.17** (deploy-time).

## Release note (follow-up, intentionally not in this PR)
Shipping v0.6.17 is a release step: CI signs/notarizes the bundle, then `LatestProviderVersion` / the release store is bumped. Per AGENTS.md these must stay coordinated with the bundle upload, so `coordinator/api/server.go` is intentionally untouched here. The new `error_reason` field is optional/omitempty → wire-compatible with older providers and coordinators.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/422"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784527768&installation_id=133247490&pr_number=422&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F422&signature=a07faf2a8a279cd01e28ed646fe605af877b489c813d4249fb374241ea6013b6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


Gajesh2007:
Thanks — addressed the threat-model review in `10a130b0`.

**1. T-034 — `errorReason` must be advisory-only (verified, no code change needed).**
Audited every coordinator use of `ErrorReason`: it flows only into telemetry persistence (`store/route_telemetry.go`, `store/postgres.go`, `api/admin_telemetry.go`) and the Datadog tag (`api/route_outcome.go`). No routing, trust-level, billing, or auto-recovery decision branches on it. It is also **whitelisted** before persistence — `normalizeInferenceErrorReason` maps any non-vocabulary value to `"unknown"`, so an adversarial provider can neither inject raw text into the DB/metric tags nor steer a decision.

**2. Prompt-bleed into the classifier `haystack` (fixed — added guard comment).**
`String(describing: error)` can embed prompt-derived template text. Confirmed it is only used to derive the normalized label and is never returned/logged/transmitted. Added an explicit `// PRIVACY: NEVER log haystack/described …` comment at the construction site so future debugging changes don't regress this.

**3. `offendingHarmonyMessageLocation` scan scope (fixed).**
Aligned the diagnostic scan to **assistant-only**, matching both the Harmony guard and the sanitizer scope (`stripHarmonyFramingFromMessage`). A `<|channel|>` token in a user/system/tool message neither trips the guard nor is stripped, so flagging it would have been misleading. Updated the test that encoded the old behavior and added a non-assistant-roles guard test.

**4. SEC-016 / T-008 (plaintext SSE fallback on encryption failure).**
Acknowledged as pre-existing and out of scope for this PR; this change introduces no regression there (the encryption-failure path still returns `inference_error` with `errorReason: nil`).

All provider unit tests + the live gpt-oss test remain green; coordinator `go build`/`vet`/`test` clean.

coordinator/store/postgres.go
```diff
@@ -1697,7 +1704,7 @@ func (s *PostgresStore) InferenceRouteRecordsSince(since time.Time) []InferenceR
 			&r.CreatedAt, &r.UpdatedAt,
```
ethenotethan, line 1704: 🟡 **[MEDIUM]** ⚡ InferenceRouteRecordsSince scans all rows without LIMIT clause

💡 _Suggestion:_ Add LIMIT clause to the SELECT query or implement pagination to prevent unbounded memory allocation when the table grows large

coordinator/api/dispatch.go
```diff
@@ -334,15 +336,36 @@ func (d *dispatchState) successRoutingOutcome() *store.InferenceRouteOutcome {
...
-	out := &store.InferenceRouteOutcome{
-		FinalStatus: status,
-		ErrorCode:   code,
-		ErrorClass:  class,
+	providerReason, errorText := "", ""
+	if routeOutcomeUsesProviderErrorText(class) {
+		providerReason = d.lastErrReason
+		errorText = d.lastErr
 	}
+	out := routeOutcomeWithReason(status, class, code, providerReason, errorText)
```
ethenotethan, line 344: 🔵 **[INFO]** ⚡ routeOutcomeUsesProviderErrorText performs redundant string operations

💡 _Suggestion:_ Cache the normalized class string or use a switch statement on the original class to avoid repeated ToLower/TrimSpace calls

coordinator/api/route_outcome.go
```diff
@@ -129,10 +189,60 @@ func completeRouteOutcome(pr *registry.PendingRequest, usage protocol.UsageInfo,
...
+
+func normalizeInferenceErrorReason(reason string) string {
+	reason = strings.ToLower(strings.TrimSpace(reason))
+	reason = strings.ReplaceAll(reason, "-", "_")
+	if reason == "" {
+		return ""
+	}
+	if _, ok := validInferenceErrorReasons[reason]; ok {
+		return reason
+	}
```
ethenotethan, line 242: 🟡 **[MEDIUM]** ⚡ inferenceErrorReason performs multiple redundant string normalizations

💡 _Suggestion:_ Normalize strings once at the beginning of the function and reuse the normalized values throughout the switch statement

coordinator/api/dispatch.go
```diff
@@ -110,14 +111,15 @@ type dispatchState struct {
...
-	lastErrCode int
-	committed   bool
+	provider      *registry.Provider
+	pr            *registry.PendingRequest
+	requestID     string
+	firstChunk    string
+	heldChunks    []string
+	lastErr       string
+	lastErrCode   int
+	lastErrReason string
```
ethenotethan, line 121: 🔵 **[INFO]** 🧩 Three similar error-setting methods with overlapping responsibilities

💡 _Suggestion:_ Consolidate setLastError and setLastInferenceError into a single method that handles both cases, or make the distinction clearer

coordinator/api/route_outcome.go
```diff
@@ -9,30 +10,79 @@ import (
...
+	errorReasonQueueFull          = "queue_full"
+	errorReasonTokenBudgetExhaust = "token_budget_exhausted"
+	errorReasonCancelled          = "cancelled"
+	errorReasonProviderError      = "provider_error"
+	errorReasonUnknown            = "unknown"
+)
+
+var validInferenceErrorReasons = map[string]struct{}{
+	errorReasonJinjaChannelTags:   {},
+	errorReasonJinjaNullBridge:    {},
```
ethenotethan, line 30: 🔵 **[INFO]** 🧩 Large enum of error reason constants that could grow unbounded

💡 _Suggestion:_ Consider grouping related error reasons or using a more structured approach to avoid a long flat list of constants

coordinator/api/route_outcome.go
```diff
@@ -129,10 +189,60 @@ func completeRouteOutcome(pr *registry.PendingRequest, usage protocol.UsageInfo,
...
+func normalizeInferenceErrorReason(reason string) string {
+	reason = strings.ToLower(strings.TrimSpace(reason))
+	reason = strings.ReplaceAll(reason, "-", "_")
+	if reason == "" {
+		return ""
+	}
+	if _, ok := validInferenceErrorReasons[reason]; ok {
+		return reason
+	}
+	return errorReasonUnknown
```
ethenotethan, line 243: 🟡 **[MEDIUM]** 🧩 Complex error reason derivation logic with many string matching conditions

💡 _Suggestion:_ Extract error classification into a dedicated type with clear rules, or use a lookup table instead of nested conditionals

coordinator/store/postgres.go
```diff
@@ -1610,27 +1615,28 @@ func (s *PostgresStore) UpdateInferenceRouteOutcome(requestID string, attempt in
...
+			used_backup = COALESCE(used_backup, FALSE) OR $22,
+			backup_won = COALESCE(backup_won, FALSE) OR $23,
 			updated_at = NOW()
 		 WHERE request_id = $1 AND attempt = $2`,
 		requestID, attempt,
-		outcome.FinalStatus, outcome.ErrorCode, outcome.ErrorClass, outcome.PromptTokens, outcome.CompletionTokens, outcome.ReasoningTokens,
+		outcome.FinalStatus, outcome.ErrorCode, outcome.ErrorClass, outcome.ErrorReason, outcome.PromptTokens, outcome.CompletionTokens, outcome.ReasoningTokens,
 		outcome.CostMicroUSD, outcome.ActualTTFTMs, outcome.DispatchToFirstChunkMs, outcome.TotalDurationMs,
 		outcome.ParseMs, outcome.ReserveMs, outcome.RouteMs, outcome.EncryptMs, outcome.QueueWaitMs, outcome.DispatchMs, outcome.ActualDecodeTPS,
 		outcome.AdmittedButFailed, outcome.UsedBackup, outcome.BackupWon,
```
ethenotethan, line 1642: 🔵 **[INFO]** 🧩 Parameter numbering becomes error-prone with many positional arguments

💡 _Suggestion:_ Consider using named parameters or a struct to reduce the risk of parameter misalignment in this complex SQL update

provider-swift/Sources/ProviderCore/Inference/InferenceErrorClassifier.swift
```diff
@@ -0,0 +1,88 @@
...
+    if haystack.contains("Cannot convert value of type") && haystack.contains("Jinja Value") {
+        return "jinja_null_bridge"
+    }
+
+    // Any other Jinja / template render failure.
+    if haystack.contains("TemplateException") || haystack.contains("Jinja") {
+        return "jinja_template"
+    }
+
+    return nil
```
ethenotethan, line 61: 🟡 **[MEDIUM]** 🧩 String concatenation and multiple contains() checks create brittle error classification

💡 _Suggestion:_ Use a more structured approach like regex patterns or a decision tree to make error classification more maintainable

provider-swift/Sources/ProviderCoreFoundation/HarmonySanitization.swift
```diff
@@ -0,0 +1,60 @@
...
+            answer = String(afterFinal)
+        }
+    }
+
+    if controlTokens.contains(where: { answer.contains($0) }) {
+        for token in controlTokens {
+            answer = answer.replacingOccurrences(of: token, with: "")
+        }
+    }
+    return answer
```
ethenotethan, line 59: 🔵 **[INFO]** 🧩 Complex string parsing logic that could be simplified

💡 _Suggestion:_ Consider using a regex or parser library instead of manual string manipulation with multiple loops and conditions

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🟡 **[MEDIUM]** `coordinator/store/postgres.go:1665-1704` — InferenceRouteRecordsSince scans all rows without LIMIT clause
  - _Suggestion:_ Add LIMIT clause to the SELECT query or implement pagination to prevent unbounded memory allocation when the table grows large
- 🔵 **[INFO]** `coordinator/api/dispatch.go:336-344` — routeOutcomeUsesProviderErrorText performs redundant string operations
  - _Suggestion:_ Cache the normalized class string or use a switch statement on the original class to avoid repeated ToLower/TrimSpace calls
- 🟡 **[MEDIUM]** `coordinator/api/route_outcome.go:204-242` — inferenceErrorReason performs multiple redundant string normalizations
  - _Suggestion:_ Normalize strings once at the beginning of the function and reuse the normalized values throughout the switch statement

### Additive_complexity — 6 finding(s) (2 blocking)
- 🔵 **[INFO]** `coordinator/api/dispatch.go:111-121` — Three similar error-setting methods with overlapping responsibilities
  - _Suggestion:_ Consolidate setLastError and setLastInferenceError into a single method that handles both cases, or make the distinction clearer
- 🔵 **[INFO]** `coordinator/api/route_outcome.go:15-30` — Large enum of error reason constants that could grow unbounded
  - _Suggestion:_ Consider grouping related error reasons or using a more structured approach to avoid a long flat list of constants
- 🟡 **[MEDIUM]** `coordinator/api/route_outcome.go:197-243` — Complex error reason derivation logic with many string matching conditions
  - _Suggestion:_ Extract error classification into a dedicated type with clear rules, or use a lookup table instead of nested conditionals
- 🔵 **[INFO]** `coordinator/store/postgres.go:1615-1642` — Parameter numbering becomes error-prone with many positional arguments
  - _Suggestion:_ Consider using named parameters or a struct to reduce the risk of parameter misalignment in this complex SQL update
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/Inference/InferenceErrorClassifier.swift:38-61` — String concatenation and multiple contains() checks create brittle error classification
  - _Suggestion:_ Use a more structured approach like regex patterns or a decision tree to make error classification more maintainable
- 🔵 **[INFO]** `provider-swift/Sources/ProviderCoreFoundation/HarmonySanitization.swift:20-59` — Complex string parsing logic that could be simplified
  - _Suggestion:_ Consider using a regex or parser library instead of manual string manipulation with multiple loops and conditions

_9 finding(s) total, 4 blocking. Verdict: REQUEST_CHANGES._

PR #425, opened 2026-06-20
fix(console-ui): stop provider dashboard KPI/vitals text overlap

### LINKED ISSUE #424: Provider dashboard numbers overlap into an unreadable mess
Noticed this on the Provider Dashboard today — the earnings numbers in the Fleet header are rendering on top of each other. Instead of five separate KPIs I get `$0.0003$0.0003$0.0003$0.00033/91` smeared together, and the machine card's vitals do the same thing (`THERMAL Nominal` runs straight into `MEMORY 57%`). It's pretty obvious once you see it — I'll drop a screenshot in below.

Hit it on a normal laptop-width window with the sidebar open, one online machine, sub-cent earnings (`$0.000350`). Gets worse the narrower you make the window.

I dug into why: the grids in `FleetHealthStrip.tsx` and `CardVitals.tsx` pick their column count off the *viewport* width (`sm:`/`md:`/`lg:`), but the whole dashboard lives inside the sidebar + the `max-w-6xl` container + that `36%/1fr` split. So it happily goes to 5 columns when each one is only ~100px wide, and `$0.000350` is wider than that. Since the string has no spaces it can't wrap, so it just overflows into the next column. Classic "breakpoints assume the full viewport but something already ate half the width" situation.

I've got a fix going (switching those grids to container queries so they size off their actual container instead of the viewport, plus making the values wrap/shrink instead of overflow as a backstop). Checked it at a bunch of widths — 1320 / 1012 / 700 / 480px — and there's no overlap anywhere now; the numbers stay on one line when there's room and stack/wrap cleanly when it's genuinely tight. Tests and build are green. PR to follow.

Files involved:
- `console-ui/src/app/providers/dashboard/FleetHealthStrip.tsx`
- `console-ui/src/app/providers/dashboard/CardVitals.tsx`

YellowFoxH4XOR:
## Summary

The provider dashboard's fleet-health strip crammed KPI values and vital
labels into each other at certain widths. This reworks the strip to use
container queries keyed off the strip's real width (not the viewport), with
breakpoints chosen so the side-by-side split only fires where the 5-column
KPI row actually fits, and adds an overflow guard so a long vital value can't
push its column past its `1fr` allocation and overlap the neighbor.

## Linked issue

Closes #424 

## Test plan

- [x] `make ui-lint` — 0 errors (pre-existing warnings only)
- [x] `make ui-test` — 309 vitest tests pass (incl. `MachineCard`, `CardVitals`-adjacent suites)
- [x] `make ui-build` — Next build succeeds, no unknown-breakpoint warnings
- [x] Live Chrome DevTools sweep across strip widths — measured computed
      `grid-template-columns` and per-column overflow at each step:

  | strip width | verdict/KPI split | KPI columns | overlap |
  |---|---|---|---|
  | <28rem | stacked | 2 | none |
  | 42–60rem | stacked | 5 | none |
  | ≥64rem | side-by-side (36% / 1fr) | 3 | none |

  Split fires only at strip ≥64rem; 5-column KPIs appear when the keyed
  region ≥42rem; no KPI/vital value clips into a neighbor with real data.

## Components touched

- [ ] coordinator (Go)
- [ ] provider (Rust, legacy)
- [x] console-ui (Next.js)
- [ ] enclave (Swift)
- [ ] infra / CI / release
- [ ] docs

## Notes for reviewers

- CSS-only change (two Tailwind utility edits): the verdict/KPI split moves
  from `@3xl` (48rem, ~768px — split fired too early on narrow desktops) to
  `@5xl` (64rem, the original `lg` viewport intent expressed as a container
  query).
- The KPI row intentionally stays at `@2xl:grid-cols-5`, **not** `@5xl`:
  `@5xl` would key off the right-region container reaching 64rem, which never
  happens inside the `max-w-6xl` (72rem) shell, so it would permanently cap
  the row at 3 columns and kill the 5-column layout.
- Tradeoff: in side-by-side state the right region tops out at ~41rem (just
  under the 42rem trigger), so it shows 3 clean columns rather than 5. That
  preserves the no-overlap guarantee, which is the actual goal.
- Earlier review feedback claimed `@3xl`/`@2xl` were *undefined* in Tailwind
  v4 — they are defined (48rem / 42rem). The real issue was the threshold
  values, not missing utilities.

### DIFF
--- a/console-ui/src/app/providers/dashboard/CardVitals.tsx
+++ b/console-ui/src/app/providers/dashboard/CardVitals.tsx
@@ -20,9 +20,9 @@ function Vital({
 }) {
   return (
     <div className="space-y-1.5">
-      <div className="flex items-center justify-between gap-2">
-        <span className="text-[10px] uppercase tracking-wider text-text-tertiary">{label}</span>
-        <span className="text-xs font-mono font-semibold text-text-primary capitalize tabular-nums">{value}</span>
+      <div className="flex items-center justify-between gap-2 min-w-0">
+        <span className="text-[10px] uppercase tracking-wider text-text-tertiary truncate">{label}</span>
+        <span className="text-xs font-mono font-semibold text-text-primary capitalize tabular-nums shrink-0">{value}</span>
       </div>
       {children}
     </div>
@@ -52,9 +52,9 @@ export function CardVitals({
   const decodePct = fleetMaxDecodeTps > 0 ? (decode / fleetMaxDecodeTps) * 100 : 0;
 
   return (
-    <div className="px-4 py-3 space-y-3">
+    <div className="px-4 py-3 space-y-3 @container">
       {sm && (
-        <div className="grid grid-cols-2 md:grid-cols-4 gap-x-4 gap-y-3">
+        <div className="grid grid-cols-2 @lg:grid-cols-4 gap-x-4 gap-y-3">
           <Vital label="Thermal" value={sm.thermal_state}>
             <ThermalPips state={sm.thermal_state} />
           </Vital>
--- a/console-ui/src/app/providers/dashboard/CardVitals.tsx
+++ b/console-ui/src/app/providers/dashboard/CardVitals.tsx
@@ -19,7 +19,7 @@ function Vital({
   children: React.ReactNode;
 }) {
   return (
-    <div className="space-y-1.5">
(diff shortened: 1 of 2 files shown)


nilswx:
Good one, earnings table is indeed unreadable on mobile (portrait). Bump!

PR #431, opened 2026-06-20
fix(coordinator): let zero-balance owners use their own machine

Gajesh2007:
## Problem

If you run your own provider but have **no credits**, you can't use your own machine. The console "Use my machine" toggle sends `X-Darkbloom-Route: prefer`, and the `prefer` path takes a worst-case balance reservation **up front** (so it can fall back to the paid fleet). A zero-balance owner failed that reservation and got a **402 `insufficient_funds`** at `coordinator/api/consumer.go` — before routing ever considered their own (free) machine. Only the exclusive `self` route skipped billing, but it has no fallback and the UI never uses it.

## Fix

When a `prefer` request fails the reservation with `ErrInsufficientBalance` **and** the owner's own machine can currently serve the model, downgrade it to exclusive self-route (free, owned-only) instead of rejecting.

- `coordinator/api/self_route.go` — new `downgradePreferToSelfRoute(policy, model)` helper (fires only for `prefer`, only when `OwnedProviderSummary` reports an owned machine serving the model).
- `coordinator/api/consumer.go` — wired into both reservation sites (chat/completions/responses and completions/messages).

No UI change needed — the toggle already sends the right `prefer` intent.

## Why billing integrity holds

- The downgrade pins the request to the owner's machine with **no paid fallback**, so it can never settle on the paid fleet for free.
- If the machine drops between the check and dispatch, the existing exclusive self-route pre-flight returns a precise 503 rather than free public inference.
- A non-owner (or owner with no available machine) with zero balance still gets a 402.

## Tests

Added two regression tests in `coordinator/api/self_route_test.go`:
- `TestSelfRoute_PreferZeroBalanceOwnerServedFree` — zero-balance owner + own machine → 200, net free.
- `TestSelfRoute_PreferZeroBalanceNoOwnedMachineStill402` — zero balance + owns nothing → still 402.

Coordinator builds clean; `api`/`registry` suites and all `TestSelfRoute_*` pass; gofmt clean.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/431"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784582937&installation_id=133247490&pr_number=431&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F431&signature=b1ee86830c2920367320113c43214a03d2d606fd5698659519488e320816f71b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -1845,8 +1845,28 @@ func (s *Server) handleChatCompletions(w http.ResponseWriter, r *http.Request) {
 		consumerKey := consumerKeyFromContext(r.Context())
 		reservedMicroUSD = s.reservationCost(model, billingPromptTokens, requestedMaxTokens)
 		// Per-key spend cap (phase 1) — checked before the reservation so a
-		// capped key never debits the account ledger.
-		if msg, ok := s.checkKeySpendCap(r.Context(), reservedMicroUSD); !ok {
+		// capped key never debits the account ledger. The reservation only runs
+		// when the cap passes, so a capped key never reaches the ledger.
+		capMsg, capOK := s.checkKeySpendCap(r.Context(), reservedMicroUSD)
+		var err error
+		if capOK {
+			serviceReservation, err = s.reserveInitialBalance(consumerKey, model, reservedMicroUSD)
+		}
+		// A capped key OR an insufficient balance blocks the request. But a
+		// zero-balance / spend-capped owner whose own machine can serve THIS
+		// request must never be blocked from their own Mac: downgrade prefer →
+		// exclusive self-route (free, owned-only). A free self-route neither
+		// spends (cap moot) nor reserves balance, exactly like a self_route_only
+		// key — so the cap/balance gates do not apply to it. The downgrade is
+		// attempted before either rejection so the cap can't pre-empt it.
+		blockedByCap := !capOK
+		blockedByBalance := errors.Is(err, store.ErrInsufficientBalance)
+		switch {
+		case (blockedByCap || blockedByBalance) &&
+			s.downgradePreferToSelfRoute(&policy, model, registry.RequestTraits{HasTools: hasTools}, requiresVision, allowedProviderSerials):
+			reservedMicroUSD = 0
+			serviceReservation = false
+		case blockedByCap:
 			s.recordRejection(rejectionInfo{
 				r:                     r,
 				stage:                 "balance",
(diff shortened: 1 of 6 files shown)


ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:1798-1799` — Redundant model shed check and downgrade call in both handlers
  - _Suggestion:_ Extract the shed check and downgrade logic into a shared helper function to avoid code duplication between handleChatCompletions and handleGenericInference
- 🔵 **[INFO]** `coordinator/registry/scheduler.go:736-741` — Map allocation inside function without pre-sizing
  - _Suggestion:_ Pre-allocate the allowedSet map with known capacity: allowedSet := make(map[string]struct{}, len(allowedSerials))
- 🟡 **[MEDIUM]** `coordinator/registry/scheduler.go:743-795` — Holding read lock while iterating all providers and acquiring individual locks
  - _Suggestion:_ Consider reducing lock contention by collecting provider IDs first, releasing the read lock, then checking each provider individually, or use a more granular locking strategy

### Type_diligence — 2 finding(s) (2 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:1854` — Missing type assertion check for errors.Is(err, store.ErrInsufficientBalance)
  - _Suggestion:_ The code relies on errors.Is() without checking if err is nil first. While errors.Is() handles nil gracefully, the pattern suggests err could be nil when capOK is false, making the type assertion potentially misleading. Consider restructuring to make the error handling more explicit.
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:4713` — Missing type assertion check for errors.Is(err, store.ErrInsufficientBalance)
  - _Suggestion:_ Same issue as above - the code relies on errors.Is() without checking if err is nil first. While errors.Is() handles nil gracefully, the pattern suggests err could be nil when capOK is false, making the type assertion potentially misleading. Consider restructuring to make the error handling more explicit.

### Additive_complexity — 6 finding(s) (4 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:1798-1800` — Model shed logic duplicated in two handlers with identical implementation
  - _Suggestion:_ Extract the shed check and pinning logic into a shared helper method to eliminate duplication
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:4687-4689` — Identical model shed logic duplicated from handleChatCompletions
  - _Suggestion:_ Extract the shed check and pinning logic into a shared helper method to eliminate duplication
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:1851-1889` — Complex nested switch statement with multiple boolean conditions creates cognitive overhead
  - _Suggestion:_ Break down the switch into separate if statements or extract helper methods for each blocking condition
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:4710-4746` — Identical complex switch statement duplicated from handleChatCompletions
  - _Suggestion:_ Extract the billing check and downgrade logic into a shared helper method
- 🔵 **[INFO]** `coordinator/api/self_route.go:108-117` — pinPreferToSelfRouteOnShed duplicates downgradePreferToSelfRoute logic with only comment differences
  - _Suggestion:_ Merge into single method with a parameter to distinguish behavior, or extract common validation logic
- 🔵 **[INFO]** `coordinator/registry/scheduler.go:716-797` — OwnedProviderEligible method is overly complex with nested conditions and multiple responsibilities
  - _Suggestion:_ Break into smaller focused methods: checkOwnership, checkCapability, checkHardwareFit

_11 finding(s) total, 8 blocking. Verdict: REQUEST_CHANGES._

PR #440, opened 2026-06-21
fix(coordinator): DAR-349 remove provider_earnings dedupe from startup path

Gajesh2007:
Fixes **DAR-349** (Urgent). Resolves the deploy outage where the coordinator never bound `:8080`.

## Root cause
`NewPostgres → migrate()` ran an expensive boot-time dedupe on the hot `provider_earnings` table (~900k rows / ~443MB):

```sql
DO $$ BEGIN DELETE FROM provider_earnings
  WHERE id NOT IN (SELECT MIN(id) FROM provider_earnings WHERE job_id <> '' GROUP BY job_id)
  AND job_id <> ''; EXCEPTION WHEN others THEN NULL; END $$
```

On deploy it full-scanned and held a relation lock for ~15m, blocking the following `CREATE INDEX` statements, so the HTTP server never started. Production duplicate count is **0**, so it did no useful work — pure downside. The swallowed `EXCEPTION WHEN others THEN NULL` also hid the behavior.

## Fix
- **Remove the boot-time dedupe `DELETE` entirely.** Serving startup no longer runs data cleanup.
- **Build the partial unique index safely**, after the migration loop, in `ensureProviderEarningsJobIndex`:
  - **Fast path** — if a valid `idx_provider_earnings_job` already exists, return immediately (no-op on every boot after the first).
  - **Non-destructive guard** — counts duplicate `job_id` groups; if any exist, fail loudly with an actionable message pointing at the offline script instead of deleting rows.
  - **`CREATE UNIQUE INDEX CONCURRENTLY`** on a dedicated connection via the simple query protocol — no write-blocking lock, so a blue-green old coordinator still writing earnings is unaffected.
  - Drops a leftover *invalid* index (from an interrupted concurrent build) so `IF NOT EXISTS` can't silently skip a rebuild.
- The idempotent write path (`RecordProviderEarning`, `CreditProviderAccount` already use `ON CONFLICT (job_id) WHERE job_id <> '' DO NOTHING`) is now reliably backed by the index.
- **Offline dedupe** moved to `coordinator/store/migrations/dedupe_provider_earnings.sql` for the (currently non-existent) duplicate case — run out of band, never at boot.

## Tests
- `TestMigrate_NoBootTimeProviderEarningsDedupe` — always-on (no DB) guard against reintroducing the boot dedupe.
- `TestProviderEarningsJobIndex_BootSafe` — postgres-gated (`DATABASE_URL`): valid index after startup, `migrate()` re-entrancy, idempotent earnings writes, and empty `job_id` rows preserved (partial index).

Local: `go build ./...`, `go vet ./store/`, `gofmt` clean; store tests pass (postgres tests skip without `DATABASE_URL`, run in CI).

## Acceptance criteria
- [x] Coordinator binds `:8080` without running the dedupe query.
- [x] No long-running `provider_earnings` startup queries (one-time concurrent index build only when missing; no locks).
- [x] Base rewards don't require startup dedupe.
- [x] Duplicate prevention via index + idempotent writes, not recurring startup cleanup.

> Coordinator-only. Per repo policy, **prod GCP deploy is human-only** — this PR does not deploy.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/440"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784613299&installation_id=133247490&pr_number=440&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F440&signature=d90acf99877f51b45337985d181090f2c7957e4fb3301656e348a2fa0bd96ec1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/store/dar349_startup_test.go
+++ b/coordinator/store/dar349_startup_test.go
@@ -1,0 +1,106 @@
+package store
+
+import (
+	"bytes"
+	"context"
+	"os"
+	"testing"
+)
+
+// TestMigrate_NoBootTimeProviderEarningsDedupe is the always-on (no-DB) guard for
+// DAR-349: a boot-time `DELETE ... GROUP BY job_id` on the hot provider_earnings
+// table once held a relation lock for ~15m and stopped the coordinator from
+// binding :8080 (production outage). It must never return to the startup
+// migration path. Dedupe, if ever needed, is an offline job
+// (coordinator/store/migrations/dedupe_provider_earnings.sql).
+func TestMigrate_NoBootTimeProviderEarningsDedupe(t *testing.T) {
+	src, err := os.ReadFile("postgres.go")
+	if err != nil {
+		t.Fatalf("read postgres.go: %v", err)
+	}
+	for _, banned := range []string{
+		"DELETE FROM provider_earnings WHERE id NOT IN",
+		"GROUP BY job_id) AND job_id",
+	} {
+		if bytes.Contains(src, []byte(banned)) {
+			t.Fatalf("DAR-349 regression: boot-time provider_earnings dedupe reintroduced "+
+				"(found %q in postgres.go); move destructive cleanup to an offline job", banned)
+		}
+	}
+}
+
+// TestProviderEarningsJobIndex_BootSafe verifies the safe replacement: startup
+// builds a valid partial unique index on provider_earnings(job_id) without a
+// dedupe DELETE, migrate() is re-entrant, and the index backs the idempotent
+// ON CONFLICT write path. Runs only with DATABASE_URL (throwaway test DB).
+func TestProviderEarningsJobIndex_BootSafe(t *testing.T) {
+	s := testPostgresStore(t) // t.Skip()s when DATABASE_URL is unset
+	ctx := context.Background()
+
+	// NewPostgres -> migrate -> ensureProviderEarningsJobIndex left a VALID index.
+	if !jobIndexValid(t, s) {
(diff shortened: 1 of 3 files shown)


Gajesh2007:
**Deploy note (on-call):** the *first* green coordinator boot after this lands will pause before binding `:8080` for the one-time `CREATE UNIQUE INDEX CONCURRENTLY` build on `provider_earnings` (~tens of seconds on the ~900k-row prod table). This is expected and intended — the build holds **no** write-blocking lock, the blue-green old coordinator keeps serving throughout, and every subsequent boot hits the valid-index fast path and binds instantly. This replaces the old ~15-min exclusive-lock dedupe that caused the outage. Quality gate: both reviewers (Codex + independent) returned PASS, no blocking findings.

PR #445, opened 2026-06-22
refactor: split coordinator + provider god-files into focused modules (stacked on #444)

Gajesh2007:
## Summary

This PR finishes the god-file teardown started in #444: the remaining monolithic
orchestrators in the **coordinator (Go)** and **provider (Swift)** are split into
thin cores plus focused, single-responsibility concern files. **No behavior
changes** — every split moved code verbatim into a new home; the public entry
points are unchanged.

It is **stacked on #444** (`chore/tech-debt-modularize`), so this PR is targeted
at that branch and its diff shows only the god-file split work. It will be
**retargeted to `master` once #444 merges**.

This branch also carries the merge of the updated base (`da789aec`), which pulled
master's **#443** ("inference reliability — kill retry storm, eject zombies, shed
cleanly as 429") into the split layout. That reconciliation is included here and
is behavior-preserving (details in *Merge reconciliation* below).

## Code structure — before / after

```mermaid
flowchart LR
  subgraph Before["Before — god-files"]
    direction TB
    C0["consumer.go ~4.9k"]
    P0["provider.go ~3.6k"]
    R0["registry.go ~4.8k"]
    S0["store/interface.go ~1.5k"]
    PL0["ProviderLoop.swift 3577"]
    BS0["BatchScheduler.swift 2503"]
    MC0["ModelCatalog.swift 1202"]
    SC0["StartCommand.swift 1133"]
    CC0["CoordinatorClient.swift 1173"]
  end
  subgraph After["After — thin cores + concern files"]
    direction TB
    C1["consumer.go core 3815<br/>+ httputil / inference_admission<br/>+ models_endpoints / responses_translate"]
    P1["provider.go core 3149<br/>+ provider_codeattest"]
    R1["registry.go core 4396<br/>+ persistence / routing_eligibility<br/>+ warm_pool_controller + characterization test"]
    S1["store/interface.go core 944<br/>+ interface_domains: Store -> 12 sub-interfaces"]
    PL1["ProviderLoop.swift 525<br/>+ 14 ProviderLoop+*.swift"]
    BS1["BatchScheduler.swift 538<br/>+ Admission/Submit/ModelLifecycle/...<br/>+ EngineFactory/CheckpointRestore/..."]
    MC1["ModelCatalog.swift 171<br/>+ ModelCatalogClient + ModelDownloader+*"]
    SC1["StartCommand.swift 121<br/>+ Modes/Picker/Preflight/Daemon/TUIPicker"]
[...]

PR #447, opened 2026-06-22
TTFT contention Phase 0: telemetry fix + occupancy-aware shadow gate (behavior-neutral)

Gajesh2007:
## Summary

Phase 0 (coordinator-only, **behavior-neutral by default**) of the TTFT-contention fix. Grounded in read-only prod telemetry analysis (see `docs/reports/2026-06-22-*`).

- **Fix the broken TTFT telemetry.** `actual_ttft_ms` now derives from `FirstContentAt` vs the *committed* attempt's `DispatchedAt` (retry-proof; the old `FirstChunkAt − DispatchedAt` went as low as −378s on retries). Negatives clamp to 0 and emit `routing.invalid_ttft`. `completion_tokens` is now persisted on cancelled/error/timeout rows (was NULL) in **both** stores.
- **Occupancy-aware TTFT estimate**, reusing the *existing* `max(pendingForModel, backend_running+backend_waiting)` occupancy (extracted into `snapshotOccupancy`; **no new reservation counter**). Gated behind `EIGENINFERENCE_TTFT_OCCUPANCY_ALPHA` (default `0` ⇒ estimate/cost/ceiling byte-for-byte unchanged).
- **Shadow admission + spread measurement** at the data-verified ~10s deadline: emits `routing.ttft_admission{would_shed|would_serve}` and `routing.ttft_spread{would_redirect_to_idle}` **without changing any routing decision**. Gated behind `EIGENINFERENCE_TTFT_ADMISSION_MODE` (default `off`); `EIGENINFERENCE_TTFT_DEADLINE_BASE_MS` (default `10000`) is consumed only by the shadow evaluator.

Why: prod data showed cancels fit a ~10s deadline (not the code's 5s), token budget was non-binding (0.4% util at cancel), and ~54%/65% of `backend_running=0` gpt-oss cancels had hidden in-flight peers while ~44 loaded-idle providers sat unused — a load-spreading failure the coordinator already has the state to fix.

## Before / After

```mermaid
flowchart TB
  subgraph Before
    A1[request] --> A2["admit onto provider reporting backend_running=0<br/>(stale gauge: hidden in-flight peers)"]
    A2 --> A3["~11s, 0 tokens delivered"]
    A3 --> A4["client_gone cancel"]
    A2 -.telemetry.-> A5["actual_ttft_ms = FirstChunkAt − DispatchedAt<br/>retry ⇒ negative (−378s); completion_tokens NULL on cancel"]
  end
  subgraph After["After (defaults alpha=0, mode=off ⇒ behavior-neutral)"]
    B1[request] --> B2["estimate reads EXISTING occupancy<br/>max(pendingForModel, backend_running+waiting) [alpha-gated]"]
    B2 --> B3{"TTFT_ADMISSION_MODE"}
    B3 -->|off default| B4["serve exactly as today"]
    B3 -->|shadow| B5["emit would_shed / would_redirect_to_idle<br/>(no decision change)"]
    B2 -.telemetry.-> B6["actual_ttft_ms = FirstContentAt − committed DispatchedAt<br/>clamp >= 0 + routing.invalid_ttft; completion_tokens persisted (0 on cancel)"]
  end
```

Code paths changed: `registry/scheduler.go` (`snapshotOccupancy`, `ttftOccupancyMs`, shadow wiring in `ReserveProviderEx`), new `registry/ttft_shadow.go`, `registry/registry.go` (`MarkFirstContentArrived`/`FirstContentAtSafe`), `api/route_outcome.go` + `api/provider.go` + `api/dispatch.go` (telemetry), new `api/ttft_shadow_metrics.go`, `store/{interface,postgres,route_telemetry}.go` (`CompletionTokensSet`), `cmd/coordinator/main.go` (env wiring).

## Test plan

- [x] `go test ./...` green; `go test -race ./registry/... ./api/...` green; `gofmt -l` empty; `go vet` clean (re-verified after rebase onto #445).
- [x] New tests: occupancy term == 0 at alpha=0; alpha>0 estimate monotonic in occupancy and crosses the 10s knee; shadow `would_shed` yet still serves; retried-request `actual_ttft_ms` clamped to 0 + metered; `completion_tokens`=0 persists on a cancelled row in memory **and** postgres impls.
- [x] Behavior-neutral by default: with `alpha=0` + `mode=off`, no routing or telemetry-decision change; prod `EIGENINFERENCE_TTFT_HARD_REJECT` path untouched.

## Rollout

Land (neutral) → set `TTFT_ADMISSION_MODE=shadow` in dev to collect `would_shed`/`would_redirect_to_idle` → **enforce step (separate PR):** wire the actual rejection/redirect, move the live `consumer.go` base 5s→10s, and stage `EIGENINFERENCE_QUALITY_CONCURRENCY_OVERCOMMIT` 2.0→1.0.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/447"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784739632&installation_id=133247490&pr_number=447&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F447&signature=e406ef8a7f764c72bb2972f6d4d8501c5fb8b9bd624728a96b729c999720720a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/api/dispatch.go
+++ b/coordinator/api/dispatch.go
@@ -388,6 +388,10 @@ func (d *dispatchState) commitFirstContent(pr *registry.PendingRequest, chunk st
 	d.firstChunk = chunk
 	pr.MarkFirstChunkArrived()
 	pr.MarkFirstContentArrived()
+	// Mark THIS attempt as the committed one so handleComplete's fallback only
+	// ever stamps FirstContentAt for the attempt that actually delivered content —
+	// never a late-completing abandoned/retried attempt sharing the same Timing.
+	pr.MarkContentCommitted()
 }
 
 // successRoutingOutcome builds a success outcome for the committed attempt.
--- a/coordinator/api/dispatch.go
+++ b/coordinator/api/dispatch.go
@@ -320,6 +320,11 @@ func (d *dispatchState) recordRoutingDecisionFor(provider *registry.Provider, pr
 		provider.Mu().Unlock()
 	}
 
+	// Phase-0 shadow TTFT admission/spread metrics. No-op unless the request was
+	// evaluated (admission mode != off AND a provider was selected). Emitted on
+	// the synchronous path (cheap counter incr), not inside the async store write.
+	s.emitTTFTShadowMetrics(d.model, decision)
+
 	s.submitTelemetry("recordInferenceRoute", func() {
 		if err := s.store.RecordInferenceRoute(record); err != nil && s.logger != nil {
 			s.logger.Error("inference_routes record write failed",
@@ -2296,9 +2301,10 @@ func (d *dispatchState) writeCommittedResponse() {
 	// (known up front, adequate for normalization) and the provider's benchmarked
 	// PrefillTPS (set once at registration, read-only thereafter).
 	if pr.Timing != nil && d.firstChunk != "" {
-		if pr.Timing.FirstContentAt.IsZero() {
-			pr.Timing.FirstContentAt = time.Now()
-		}
+		// Stamp under timingMu (MarkFirstContentArrived) because the route-
+		// telemetry path now reads FirstContentAt for actual_ttft_ms from the
+		// provider read-loop goroutine (handleComplete -> applyPendingRouteTelemetry).
+		pr.MarkFirstContentArrived()
 		sample := adjustLatencyForPrefill(contentLatency(pr.Timing), pr.EstimatedPromptTokens, provider.PrefillTPS)
 		s.registry.RecordLatency(provider.ID, sample)
 	}
--- a/coordinator/api/dispatch.go
+++ b/coordinator/api/dispatch.go
@@ -372,6 +372,24 @@ func applyTimingDecomposition(out *store.InferenceRouteOutcome, t *registry.Requ
 	out.DispatchMs = timingMsBetween(t.DispatchedAt, firstChunk)
 }
 
+// commitFirstContent records the first CONTENT chunk on the committed attempt and
+// stamps FirstContentAt (the actual_ttft_ms anchor) in the SAME instant, on the
+// dispatch goroutine that reads the chunk. Stamping HERE — rather than later in
+// writeCommittedResponse — guarantees FirstContentAt is set before ANY route
+// outcome is built for this attempt: the committed/success outcome written by
+// this goroutine (e.g. waitFirstChunk / waitAccepted's defer) AND the terminal
+// completeRouteOutcome written concurrently by handleComplete on the provider
+// read-loop. Without it a fast single-chunk completion could persist
+// actual_ttft_ms as 0/NULL (applyPendingRouteTelemetry derives it solely from
+// FirstContentAt). pr is the COMMITTED attempt — the backup on a speculative
+// backup win, the primary otherwise. MarkFirstChunkArrived is kept (idempotent:
+// it preserves an earlier preamble's first-byte time for dispatch_to_first_chunk_ms).
+func (d *dispatchState) commitFirstContent(pr *registry.PendingRequest, chunk string) {
+	d.firstChunk = chunk
+	pr.MarkFirstChunkArrived()
+	pr.MarkFirstContentArrived()
+}
+
 // successRoutingOutcome builds a success outcome for the committed attempt.
 // Token counts and final_status are left empty because the final terminal is
 // only known when the provider later sends complete/error; handleComplete or
@@ -1060,8 +1078,7 @@ func (d *dispatchState) waitFirstChunk() (outcome dispatchOutcome) {
 			speculativeTimer.Stop()
 			deadlineTimer.Stop()
 			if ok {
-				d.firstChunk = chunk
-				pr.MarkFirstChunkArrived()
+				d.commitFirstContent(pr, chunk)
 				d.committed = true
 			} else {
 				select {
@@ -1276,8 +1293,7 @@ func (d *dispatchState) waitNoBackup() dispatchOutcome {
 			}
 			remainingDeadline.Stop()
 			if ok {
-				d.firstChunk = chunk
-				pr.MarkFirstChunkArrived()
+				d.commitFirstContent(pr, chunk)
 				d.committed = true
 			} else {
 				select {
@@ -1389,8 +1405,7 @@ func (d *dispatchState) runRace(backupProvider *registry.Provider, backupPR *reg
 			s.cancelDispatch(backupProvider, backupPR)
 			if ok {
 				d.markSpeculativeLoser(backupPR)
-				d.firstChunk = chunk
-				pr.MarkFirstChunkArrived()
+				d.commitFirstContent(pr, chunk)
 				d.committed = true
 			} else {
 				select {
@@ -1431,8 +1446,7 @@ func (d *dispatchState) runRace(backupProvider *registry.Provider, backupPR *reg
 				d.pr = backupPR
 				d.requestID = d.pr.RequestID
 				d.heldChunks = backupHeld
-				d.firstChunk = chunk
-				d.pr.MarkFirstChunkArrived()
+				d.commitFirstContent(d.pr, chunk)
 				d.committed = true
 			} else {
 				select {
@@ -1578,8 +1592,7 @@ func (d *dispatchState) raceBackupChunkClosedWaitPrimary(provider *registry.Prov
 			}
 			remainingPrimary.Stop()
 			if ok {
-				d.firstChunk = chunk
-				pr.MarkFirstChunkArrived()
+				d.commitFirstContent(pr, chunk)
 				d.committed = true
 			} else {
 				select {
@@ -1676,8 +1689,7 @@ func (d *dispatchState) racePrimaryFailedWaitBackup(backupProvider *registry.Pro
 				d.pr = backupPR
 				d.requestID = d.pr.RequestID
 				d.heldChunks = backupHeld
-				d.firstChunk = chunk
-				d.pr.MarkFirstChunkArrived()
+				d.commitFirstContent(d.pr, chunk)
 				d.committed = true
 			} else {
 				select {
@@ -1773,8 +1785,7 @@ func (d *dispatchState) raceBackupErrWaitPrimary(provider *registry.Provider, pr
 			}
 			primaryDeadline.Stop()
 			if ok {
-				d.firstChunk = chunk
-				pr.MarkFirstChunkArrived()
+				d.commitFirstContent(pr, chunk)
 				d.committed = true
 			} else {
 				select {
@@ -1887,8 +1898,7 @@ func (d *dispatchState) waitAccepted() (outcome dispatchOutcome) {
 			}
 			chunkTimer.Stop()
 			if ok {
-				d.firstChunk = chunk
-				pr.MarkFirstChunkArrived()
+				d.commitFirstContent(pr, chunk)
 				d.committed = true
 			} else {
 				// Closed — check for error. Use a short grace
@@ -2301,10 +2311,10 @@ func (d *dispatchState) writeCommittedResponse() {
 	// (known up front, adequate for normalization) and the provider's benchmarked
 	// PrefillTPS (set once at registration, read-only thereafter).
 	if pr.Timing != nil && d.firstChunk != "" {
-		// Stamp under timingMu (MarkFirstContentArrived) because the route-
-		// telemetry path now reads FirstContentAt for actual_ttft_ms from the
-		// provider read-loop goroutine (handleComplete -> applyPendingRouteTelemetry).
-		pr.MarkFirstContentArrived()
+		// FirstContentAt was already stamped at the content-commit site
+		// (commitFirstContent), earlier in THIS goroutine, so contentLatency reads
+		// a set value here. No re-stamp needed; just read it for the reputation
+		// latency sample.
 		sample := adjustLatencyForPrefill(contentLatency(pr.Timing), pr.EstimatedPromptTokens, provider.PrefillTPS)
 		s.registry.RecordLatency(provider.ID, sample)
 	}
--- a/coordinator/api/route_outcome.go
+++ b/coordinator/api/route_outcome.go
@@ -47,7 +47,17 @@ var validInferenceErrorReasons = map[string]struct{}{
 }
 
 func (s *Server) updateInferenceRouteOutcomeWithModel(requestID string, attempt int, model string, outcome *store.InferenceRouteOutcome) {
-	if s == nil || s.store == nil || requestID == "" || outcome == nil {
+	if s == nil || outcome == nil {
+		return
+	}
+	// Loud guard: a negative raw TTFT was clamped to 0 (see
+	// applyPendingRouteTelemetry). Emitting here — the single store-submit funnel
+	// every terminal/commit outcome flows through — makes any regression of the
+	// retried-request shared-Timing bug visible instead of silent.
+	if outcome.InvalidTTFT {
+		s.emitInvalidTTFT(model, "negative")
+	}
+	if s.store == nil || requestID == "" {
 		return
 	}
 	s.emitInferenceErrorMetric(model, outcome)
@@ -97,6 +107,24 @@ func routeOutcomeWithReason(status, class string, code int, providerReason, erro
 		ErrorCode:   code,
 		ErrorClass:  class,
 		ErrorReason: inferenceErrorReason(providerReason, status, class, code, errorText),
+		// Terminal cancel/error/timeout rows deliver 0 tokens; force-persist that
+		// 0 (instead of leaving completion_tokens NULL) so the incident-majority
+		// 0-token cancels are visible. Success writes its real count separately
+		// via completeRouteOutcome.
+		CompletionTokensSet: terminalForcesCompletionTokens(status),
+	}
+}
+
+// terminalForcesCompletionTokens reports whether a terminal final_status must
+// persist completion_tokens even when it is 0. Cancel/error/timeout rows deliver
+// zero tokens and must record 0 (not NULL) so the 0-token cancel population is
+// queryable; partial_success and success are handled by their own count writers.
+func terminalForcesCompletionTokens(status string) bool {
+	switch strings.ToLower(strings.TrimSpace(status)) {
+	case "cancelled", "error", "timeout":
+		return true
+	default:
+		return false
 	}
 }
 
@@ -207,8 +235,11 @@ func completeRouteOutcome(pr *registry.PendingRequest, usage protocol.UsageInfo,
 		ErrorClass:       errorClass,
 		PromptTokens:     usage.PromptTokens,
 		CompletionTokens: usage.CompletionTokens,
-		ReasoningTokens:  usage.ReasoningTokens,
-		CostMicroUSD:     costMicroUSD,
+		// Authoritative provider-reported count; persist it even when 0 so a
+		// 0-token success records 0 rather than NULL.
+		CompletionTokensSet: true,
+		ReasoningTokens:     usage.ReasoningTokens,
+		CostMicroUSD:        costMicroUSD,
 	}
 	if errorClass != "" {
 		out.ErrorReason = inferenceErrorReason("", status, errorClass, 0, "")
@@ -277,10 +308,34 @@ func applyPendingRouteTelemetry(out *store.InferenceRouteOutcome, pr *registry.P
 	}
 	t := pr.Timing
 	firstChunk := pr.FirstChunkAtSafe()
-	if !firstChunk.IsZero() && !t.DispatchedAt.IsZero() {
-		ms := float64(firstChunk.Sub(t.DispatchedAt).Milliseconds())
+	// actual_ttft_ms is time-to-first-DELIVERED-content (FirstContentAt) measured
+	// against the COMMITTED attempt's DispatchedAt. FirstContentAt is stamped only
+	// on the committed attempt and DispatchedAt is that same attempt's dispatch,
+	// so the two cannot come from different attempts — eliminating the
+	// retried-request shared-Timing bug (FirstChunkAt of an early attempt minus a
+	// later attempt's overwritten DispatchedAt) that produced the -378s rows.
+	// Held role-only / lifecycle preamble (FirstChunkAt) is deliberately NOT used
+	// here, so a fast-preamble-then-stall provider cannot look responsive. A
+	// non-committed terminal (cancel/error: no content) leaves FirstContentAt zero
+	// => actual_ttft_ms 0 (correct: zero tokens delivered).
+	firstContent := pr.FirstContentAtSafe()
+	if !firstContent.IsZero() && !t.DispatchedAt.IsZero() {
+		ms := float64(firstContent.Sub(t.DispatchedAt).Milliseconds())
+		if ms < 0 {
+			// Should be impossible (same attempt), but clamp + flag so any
+			// regression is loud (routing.invalid_ttft) rather than a poison -ms row.
+			out.InvalidTTFT = true
+			ms = 0
+		}
 		out.ActualTTFTMs = ms
-		out.DispatchToFirstChunkMs = ms
+	}
+	// dispatch_to_first_chunk_ms stays the held-preamble (first-byte) diagnostic.
+	// Clamp negatives so a stale-pointer regression cannot write a -ms value here
+	// either.
+	if !firstChunk.IsZero() && !t.DispatchedAt.IsZero() {
+		if ms := float64(firstChunk.Sub(t.DispatchedAt).Milliseconds()); ms >= 0 {
+			out.DispatchToFirstChunkMs = ms
+		}
 	}
 	if !t.ReceivedAt.IsZero() {
 		out.TotalDurationMs = float64(time.Since(t.ReceivedAt).Milliseconds())
--- a/coordinator/api/route_outcome.go
+++ b/coordinator/api/route_outcome.go
@@ -32,6 +32,17 @@ const (
 // every provider — so it is NOT a provider fault and NOT an admission mismatch.
 const errorClassClientError = "client_error"
 
+// Final-status values persisted on inference_routes (store.InferenceRouteOutcome
+// .FinalStatus). Centralized so status comparisons/constructions don't drift on a
+// bare string literal.
+const (
+	finalStatusSuccess        = "success"
+	finalStatusPartialSuccess = "partial_success"
+	finalStatusError          = "error"
+	finalStatusCancelled      = "cancelled"
+	finalStatusTimeout        = "timeout"
+)
+
 var validInferenceErrorReasons = map[string]struct{}{
 	errorReasonJinjaChannelTags:   {},
 	errorReasonJinjaNullBridge:    {},
@@ -77,7 +88,7 @@ func (s *Server) updateInferenceRouteOutcomeWithModel(requestID string, attempt 
 }
 
 func (s *Server) emitInferenceErrorMetric(model string, outcome *store.InferenceRouteOutcome) {
-	if s == nil || outcome == nil || outcome.ErrorReason == "" || outcome.FinalStatus == "" || outcome.FinalStatus == "success" {
+	if s == nil || outcome == nil || outcome.ErrorReason == "" || outcome.FinalStatus == "" || outcome.FinalStatus == finalStatusSuccess {
 		return
 	}
 	tags := []string{"reason:" + outcome.ErrorReason}
@@ -121,7 +132,7 @@ func routeOutcomeWithReason(status, class string, code int, providerReason, erro
 // queryable; partial_success and success are handled by their own count writers.
 func terminalForcesCompletionTokens(status string) bool {
 	switch strings.ToLower(strings.TrimSpace(status)) {
-	case "cancelled", "error", "timeout":
+	case finalStatusCancelled, finalStatusError, finalStatusTimeout:
 		return true
 	default:
 		return false
@@ -157,7 +168,7 @@ func providerFailedPendingRouteOutcomeWithReason(pr *registry.PendingRequest, st
 }
 
 func dispatchFailedPendingRouteOutcome(pr *registry.PendingRequest, class string, code int) *store.InferenceRouteOutcome {
-	return pendingRouteOutcome(pr, "error", class, code)
+	return pendingRouteOutcome(pr, finalStatusError, class, code)
 }
 
 func providerDisconnectedError(errorText string, statusCode int) bool {
@@ -169,7 +180,7 @@ func postCommitProviderErrorOutcome(pr *registry.PendingRequest, msg protocol.In
 	if providerDisconnectedError(msg.Error, msg.StatusCode) {
 		class = "provider_disconnect_after_commit"
 	}
-	return providerFailedPendingRouteOutcomeWithReason(pr, "partial_success", class, msg.StatusCode, msg.ErrorReason, msg.Error)
+	return providerFailedPendingRouteOutcomeWithReason(pr, finalStatusPartialSuccess, class, msg.StatusCode, msg.ErrorReason, msg.Error)
 }
 
 func preResponseProviderErrorOutcome(pr *registry.PendingRequest, msg protocol.InferenceErrorMessage) *store.InferenceRouteOutcome {
@@ -177,7 +188,7 @@ func preResponseProviderErrorOutcome(pr *registry.PendingRequest, msg protocol.I
 	if providerDisconnectedError(msg.Error, msg.StatusCode) {
 		class = "provider_disconnect_before_response"
 	}
-	return providerFailedPendingRouteOutcomeWithReason(pr, "error", class, msg.StatusCode, msg.ErrorReason, msg.Error)
+	return providerFailedPendingRouteOutcomeWithReason(pr, finalStatusError, class, msg.StatusCode, msg.ErrorReason, msg.Error)
 }
 
 func preCommitProviderErrorOutcome(pr *registry.PendingRequest, msg protocol.InferenceErrorMessage) *store.InferenceRouteOutcome {
@@ -186,48 +197,48 @@ func preCommitProviderErrorOutcome(pr *registry.PendingRequest, msg protocol.Inf
 		// by shape (fails identically on every provider), not a provider fault.
 		// Record as client_error WITHOUT AdmittedButFailed so it never pollutes the
 		// admission-mismatch gauge.
-		return pendingRouteOutcomeWithReason(pr, "error", errorClassClientError, msg.StatusCode, msg.ErrorReason, msg.Error)
+		return pendingRouteOutcomeWithReason(pr, finalStatusError, errorClassClientError, msg.StatusCode, msg.ErrorReason, msg.Error)
 	}
 	class := "provider_error"
 	if providerDisconnectedError(msg.Error, msg.StatusCode) {
 		class = "provider_disconnect_pre_commit"
 	}
-	return providerFailedPendingRouteOutcomeWithReason(pr, "error", class, msg.StatusCode, msg.ErrorReason, msg.Error)
+	return providerFailedPendingRouteOutcomeWithReason(pr, finalStatusError, class, msg.StatusCode, msg.ErrorReason, msg.Error)
 }
 
 func postCommitProviderIncompleteOutcome(pr *registry.PendingRequest) *store.InferenceRouteOutcome {
-	return providerFailedPendingRouteOutcome(pr, "partial_success", "provider_incomplete_after_commit", 502)
+	return providerFailedPendingRouteOutcome(pr, finalStatusPartialSuccess, "provider_incomplete_after_commit", 502)
 }
 
 func preResponseProviderIncompleteOutcome(pr *registry.PendingRequest) *store.InferenceRouteOutcome {
-	return providerFailedPendingRouteOutcome(pr, "error", "provider_incomplete_before_response", 502)
+	return providerFailedPendingRouteOutcome(pr, finalStatusError, "provider_incomplete_before_response", 502)
 }
 
 func postCommitStreamTimeoutOutcome(pr *registry.PendingRequest) *store.InferenceRouteOutcome {
-	return pendingRouteOutcome(pr, "partial_success", "stream_timeout_after_commit", 504)
+	return pendingRouteOutcome(pr, finalStatusPartialSuccess, "stream_timeout_after_commit", 504)
 }
 
 func preResponseTimeoutOutcome(pr *registry.PendingRequest, class string) *store.InferenceRouteOutcome {
-	return pendingRouteOutcome(pr, "timeout", class, 504)
+	return pendingRouteOutcome(pr, finalStatusTimeout, class, 504)
 }
 
 func noTerminalAfterCancelOutcome(pr *registry.PendingRequest) *store.InferenceRouteOutcome {
-	return pendingRouteOutcome(pr, "partial_success", "no_terminal_after_cancel", 504)
+	return pendingRouteOutcome(pr, finalStatusPartialSuccess, "no_terminal_after_cancel", 504)
 }
 
 func speculativeLoserOutcome(pr *registry.PendingRequest) *store.InferenceRouteOutcome {
-	return pendingRouteOutcome(pr, "cancelled", "speculative_loser", 0)
+	return pendingRouteOutcome(pr, finalStatusCancelled, "speculative_loser", 0)
 }
 
 func clientGoneBeforeResponseOutcome(pr *registry.PendingRequest) *store.InferenceRouteOutcome {
-	return pendingRouteOutcome(pr, "cancelled", "client_gone_before_response", 0)
+	return pendingRouteOutcome(pr, finalStatusCancelled, "client_gone_before_response", 0)
 }
 
 func completeRouteOutcome(pr *registry.PendingRequest, usage protocol.UsageInfo, costMicroUSD int64, consumerGone bool) *store.InferenceRouteOutcome {
-	status := "success"
+	status := finalStatusSuccess
 	errorClass := ""
 	if consumerGone {
-		status = "partial_success"
+		status = finalStatusPartialSuccess
 		errorClass = errorClassClientGoneAfterCommitCompleted
 	}
 	out := &store.InferenceRouteOutcome{
@@ -259,7 +270,7 @@ func inferenceErrorReason(providerReason, status, class string, code int, messag
 	if status == "" && class == "" && code == 0 && message == "" {
 		return ""
 	}
-	if strings.EqualFold(strings.TrimSpace(status), "success") {
+	if strings.EqualFold(strings.TrimSpace(status), finalStatusSuccess) {
 		return ""
 	}
 
--- a/coordinator/api/ttft_first_content_test.go
+++ b/coordinator/api/ttft_first_content_test.go
@@ -17,10 +17,12 @@ import (
 // completeRouteOutcome would then persist actual_ttft_ms as 0/NULL even though
 // tokens were delivered.
 //
-// This drives handleComplete with FirstContentAt deliberately unstamped (the
-// race) and a positive completion-token count. handleComplete's fallback stamps
-// FirstContentAt because the provider reported delivered tokens, so the persisted
-// actual_ttft_ms is > 0. Without the fix this test fails (ActualTTFTMs == 0).
+// This drives handleComplete for the COMMITTED attempt (MarkContentCommitted set,
+// as commitFirstContent does in the dispatch path) with FirstContentAt
+// deliberately unstamped (the race) and a positive completion-token count.
+// handleComplete's committed-attempt fallback stamps FirstContentAt, so the
+// persisted actual_ttft_ms is > 0. Without the fix this test fails
+// (ActualTTFTMs == 0).
 func TestFastCompletionStampsActualTTFT(t *testing.T) {
 	srv, st, _ := billingTestServer(t)
 
@@ -46,6 +48,9 @@ func TestFastCompletionStampsActualTTFT(t *testing.T) {
 			// FirstChunkAt / FirstContentAt intentionally unset (zero).
 		},
 	}
+	// This IS the committed attempt (the dispatch path marks it via
+	// commitFirstContent); the fallback is scoped to committed attempts.
+	pr.MarkContentCommitted()
 	provider.AddPending(pr)
 
 	// Pre-create the route row so the (best-effort, async) outcome update lands.
@@ -81,5 +86,77 @@ func TestFastCompletionStampsActualTTFT(t *testing.T) {
 	}
 	if rec.ActualTTFTMs <= 0 {
 		t.Fatalf("actual_ttft_ms must be populated and > 0 for a delivered completion, got %f (0/NULL = the FirstContentAt stamp-ordering bug)", rec.ActualTTFTMs)
+	}
+}
+
+// TestAbandonedAttemptDoesNotStampCommittedTTFT is the regression for the
+// committed-attempt fallback guard (Codex P2 round 3): an abandoned/retried
+// attempt that completes LATE (the provider finished after the coordinator timed
+// it out and retried) must NOT stamp FirstContentAt on the Timing it SHARES with
+// the committed retry. FirstContentAt is first-write-wins, so a stale stamp from
+// the abandoned attempt would clamp/zero the committed retry's actual_ttft_ms.
+// The fallback is gated on ContentCommittedSafe, and the abandoned attempt was
+// never committed, so it must leave FirstContentAt untouched — letting the
+// committed retry's first-content time win.
+func TestAbandonedAttemptDoesNotStampCommittedTTFT(t *testing.T) {
+	srv, st, _ := billingTestServer(t)
+
+	model := "abandoned-ttft-model"
+	provider := srv.registry.Register("abandoned-ttft-provider", nil, &protocol.RegisterMessage{
+		Models: []protocol.ModelInfo{{ID: model, ModelType: "chat", Quantization: "4bit"}},
+	})
+
+	// Timing is SHARED across attempts (dispatch.go sets Timing: d.timing on every
+	// attempt). DispatchedAt reflects the committed retry's dispatch.
+	shared := &registry.RequestTiming{
+		ReceivedAt:   time.Now().Add(-time.Second),
+		DispatchedAt: time.Now().Add(-200 * time.Millisecond),
+		// FirstContentAt unset: the committed retry has not delivered content yet.
+	}
+	// The abandoned attempt (attempt 0) — never committed (no MarkContentCommitted).
+	abandoned := &registry.PendingRequest{
+		RequestID:        "abandoned-attempt",
+		Attempt:          0,
+		Model:            model,
+		ConsumerKey:      testConsumerID,
+		ReservedMicroUSD: 10_000_000,
+		ChunkCh:          make(chan string, 1),
+		CompleteCh:       make(chan protocol.UsageInfo, 1),
+		ErrorCh:          make(chan protocol.InferenceErrorMessage, 1),
+		Timing:           shared,
+	}
+	provider.AddPending(abandoned)
+	if err := st.RecordInferenceRoute(&store.InferenceRouteRecord{
+		RequestID: abandoned.RequestID, Attempt: abandoned.Attempt, Model: model, ProviderID: provider.ID,
+	}); err != nil {
+		t.Fatalf("record route: %v", err)
+	}
+
+	// The abandoned attempt completes late, with delivered tokens. The fallback
+	// must NOT stamp the shared FirstContentAt (it was never the committed attempt).
+	srv.handleComplete(provider.ID, provider, &protocol.InferenceCompleteMessage{
+		Type:      protocol.TypeInferenceComplete,
+		RequestID: abandoned.RequestID,
+		Usage:     protocol.UsageInfo{PromptTokens: 1000, CompletionTokens: 5},
+	})
+
+	if !shared.FirstContentAt.IsZero() {
+		t.Fatalf("abandoned attempt must NOT stamp the shared FirstContentAt, got %v (would clamp the committed retry's actual_ttft_ms)", shared.FirstContentAt)
+	}
+
+	// The committed retry then delivers its first content — its stamp wins, and
+	// its actual_ttft_ms is measured from the shared DispatchedAt (positive).
+	committed := &registry.PendingRequest{RequestID: "committed-retry", Attempt: 1, Model: model, Timing: shared}
+	committed.MarkContentCommitted()
+	committed.MarkFirstContentArrived()
+	if shared.FirstContentAt.IsZero() {
+		t.Fatal("committed retry's first-content stamp must win")
+	}
+	out := completeRouteOutcome(committed, protocol.UsageInfo{PromptTokens: 1000, CompletionTokens: 5}, 0, false)
+	if out.ActualTTFTMs <= 0 {
+		t.Fatalf("committed retry actual_ttft_ms must be > 0 (its own first-content time), got %f", out.ActualTTFTMs)
+	}
+	if out.InvalidTTFT {
+		t.Fatal("committed retry actual_ttft_ms must be a clean positive, not the negative-clamp guard")
 	}
 }
--- a/coordinator/api/ttft_first_content_test.go
+++ b/coordinator/api/ttft_first_content_test.go
@@ -1,0 +1,85 @@
+package api
+
+import (
+	"testing"
+	"time"
+
+	"github.com/eigeninference/d-inference/coordinator/protocol"
+	"github.com/eigeninference/d-inference/coordinator/registry"
+	"github.com/eigeninference/d-inference/coordinator/store"
+)
+
+// TestFastCompletionStampsActualTTFT is the regression for the FirstContentAt
+// stamp-ordering bug (Codex P2, dispatch.go ~2301): a fast single-chunk
+// completion can have its TypeInferenceComplete reach handleComplete BEFORE the
+// dispatch goroutine stamps FirstContentAt. Since applyPendingRouteTelemetry now
+// derives actual_ttft_ms SOLELY from FirstContentAt, the terminal
+// completeRouteOutcome would then persist actual_ttft_ms as 0/NULL even though
+// tokens were delivered.
+//
+// This drives handleComplete with FirstContentAt deliberately unstamped (the
+// race) and a positive completion-token count. handleComplete's fallback stamps
+// FirstContentAt because the provider reported delivered tokens, so the persisted
+// actual_ttft_ms is > 0. Without the fix this test fails (ActualTTFTMs == 0).
+func TestFastCompletionStampsActualTTFT(t *testing.T) {
+	srv, st, _ := billingTestServer(t)
+
+	model := "fast-ttft-model"
+	provider := srv.registry.Register("fast-ttft-provider", nil, &protocol.RegisterMessage{
+		Models: []protocol.ModelInfo{{ID: model, ModelType: "chat", Quantization: "4bit"}},
+	})
+
+	// Committed attempt whose dispatch happened 50ms ago, but whose first content
+	// chunk the dispatch goroutine has NOT yet stamped (FirstContentAt zero) —
+	// exactly the fast-completion race the fix closes.
+	pr := &registry.PendingRequest{
+		RequestID:        "fast-ttft-req",
+		Model:            model,
+		ConsumerKey:      testConsumerID,
+		ReservedMicroUSD: 10_000_000,
+		ChunkCh:          make(chan string, 1),
+		CompleteCh:       make(chan protocol.UsageInfo, 1),
+		ErrorCh:          make(chan protocol.InferenceErrorMessage, 1),
+		Timing: &registry.RequestTiming{
+			ReceivedAt:   time.Now().Add(-60 * time.Millisecond),
+			DispatchedAt: time.Now().Add(-50 * time.Millisecond),
+			// FirstChunkAt / FirstContentAt intentionally unset (zero).
+		},
+	}
+	provider.AddPending(pr)
+
+	// Pre-create the route row so the (best-effort, async) outcome update lands.
+	if err := st.RecordInferenceRoute(&store.InferenceRouteRecord{
+		RequestID:  pr.RequestID,
+		Attempt:    pr.Attempt,
+		Model:      model,
+		ProviderID: provider.ID,
+	}); err != nil {
+		t.Fatalf("record route: %v", err)
+	}
+
+	srv.handleComplete(provider.ID, provider, &protocol.InferenceCompleteMessage{
+		Type:      protocol.TypeInferenceComplete,
+		RequestID: pr.RequestID,
+		Usage:     protocol.UsageInfo{PromptTokens: 1000, CompletionTokens: 1},
+	})
+
+	// The outcome write is best-effort async (telemetry sink), so poll for it.
+	deadline := time.Now().Add(2 * time.Second)
+	var rec *store.InferenceRouteRecord
+	for time.Now().Before(deadline) {
+		if rec = findRouteRecord(st, pr.RequestID); rec != nil && rec.FinalStatus != "" {
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+	if rec == nil || rec.FinalStatus == "" {
+		t.Fatal("route record not found / not finalized")
+	}
+	if rec.FinalStatus != "success" {
+		t.Fatalf("final_status = %q, want success", rec.FinalStatus)
+	}
+	if rec.ActualTTFTMs <= 0 {
+		t.Fatalf("actual_ttft_ms must be populated and > 0 for a delivered completion, got %f (0/NULL = the FirstContentAt stamp-ordering bug)", rec.ActualTTFTMs)
+	}
+}
--- a/coordinator/api/ttft_shadow_metrics.go
+++ b/coordinator/api/ttft_shadow_metrics.go
@@ -1,0 +1,79 @@
+package api
+
+import "github.com/eigeninference/d-inference/coordinator/registry"
+
+// Phase-0 TTFT telemetry emission (DogStatsD + the in-process /v1/admin/metrics
+// registry). These are pure observability hooks — no routing behavior changes.
+
+// emitInvalidTTFT fires when applyPendingRouteTelemetry clamped a negative raw
+// time-to-first-token to 0. It is the loud guard against any regression of the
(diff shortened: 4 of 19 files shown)


coordinator/cmd/coordinator/main.go
```diff
@@ -383,6 +383,52 @@ func main() {
...
+	if v := os.Getenv("EIGENINFERENCE_TTFT_ADMISSION_MODE"); v != "" {
+		mode := registry.ParseTTFTAdmissionMode(v)
+		registry.SetTTFTAdmissionMode(mode)
+		if mode == registry.TTFTAdmissionOff {
+			logger.Info("TTFT admission shadow evaluation OFF (EIGENINFERENCE_TTFT_ADMISSION_MODE)", "value", v)
+		} else {
+			logger.Warn("TTFT admission shadow evaluation ENABLED (measurement only — no decision change)",
+				"mode", mode.String(), "deadline_base_ms", registry.TTFTDeadlineBaseMs(), "occupancy_alpha", registry.TTFTOccupancyAlpha())
+		}
+	}
```
ethenotethan, line 433: 🔵 **[INFO]** 🔒 Environment variable parsing lacks bounds validation

💡 _Suggestion:_ Add bounds validation for EIGENINFERENCE_TTFT_OCCUPANCY_ALPHA (e.g., max value) and EIGENINFERENCE_TTFT_DEADLINE_BASE_MS (reasonable range checks) to prevent extreme values that could cause performance issues or unexpected behavior

coordinator/registry/ttft_shadow.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+		owned := providerOwnedBy(p, pr.OwnerAccountID)
+		if pr.SelfRouteOnly && !owned {
+			continue
+		}
+		relaxTrust := owned && (pr.SelfRouteOnly || pr.PreferOwner)
+		snap, ok := r.snapshotProviderLockedEx(p, model, pr.Traits, relaxTrust, false)
+		if !ok || !snap.modelLoaded {
+			continue
+		}
```
ethenotethan: 🟡 **[MEDIUM]** ⚡ loadedIdleAlternativeExistsLocked iterates all providers and snapshots each under its own lock

💡 _Suggestion:_ Consider batching provider snapshots or maintaining a cached index of idle providers to avoid O(n) provider iteration on every request

📊 _Score: 3×4 = 12 · Category: N+1 queries_

coordinator/api/dispatch.go
```diff
@@ -2296,9 +2301,10 @@ func (d *dispatchState) writeCommittedResponse() {
 	// (known up front, adequate for normalization) and the provider's benchmarked
 	// PrefillTPS (set once at registration, read-only thereafter).
 	if pr.Timing != nil && d.firstChunk != "" {
-		if pr.Timing.FirstContentAt.IsZero() {
-			pr.Timing.FirstContentAt = time.Now()
-		}
+		// Stamp under timingMu (MarkFirstContentArrived) because the route-
+		// telemetry path now reads FirstContentAt for actual_ttft_ms from the
+		// provider read-loop goroutine (handleComplete -> applyPendingRouteTelemetry).
+		pr.MarkFirstContentArrived()
```
ethenotethan: 🟡 **[MEDIUM]** ⚡ MarkFirstContentArrived() acquires mutex on hot path during response streaming

💡 _Suggestion:_ Consider using atomic operations or moving timing updates to a separate goroutine to avoid blocking the response stream

📊 _Score: 2×5 = 10 · Category: Blocking I/O on hot paths_

coordinator/api/ttft_shadow_metrics.go
```diff
@@ -0,0 +1,79 @@
...
+			"model", model,
+			"mode", decision.ShadowMode,
+			"would_shed", decision.ShadowWouldShed,
+			"would_redirect_to_idle", decision.ShadowIdleAlternativeExists,
+			"estimate_ms", decision.ShadowEstimateMs,
+			"deadline_ms", decision.ShadowDeadlineMs,
+			"occupancy", decision.ShadowOccupancy,
+			"provider_id", decision.ProviderID,
+		)
+	}
```
ethenotethan, line 78: 🟡 **[MEDIUM]** ⚡ emitTTFTShadowMetrics constructs metric tags and labels repeatedly for each request

💡 _Suggestion:_ Pre-allocate or cache common tag combinations to reduce string allocations on the hot path

coordinator/api/ttft_shadow_metrics.go
```diff
@@ -0,0 +1,79 @@
+package api
+
+import "github.com/eigeninference/d-inference/coordinator/registry"
```
ethenotethan, line 3: 🔵 **[INFO]** 🏷️ Function parameter uses bare interface{} in registry import

💡 _Suggestion:_ Consider using a more specific type instead of relying on interface{} from the registry package if possible

📊 _Score: 2×3 = 6 · Category: bare_interface_

coordinator/store/interface.go
```diff
@@ -212,6 +212,23 @@ type InferenceRouteOutcome struct {
...
+	// on the terminal cancel/error/timeout constructors (and success, harmlessly).
+	CompletionTokensSet bool `json:"completion_tokens_set,omitempty"`
+
+	// InvalidTTFT is a transient (never-persisted) signal that the raw
+	// time-to-first-token computed for this outcome was negative and was clamped
+	// to 0. The single store-submit funnel emits routing.invalid_ttft when it is
+	// set, so any regression of the retried-request shared-Timing bug
+	// (FirstContentAt of an early attempt minus a later attempt's DispatchedAt)
+	// is loud rather than silent. json:"-" keeps it out of every API payload and
+	// neither store impl persists it.
```
ethenotethan, line 230: 🔵 **[INFO]** 🏷️ InvalidTTFT field uses bool but could benefit from typed enum

💡 _Suggestion:_ Consider using a typed enum for InvalidTTFT reasons (e.g., type InvalidTTFTReason int with constants) instead of a boolean flag

📊 _Score: 1×2 = 2 · Category: stringly_typed_

coordinator/api/ttft_shadow_metrics.go
```diff
@@ -0,0 +1,79 @@
...
+			"mode", decision.ShadowMode,
+			"would_shed", decision.ShadowWouldShed,
+			"would_redirect_to_idle", decision.ShadowIdleAlternativeExists,
+			"estimate_ms", decision.ShadowEstimateMs,
+			"deadline_ms", decision.ShadowDeadlineMs,
+			"occupancy", decision.ShadowOccupancy,
+			"provider_id", decision.ProviderID,
+		)
+	}
+}
```
ethenotethan, line 79: 🔵 **[INFO]** 🧩 New file for simple metric emission could be inlined

💡 _Suggestion:_ Consider inlining these simple metric emission functions into the main API file rather than creating a separate file for ~80 lines of straightforward telemetry code

coordinator/registry/ttft_shadow.go
```diff
@@ -0,0 +1,216 @@
...
+		if snapshotOccupancy(snap) == 0 {
+			return true
+		}
+	}
+	return false
+}
```
ethenotethan, line 230: 🟡 **[MEDIUM]** 🧩 Complex configuration system for shadow evaluation with multiple global variables

💡 _Suggestion:_ Consider consolidating the three global config variables (ttftAdmissionMode, ttftDeadlineBaseMs, ttftOccupancyAlpha) into a single config struct to reduce global state and simplify configuration management

📊 _Score: 3×4 = 12 · Category: over-configuration_

coordinator/store/interface.go
ethenotethan, line 230: 🔵 **[INFO]** 🧩 CompletionTokensSet flag adds complexity to handle zero values

💡 _Suggestion:_ Consider using a pointer type (*int) for CompletionTokens to distinguish between unset (nil) and zero (0) values, eliminating the need for the separate CompletionTokensSet boolean flag

📊 _Score: 2×2 = 4 · Category: duplicate logic_

coordinator/registry/scheduler.go
```diff
@@ -1403,6 +1442,31 @@ func PrefillToDecodeRatio() float64 {
...
+func SetTTFTOccupancyAlpha(alpha float64) {
+	if alpha < 0 {
+		alpha = 0
+	}
+	ttftOccupancyAlpha = alpha
+}
+
+// TTFTOccupancyAlpha returns the configured occupancy-term coefficient.
+func TTFTOccupancyAlpha() float64 {
+	return ttftOccupancyAlpha
```
ethenotethan, line 1517: 🔵 **[INFO]** 🧩 Global variable pattern for ttftOccupancyAlpha mirrors existing pattern

💡 _Suggestion:_ While this follows the existing pattern (prefillToDecodeRatio), consider whether these global configuration variables could be consolidated into a configuration struct to reduce global state

coordinator/api/route_outcome.go
```diff
@@ -277,10 +308,34 @@ func applyPendingRouteTelemetry(out *store.InferenceRouteOutcome, pr *registry.P
...
+	}
+	// dispatch_to_first_chunk_ms stays the held-preamble (first-byte) diagnostic.
+	// Clamp negatives so a stale-pointer regression cannot write a -ms value here
+	// either.
+	if !firstChunk.IsZero() && !t.DispatchedAt.IsZero() {
+		if ms := float64(firstChunk.Sub(t.DispatchedAt).Milliseconds()); ms >= 0 {
+			out.DispatchToFirstChunkMs = ms
+		}
 	}
 	if !t.ReceivedAt.IsZero() {
```
ethenotethan, line 351: 🟡 **[MEDIUM]** 🧩 Complex timing logic mixed with outcome construction

💡 _Suggestion:_ Extract the timing calculation logic (FirstContentAt vs FirstChunkAt, negative clamping) into a separate timing utility function to separate timing concerns from outcome construction

📊 _Score: 3×3 = 9 · Category: misplaced responsibility_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 1 finding(s)
- 🔵 **[INFO]** `coordinator/cmd/coordinator/main.go:405-430` — Environment variable parsing lacks bounds validation
  - _Suggestion:_ Add bounds validation for EIGENINFERENCE_TTFT_OCCUPANCY_ALPHA (e.g., max value) and EIGENINFERENCE_TTFT_DEADLINE_BASE_MS (reasonable range checks) to prevent extreme values that could cause performance issues or unexpected behavior

- 🟡 **[MEDIUM]** `coordinator/registry/ttft_shadow.go:192-210` — loadedIdleAlternativeExistsLocked iterates all providers and snapshots each under its own lock
  - _Suggestion:_ Consider batching provider snapshots or maintaining a cached index of idle providers to avoid O(n) provider iteration on every request
- 🟡 **[MEDIUM]** `coordinator/api/dispatch.go:2301-2307` — MarkFirstContentArrived() acquires mutex on hot path during response streaming
  - _Suggestion:_ Consider using atomic operations or moving timing updates to a separate goroutine to avoid blocking the response stream
- 🔵 **[INFO]** `coordinator/registry/scheduler.go:1898-1941` — ttftOccupancyMs recalculates projectedPerRequestDecodeTPS on every call
  - _Suggestion:_ Cache the projectedPerRequestDecodeTPS calculation in the routingSnapshot to avoid recomputation
- 🟡 **[MEDIUM]** `coordinator/api/ttft_shadow_metrics.go:32-78` — emitTTFTShadowMetrics constructs metric tags and labels repeatedly for each request
  - _Suggestion:_ Pre-allocate or cache common tag combinations to reduce string allocations on the hot path

### Type_diligence — 2 finding(s)
- 🔵 **[INFO]** `coordinator/api/ttft_shadow_metrics.go:3` — Function parameter uses bare interface{} in registry import
  - _Suggestion:_ Consider using a more specific type instead of relying on interface{} from the registry package if possible
- 🔵 **[INFO]** `coordinator/store/interface.go:230` — InvalidTTFT field uses bool but could benefit from typed enum
  - _Suggestion:_ Consider using a typed enum for InvalidTTFT reasons (e.g., type InvalidTTFTReason int with constants) instead of a boolean flag

### Additive_complexity — 6 finding(s) (2 blocking)
- 🔵 **[INFO]** `coordinator/api/ttft_shadow_metrics.go:1-79` — New file for simple metric emission could be inlined
  - _Suggestion:_ Consider inlining these simple metric emission functions into the main API file rather than creating a separate file for ~80 lines of straightforward telemetry code
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_shadow.go:1-216` — Complex configuration system for shadow evaluation with multiple global variables
  - _Suggestion:_ Consider consolidating the three global config variables (ttftAdmissionMode, ttftDeadlineBaseMs, ttftOccupancyAlpha) into a single config struct to reduce global state and simplify configuration management
- 🔵 **[INFO]** `coordinator/store/interface.go:215-230` — CompletionTokensSet flag adds complexity to handle zero values
  - _Suggestion:_ Consider using a pointer type (*int) for CompletionTokens to distinguish between unset (nil) and zero (0) values, eliminating the need for the separate CompletionTokensSet boolean flag
- 🔵 **[INFO]** `coordinator/api/route_outcome.go:220-235` — terminalForcesCompletionTokens duplicates status checking logic
  - _Suggestion:_ This status string checking logic appears to duplicate validation patterns elsewhere. Consider centralizing status validation in a shared utility or using constants for status values
- 🔵 **[INFO]** `coordinator/registry/scheduler.go:1442-1467` — Global variable pattern for ttftOccupancyAlpha mirrors existing pattern
  - _Suggestion:_ While this follows the existing pattern (prefillToDecodeRatio), consider whether these global configuration variables could be consolidated into a configuration struct to reduce global state
- 🟡 **[MEDIUM]** `coordinator/api/route_outcome.go:308-340` — Complex timing logic mixed with outcome construction
  - _Suggestion:_ Extract the timing calculation logic (FirstContentAt vs FirstChunkAt, negative clamping) into a separate timing utility function to separate timing concerns from outcome construction

_13 finding(s) total, 5 blocking. Verdict: REQUEST_CHANGES._

coordinator/cmd/coordinator/main.go
ethenotethan: :white_check_mark: Looks addressed in `3ffa01b` — Added comprehensive bounds validation with constants and dedicated validation functions for both environment variables.. Resolving.

Gajesh2007:
## Review fixes pushed (`3ffa01b2`)

All four Codex **P2** functional bugs + the Centaur env-validation **INFO** are addressed. Default behavior stays byte-for-byte (occupancy `alpha=0`, admission `mode=off`); only the always-on telemetry-correctness path changes. `gofmt`, `go vet`, `go test ./...` (21 pkgs) and `go test -race ./registry/... ./api/...` are green.

### Finding → fix map

| Finding | Where | Fix | Test |
|---|---|---|---|
| **A** — `FirstContentAt` stamped too late (Codex P2, `dispatch.go ~2301`) + generic route drops TTFT (Codex P2, `route_outcome.go`) | `api/dispatch.go`, `api/consumer.go`, `api/provider.go` | New `commitFirstContent` helper stamps `FirstContentAt` at the content‑commit site in all 8 dispatch variants (before `waitFirstChunk`/`waitAccepted`'s `successRoutingOutcome` defer); generic `/v1/completions` + `/v1/messages` stamp at first content before `committedRouteOutcome`; **universal fallback** in `handleComplete` stamps when the provider reported delivered tokens, closing the cross‑goroutine race so `completeRouteOutcome` can never persist `actual_ttft_ms` 0/NULL for a delivered reply. | `api/TestFastCompletionStampsActualTTFT` — drives the fast single‑chunk race; **fails without the fix** (verified `0 is not > 0`). |
| **B** — idle shadow scan ignored selection gates (Codex P2, `ttft_shadow.go ~192`) | `registry/ttft_shadow.go` | `loadedIdleAlternativeExistsLocked` now applies the **same** gates the scheduler does: vision capability (`providerServesVisionModelLocked`) + capacity/token‑budget/hardware‑fit feasibility (`buildCandidateWithReason`) — reusing the existing predicates, not reinventing. | `registry/TestLoadedIdleAlternativeHonorsVisionGate` — vision request w/ only text‑only idle peers ⇒ `would_redirect_to_idle=false`; counts a vision‑capable peer once present. |
| **C** — occupancy term used `backendRunning`, not `occ`, for the rate (Codex P2, `scheduler.go ~1924`) | `registry/scheduler.go` | New `projectedPerRequestDecodeTPSAtBatch(snap, occ)`: the observed‑rate **unwind** stays paired with the heartbeat `backendRunning` it was measured at, but the **reapply** uses `occ`, so a herd (`pendingForModel > backendRunning`) is charged the contended rate, not the idle‑batch rate. Super‑linear in `occ`. | `registry/TestTTFTOccupancyTermRateUsesOccupancyNotBackendRunning` — asserts the term uses `occ` and grows with the pending burst at equal heartbeat gauge. |
| **D** — `alpha` could move the live 5s HARD_REJECT ceiling (the approved decoupling) | `registry/scheduler.go`, `registry/ttft_shadow.go` | Occupancy term **removed from `ttftMsFromSnapshot`** (the live cost / `MaxTTFTMs` ceiling / `bestTTFT` input); it now lives only in the new `occupancyAwareTTFTMsFromSnapshot`, used **only** by the shadow evaluator. So with prod `HARD_REJECT=true`, raising `alpha` can no longer tighten the live ceiling and over‑shed ~2×. | `registry/TestTTFTOccupancyAlphaDoesNotMoveLiveCeiling` — `alpha>0`+`shadow` leaves selection + live `TTFTMs` identical to `alpha=0`, while the shadow still computes occupancy‑aware `would_shed`. Existing monotonic test retargeted to the shadow estimate. |
| **E** — env parsing lacked bounds validation (Centaur INFO, `main.go`) | `cmd/coordinator/main.go` | `validateTTFTOccupancyAlpha` (negative→0; `>1e6`/non‑finite→warn+default 0) and `validateTTFTDeadlineBaseMs` (range `[1000,120000]`ms; else warn+default ~10s). Warnings log the offending value, never a secret. | `cmd/coordinator/TestValidateTTFTOccupancyAlpha` + `TestValidateTTFTDeadlineBaseMs` (table tests). |

### Before / After — the load-path safety invariant (D)

```mermaid
flowchart LR
  subgraph Before
    A1[alpha&gt;0] --> B1[ttftMsFromSnapshot += occupancy]
    B1 --> C1[breakdown.TTFTMs]
    C1 --> D1{TTFTMs &gt; MaxTTFTMs 5s?}
    D1 -->|yes| E1[LIVE hard-reject ~2x over-shed]
  end
  subgraph After
    A2[alpha&gt;0] --> B2[ttftMsFromSnapshot occupancy-FREE]
    B2 --> C2[breakdown.TTFTMs unchanged]
    C2 --> D2{TTFTMs &gt; MaxTTFTMs 5s?}
    D2 -->|identical to alpha=0| F2[serve]
    A2 -.shadow only.-> G2[occupancyAwareTTFTMsFromSnapshot vs ~10s base]
    G2 --> H2[would_shed metric — no decision change]
  end
```

### F — E2E `TestIntegration_ConcurrentRequests` verdict: **pre-existing / not our regression**

Root cause is a **provider-side MLX continuous-batching KV-cache crash**, not the coordinator diff:

```
MLX/ErrorHandler.swift:345: Fatal error: [concatenate] All the input array dimensions must match
exactly except for the concatenation axis. However, the provided shapes are (4,8,18,64), (1,8,18,192)
```

The provider process `fatalError`s mid-batch → websocket EOF → all 5 concurrent requests return 502 → `successCount == 0` → the test's only assertion (`successCount > 0`) fails. Evidence it is not ours:

- The **same failure with the identical shapes** `(4,8,18,64)` vs `(1,8,18,192)` occurs on **master at `e118b69a`** (this PR's base) — run `27934769345` — plus a second MLX `broadcast_shapes` crash in `profile_test.go`. The integration suite is intermittently red on master independent of this PR.
- PR #447 is **coordinator-only Go**; nothing here touches `provider-swift`/`libs/mlx-swift`, batching, or KV cache.
- The test asserts **HTTP success count only** — no telemetry/usage fields — so bug A could not have caused it. (Fix A is still in, but it's orthogonal.)

Not blocking; the underlying provider batching crash is tracked separately.

### Nits intentionally not changed (matches existing pattern / out of scope for Phase 0)

- **Consolidate the three globals into a struct** — the `ttftOccupancyAlpha`/`ttftDeadlineBaseMs`/`ttftAdmissionMode` globals mirror the accepted `prefillToDecodeRatio` global pattern; restructuring fights repo convention.
- **`*int` for `CompletionTokens`** — the `CompletionTokensSet` flag is intentional and already tested (sticky 0-on-cancel in both stores).
- **Inline `ttft_shadow_metrics.go`** — a small focused telemetry file matches the repo's "small single-responsibility files" rule.
- **bool → typed enum for `InvalidTTFT`** — single boolean reason today; an enum is premature for Phase 0.
- **Perf nits** (idle-scan O(n), `MarkFirstContentArrived` mutex, `emitTTFTShadowMetrics` tag alloc) — these run only in shadow mode (default off) or once per request via an already-guarded mutex; a cached idle index / atomics are premature optimization for a measurement-only slice.

_Note: in my worktree `core.hooksPath` resolved to the main repo's default `.git/hooks`, so the repo's `.githooks/pre-push` did not auto-run; I ran its exact checks (gofmt + full `go test`) manually instead and did **not** use `--no-verify`._

coordinator/api/dispatch.go
ethenotethan: :white_check_mark: Looks addressed in `3ffa01b` — The MarkFirstContentArrived() call is now consolidated into commitFirstContent() which is called on the same dispatch goroutine that reads chunks, eliminating the mutex acquisition on the hot streamin. Resolving.

coordinator/api/route_outcome.go
ethenotethan: :white_check_mark: Looks addressed in `3ffa01b` — The timing calculation logic has been extracted and separated with clear comments explaining the different timing measurements and their purposes.. Resolving.

coordinator/registry/ttft_shadow.go
```diff
@@ -0,0 +1,249 @@
...
+		// the selection loop applies via buildCandidateWithReason. Without it an
+		// oversized-token or memory-starved peer would be counted as a valid idle
+		// alternative the scheduler could never actually have selected.
+		if _, _, admissible := r.buildCandidateWithReason(snap, pr); !admissible {
+			continue
+		}
+		return true
+	}
+	return false
+}
```
ethenotethan: 🟡 **[MEDIUM]** ⚡ Provider iteration with per-provider mutex acquisition

💡 _Suggestion:_ The loadedIdleAlternativeExistsLocked function iterates through all providers and acquires each provider's mutex individually. Consider batching or restructuring to reduce lock contention, especially under high provider counts.

📊 _Score: 3×3 = 9 · Category: n_plus_one_

coordinator/store/postgres.go
```diff
@@ -1790,6 +1794,7 @@ func (s *PostgresStore) UpdateInferenceRouteOutcome(requestID string, attempt in
 		outcome.CostMicroUSD, outcome.ActualTTFTMs, outcome.DispatchToFirstChunkMs, outcome.TotalDurationMs,
 		outcome.ParseMs, outcome.ReserveMs, outcome.RouteMs, outcome.EncryptMs, outcome.QueueWaitMs, outcome.DispatchMs, outcome.ActualDecodeTPS,
 		outcome.AdmittedButFailed, outcome.UsedBackup, outcome.BackupWon,
+		outcome.CompletionTokensSet,
 	)
 	if err != nil {
 		return fmt.Errorf("store: update inference route outcome: %w", err)
```
ethenotethan, line 1800: 🟡 **[MEDIUM]** ⚡ Synchronous database update on telemetry path

💡 _Suggestion:_ The UpdateInferenceRouteOutcome performs a synchronous database update. Since this is called from telemetry paths that should be non-blocking, consider using async writes or batching updates.

📊 _Score: 2×4 = 8 · Category: blocking_io_

coordinator/registry/scheduler.go
```diff
@@ -1335,6 +1354,26 @@ func effectiveDecodeTPS(staticTPS float64, backendRunning int) float64 {
...
+// every occupancy-keyed decision reads one signal.
+func snapshotOccupancy(snap routingSnapshot) int {
+	occ := snap.pendingForModel
+	if backendDepth := snap.backendRunning + snap.backendWaiting; backendDepth > occ {
+		occ = backendDepth
+	}
+	if occ < 0 {
+		occ = 0
+	}
+	return occ
```
ethenotethan, line 1421: 🔵 **[INFO]** ⚡ Redundant occupancy calculation in snapshotOccupancy

💡 _Suggestion:_ The snapshotOccupancy function is called multiple times with the same snapshot. Consider caching the result in the snapshot struct or passing the pre-calculated value to avoid recomputation.

📊 _Score: 2×2 = 4 · Category: repeated_work_

coordinator/api/route_outcome.go
```diff
@@ -97,6 +107,24 @@ func routeOutcomeWithReason(status, class string, code int, providerReason, erro
...
+// persist completion_tokens even when it is 0. Cancel/error/timeout rows deliver
+// zero tokens and must record 0 (not NULL) so the 0-token cancel population is
+// queryable; partial_success and success are handled by their own count writers.
+func terminalForcesCompletionTokens(status string) bool {
+	switch strings.ToLower(strings.TrimSpace(status)) {
+	case "cancelled", "error", "timeout":
+		return true
+	default:
+		return false
 	}
```
ethenotethan, line 139: 🟡 **[MEDIUM]** 🏷️ terminalForcesCompletionTokens uses string comparison for status values instead of typed constants

💡 _Suggestion:_ Define status constants (e.g., const StatusCancelled = "cancelled") or use an enum type to prevent typos and improve type safety

📊 _Score: 3×3 = 9 · Category: stringly_typed_

coordinator/api/dispatch.go
```diff
@@ -367,6 +372,24 @@ func applyTimingDecomposition(out *store.InferenceRouteOutcome, t *registry.Requ
...
+// read-loop. Without it a fast single-chunk completion could persist
+// actual_ttft_ms as 0/NULL (applyPendingRouteTelemetry derives it solely from
+// FirstContentAt). pr is the COMMITTED attempt — the backup on a speculative
+// backup win, the primary otherwise. MarkFirstChunkArrived is kept (idempotent:
+// it preserves an earlier preamble's first-byte time for dispatch_to_first_chunk_ms).
+func (d *dispatchState) commitFirstContent(pr *registry.PendingRequest, chunk string) {
+	d.firstChunk = chunk
+	pr.MarkFirstChunkArrived()
+	pr.MarkFirstContentArrived()
+}
```
ethenotethan, line 395: 🔵 **[INFO]** 🧩 commitFirstContent wrapper adds minimal value over direct calls

💡 _Suggestion:_ Consider inlining the two calls (MarkFirstChunkArrived + MarkFirstContentArrived) at call sites rather than wrapping in a function

coordinator/api/ttft_shadow_metrics.go
ethenotethan, line 79: 🔵 **[INFO]** 🧩 Metrics emission logic in API layer when registry has the data

💡 _Suggestion:_ Consider moving shadow metrics emission to registry layer where the shadow evaluation happens, keeping API layer focused on HTTP concerns

📊 _Score: 1×2 = 2 · Category: misplaced responsibility_

coordinator/registry/ttft_shadow.go
```diff
@@ -0,0 +1,249 @@
...
+func (e ttftShadowEval) applyTo(d *RoutingDecision) {
+	if d == nil || !e.Evaluated {
+		return
+	}
+	d.ShadowEvaluated = true
+	d.ShadowMode = e.Mode
+	d.ShadowWouldShed = e.WouldShed
+	d.ShadowIdleAlternativeExists = e.IdleAlternativeExists
+	d.ShadowEstimateMs = e.EstimateMs
+	d.ShadowDeadlineMs = e.DeadlineMs
```
ethenotethan, line 142: 🔵 **[INFO]** 🧩 ttftShadowEval.applyTo manually copies fields to RoutingDecision

💡 _Suggestion:_ Use struct embedding or a more direct assignment pattern to avoid manual field-by-field copying

coordinator/store/route_telemetry.go
```diff
@@ -23,7 +23,14 @@ func mergeInferenceRouteOutcome(dst *InferenceRouteOutcome, src *InferenceRouteO
...
+	// CompletionTokensSet force-writes the count even when 0 (terminal cancel/
+	// error/timeout rows deliver 0 tokens and must persist 0, not be skipped as a
+	// zero-value). The flag is sticky so a later commit/latency update with the
+	// default (unset) flag cannot un-set an explicitly recorded 0.
+	if src.CompletionTokensSet {
+		dst.CompletionTokens = src.CompletionTokens
+		dst.CompletionTokensSet = true
+	} else if src.CompletionTokens != 0 {
 		dst.CompletionTokens = src.CompletionTokens
 	}
```
ethenotethan, line 35: 🔵 **[INFO]** 🧩 CompletionTokensSet flag logic duplicated between memory and postgres stores

💡 _Suggestion:_ Extract the completion tokens merge logic into a shared helper function used by both store implementations

📊 _Score: 2×2 = 4 · Category: duplicate logic_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Performance — 5 finding(s) (3 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/dispatch.go:2296-2311` — Synchronous store write on hot path inside submitTelemetry
  - _Suggestion:_ The store.RecordInferenceRoute call is wrapped in submitTelemetry but still executes synchronously on the dispatch goroutine. Consider making this truly async or using a buffered channel to avoid blocking request processing.
- 🔵 **[INFO]** `coordinator/api/provider.go:1920-1950` — Multiple mutex-guarded timing accessor calls
  - _Suggestion:_ Cache the result of pr.FirstChunkAtSafe() instead of calling it multiple times, as each call acquires and releases timingMu.
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_shadow.go:200-249` — Provider iteration with per-provider mutex acquisition
  - _Suggestion:_ The loadedIdleAlternativeExistsLocked function iterates through all providers and acquires each provider's mutex individually. Consider batching or restructuring to reduce lock contention, especially under high provider counts.
- 🟡 **[MEDIUM]** `coordinator/store/postgres.go:1767-1800` — Synchronous database update on telemetry path
  - _Suggestion:_ The UpdateInferenceRouteOutcome performs a synchronous database update. Since this is called from telemetry paths that should be non-blocking, consider using async writes or batching updates.
- 🔵 **[INFO]** `coordinator/registry/scheduler.go:1354-1374` — Redundant occupancy calculation in snapshotOccupancy
  - _Suggestion:_ The snapshotOccupancy function is called multiple times with the same snapshot. Consider caching the result in the snapshot struct or passing the pre-calculated value to avoid recomputation.

### Type_diligence — 3 finding(s) (1 blocking)
- 🔵 **[INFO]** `coordinator/registry/timing.go:234-264` — FirstContentAtSafe returns zero time.Time when pr or pr.Timing is nil without explicit documentation
  - _Suggestion:_ Document the zero-value return behavior or consider returning (time.Time, bool) to make the nil case explicit
- 🔵 **[INFO]** `coordinator/registry/timing.go:222-235` — FirstChunkAtSafe returns zero time.Time when pr or pr.Timing is nil without explicit documentation
  - _Suggestion:_ Document the zero-value return behavior or consider returning (time.Time, bool) to make the nil case explicit
- 🟡 **[MEDIUM]** `coordinator/api/route_outcome.go:120-128` — terminalForcesCompletionTokens uses string comparison for status values instead of typed constants
  - _Suggestion:_ Define status constants (e.g., const StatusCancelled = "cancelled") or use an enum type to prevent typos and improve type safety

### Additive_complexity — 6 finding(s) (1 blocking)
- 🔵 **[INFO]** `coordinator/registry/timing.go:234-264` — FirstContentAtSafe duplicates FirstChunkAtSafe pattern exactly
  - _Suggestion:_ Extract a generic timing field accessor helper to avoid the identical mutex-guarded pattern
- 🔵 **[INFO]** `coordinator/api/dispatch.go:375-391` — commitFirstContent wrapper adds minimal value over direct calls
  - _Suggestion:_ Consider inlining the two calls (MarkFirstChunkArrived + MarkFirstContentArrived) at call sites rather than wrapping in a function
- 🔵 **[INFO]** `coordinator/api/ttft_shadow_metrics.go:1-79` — Metrics emission logic in API layer when registry has the data
  - _Suggestion:_ Consider moving shadow metrics emission to registry layer where the shadow evaluation happens, keeping API layer focused on HTTP concerns
- 🔵 **[INFO]** `coordinator/registry/ttft_shadow.go:129-142` — ttftShadowEval.applyTo manually copies fields to RoutingDecision
  - _Suggestion:_ Use struct embedding or a more direct assignment pattern to avoid manual field-by-field copying
- 🟡 **[MEDIUM]** `coordinator/cmd/coordinator/main.go:917-969` — Three separate validation functions for similar float parsing with overlapping bounds checking
  - _Suggestion:_ Create a generic validateFloatEnv(key, min, max) helper to reduce the duplicated parsing and bounds logic
- 🔵 **[INFO]** `coordinator/store/route_telemetry.go:25-35` — CompletionTokensSet flag logic duplicated between memory and postgres stores
  - _Suggestion:_ Extract the completion tokens merge logic into a shared helper function used by both store implementations

_14 finding(s) total, 5 blocking. Verdict: REQUEST_CHANGES._

Gajesh2007:
## Round 2 review fixes pushed (`02b70845`)

The second Codex pass confirmed the vision + capacity gates from `3ffa01b2` are good, and found the idle-spread scan still missed **three** more selection-eligibility filters — each corrupting `would_redirect_to_idle` for a traffic class. Fixed robustly: `loadedIdleAlternativeExistsLocked` now mirrors `selectBestCandidateScanLocked`'s per-provider eligibility in the same order, with the same helpers, behind a documented correspondence block so it can't drift again. Default behavior stays byte-for-byte (shadow mode off). `gofmt`/`go vet`/`go test ./...`/`go test -race ./registry/... ./api/...` all green.

### Finding → fix map

| Finding | Where | Fix | Test (all verified fail-without-fix) |
|---|---|---|---|
| **1 — prefer-owner pool** | `registry/ttft_shadow.go ~215` | When prefer-owner selected an **owned** winner the selector's pool was owned-only, so a public idle peer is not a real alternative: `winnerOwnedPool := pr.PreferOwner && providerOwnedBy(winner, …)` → skip non-owned peers. | `registry/TestLoadedIdleAlternativeHonorsPreferOwnerPool` (owned-winner ⇒ public peer not counted; plain request + owned idle peer contrasts). |
| **2 — provider allowlist** | `registry/ttft_shadow.go ~210` | Honor `AllowedProviderSerials` via `providerMatchesAllowedSerial` (the selector's allowlist gate). | `registry/TestLoadedIdleAlternativeHonorsAllowlist`. |
| **3 — retry/backup excludeIDs** | `registry/scheduler.go ~285` + `registry/ttft_shadow.go` | `ReserveProviderEx` now threads its `excludeIDs` into `evaluateTTFTShadowLocked` → `loadedIdleAlternativeExistsLocked`, which skips excluded providers (the selector's `excludeSet`). | `registry/TestLoadedIdleAlternativeHonorsExcludeIDs`. |
| **status constants (nit)** | `api/route_outcome.go ~128` | Added `finalStatus{Success,PartialSuccess,Error,Cancelled,Timeout}` and replaced bare `"cancelled"/"error"/"timeout"/"success"/"partial_success"` comparisons + constructors. | covered by existing route-outcome tests. |

The scan now honors the full set: exclude → self-route → allowlist → prefer-owner pool → structural/trait/trust (`snapshotProviderLockedEx`) → instantly-usable (resident + zero occupancy) → vision → capacity (`buildCandidateWithReason`).

### Already addressed in `3ffa01b2` — stale re-posts, no action
- **Generic-route TTFT** (`route_outcome.go:322`): already fixed — content is stamped at `consumer.go:3658`/`:3728` and via the `handleComplete` fallback (`provider.go:1913`); successful `/v1/completions` + `/v1/messages` persist a non-zero `actual_ttft_ms`.
- **Idle-scan vision gate** (`ttft_shadow.go:219`): already added in `3ffa01b2` (vision + capacity), now extended with the three filters above.

### Intentionally skipped nits (fight existing patterns / out of scope for Phase 0)
- **Consolidate the three globals into a struct** — they mirror the accepted `prefillToDecodeRatio` global pattern.
- **`*int` for `CompletionTokens`** — the `CompletionTokensSet` flag is intentional and tested (sticky 0-on-cancel, both stores).
- **Inline `ttft_shadow_metrics.go`** — a small focused file matches the repo's "small single-responsibility files" rule.
- **bool → typed enum for `InvalidTTFT`** — single boolean reason; premature for Phase 0.
- **"Inline `commitFirstContent`"** — the helper de-duplicates the stamp across 8 dispatch sites; inlining would copy-paste it 8×.
- **Async store write for the outcome** — already async via the telemetry sink; no change needed.

_Note: in this worktree `core.hooksPath` resolves to the main repo's default `.git/hooks`, so the repo `.githooks/pre-push` doesn't auto-run; I ran its exact checks (gofmt + full `go test` + `-race`) manually and did **not** use `--no-verify`._

coordinator/api/ttft_shadow_metrics.go
ethenotethan: :white_check_mark: Looks addressed in `02b7084` — The code now uses the specific type registry.RoutingDecision instead of interface{}. Resolving.

coordinator/registry/scheduler.go
ethenotethan: :white_check_mark: Looks addressed in `02b7084` — The code now uses a centralized snapshotOccupancy function instead of recalculating occupancy inline.. Resolving.

coordinator/api/route_outcome.go
ethenotethan: :white_check_mark: Looks addressed in `02b7084` — String literals replaced with typed constants (finalStatusSuccess, finalStatusError, etc.) throughout the terminalForcesCompletionTokens function and related code.. Resolving.

coordinator/store/route_telemetry.go
ethenotethan: :white_check_mark: Looks addressed in `02b7084` — The completion tokens merge logic has been consolidated into the shared mergeInferenceRouteOutcome function.. Resolving.

coordinator/registry/ttft_shadow.go
```diff
@@ -0,0 +1,289 @@
...
+		// the selection loop applies via buildCandidateWithReason. Without it an
+		// oversized-token or memory-starved peer would be counted as a valid idle
+		// alternative the scheduler could never actually have selected.
+		if _, _, admissible := r.buildCandidateWithReason(snap, pr); !admissible {
+			continue
+		}
+		return true
+	}
+	return false
+}
```
ethenotethan, line 230: 🔵 **[INFO]** 🔒 New shadow evaluation code lacks input validation on environment variables

💡 _Suggestion:_ Add bounds checking for EIGENINFERENCE_TTFT_OCCUPANCY_ALPHA and EIGENINFERENCE_TTFT_DEADLINE_BASE_MS in the shadow evaluation logic

📊 _Score: 2×1 = 2 · Category: missing-input-validation_

coordinator/registry/scheduler.go
```diff
@@ -1527,7 +1615,7 @@ func projectedPerRequestDecodeTPS(snap routingSnapshot) float64 {
 	if solo <= 0 {
```
ethenotethan, line 1659: 🟡 **[MEDIUM]** ⚡ Redundant TPS calculations in projectedPerRequestDecodeTPSAtBatch

💡 _Suggestion:_ Cache the solo TPS calculation when bObserved == joinBatch to avoid recomputing the same unwinding formula

📊 _Score: 3×4 = 12 · Category: repeated_work_

coordinator/registry/ttft_shadow.go
```diff
@@ -0,0 +1,289 @@
...
+		if pr.RequiresVision {
+			p.mu.Lock()
+			servesVision := r.providerServesVisionModelLocked(p, model)
+			p.mu.Unlock()
+			if !servesVision {
+				continue
+			}
+		}
+		// Capacity / token-budget / hardware-fit feasibility — the same predicate
+		// the selection loop applies via buildCandidateWithReason. Without it an
```
ethenotethan: 🔵 **[INFO]** ⚡ Map allocations in loadedIdleAlternativeExistsLocked grow with provider count

💡 _Suggestion:_ Pre-allocate excludeSet and allowedSerials maps with known capacity: make(map[string]struct{}, len(excludeIDs)) and make(map[string]struct{}, len(pr.AllowedProviderSerials))

📊 _Score: 2×3 = 6 · Category: unbounded_allocations_

coordinator/api/dispatch.go
```diff
@@ -320,6 +320,11 @@ func (d *dispatchState) recordRoutingDecisionFor(provider *registry.Provider, pr
 		provider.Mu().Unlock()
 	}
 
+	// Phase-0 shadow TTFT admission/spread metrics. No-op unless the request was
+	// evaluated (admission mode != off AND a provider was selected). Emitted on
+	// the synchronous path (cheap counter incr), not inside the async store write.
+	s.emitTTFTShadowMetrics(d.model, decision)
+
```
ethenotethan, line 327: 🟡 **[MEDIUM]** ⚡ Synchronous telemetry emission on hot dispatch path

💡 _Suggestion:_ Move emitTTFTShadowMetrics call inside the async submitTelemetry block to avoid blocking the dispatch goroutine

📊 _Score: 2×4 = 8 · Category: blocking_io_

coordinator/registry/ttft_shadow.go
ethenotethan: 🟡 **[MEDIUM]** ⚡ O(n) provider scan for idle alternatives on every shadow evaluation

💡 _Suggestion:_ Maintain an index of idle providers per model or use early termination when first idle alternative is found

📊 _Score: 3×3 = 9 · Category: inefficient_data_structures_

coordinator/registry/ttft_shadow.go
```diff
@@ -0,0 +1,289 @@
...
+	// owned candidate can serve — and the winner being owned is exactly that
+	// signal (prefer-owner picks from the owned pool when it is non-empty). So a
+	// public peer is only a real alternative when the winner itself is public.
+	winnerOwnedPool := pr.PreferOwner && providerOwnedBy(winner, pr.OwnerAccountID)
+
+	for _, p := range r.providers {
+		if p == nil || p.ID == winnerID {
+			continue
+		}
+		if _, excluded := excludeSet[p.ID]; excluded {
```
ethenotethan: 🟡 **[MEDIUM]** 🏷️ map[string]struct{} used for set membership without type safety

💡 _Suggestion:_ Consider using a typed set type or generic Set[string] to prevent accidental insertion of non-string keys

📊 _Score: 3×3 = 9 · Category: bare_interface_

coordinator/registry/ttft_shadow.go
```diff
@@ -0,0 +1,289 @@
...
+			continue
+		}
+		owned := providerOwnedBy(p, pr.OwnerAccountID)
+		if pr.SelfRouteOnly && !owned {
+			continue
```
ethenotethan: 🟡 **[MEDIUM]** 🏷️ map[string]struct{} used for set membership without type safety

💡 _Suggestion:_ Consider using a typed set type or generic Set[string] to prevent accidental insertion of non-string keys

📊 _Score: 3×3 = 9 · Category: bare_interface_

coordinator/registry/ttft_shadow.go
ethenotethan, line 230: 🟡 **[MEDIUM]** 🧩 Entire new file for shadow evaluation that could be simpler

💡 _Suggestion:_ Consider if this 289-line shadow evaluation system could be implemented as a simpler function or method rather than a separate file with multiple types and configuration globals

coordinator/api/ttft_shadow_metrics.go
ethenotethan, line 79: 🔵 **[INFO]** 🧩 Separate file for metrics emission that could be inline

💡 _Suggestion:_ These metric emission functions could be methods on Server rather than a separate file, reducing file proliferation

coordinator/registry/ttft_shadow.go
```diff
@@ -0,0 +1,289 @@
...
+// before serving starts.
+func SetTTFTDeadlineBaseMs(ms float64) {
+	if ms > 0 {
+		ttftDeadlineBaseMs = ms
+	}
+}
+
+// TTFTDeadlineBaseMs returns the configured shadow-evaluation deadline base (ms).
+func TTFTDeadlineBaseMs() float64 {
+	return ttftDeadlineBaseMs
```
ethenotethan, line 106: 🟡 **[MEDIUM]** 🧩 Multiple global configuration variables for shadow evaluation

💡 _Suggestion:_ Consider consolidating ttftAdmissionMode, ttftDeadlineBaseMs, and ttftOccupancyAlpha into a single configuration struct to reduce global state

📊 _Score: 3×4 = 12 · Category: over-configuration_

coordinator/cmd/coordinator/main.go
```diff
@@ -846,6 +895,55 @@ func parseAPNsEnforceAfter() (time.Time, error) {
...
+	if v < minTTFTDeadlineBaseMs || v > maxTTFTDeadlineBaseMs {
+		return 0, false
+	}
+	return v, true
+}
+
 // loadAPNsAttestor builds the production APNs code-identity attestor from the
 // environment, or returns nil (feature disabled) when unconfigured. Required:
 // APNS_KEY_ID, APNS_TEAM_ID, and the .p8 auth key via APNS_AUTH_KEY_P8_B64
```
ethenotethan, line 949: 🟡 **[MEDIUM]** 🧩 Complex validation functions for environment variables

💡 _Suggestion:_ The validation logic for TTFT configuration could be simplified using a common validation pattern or moved to a configuration package

📊 _Score: 3×3 = 9 · Category: over-configuration_

coordinator/registry/scheduler.go
ethenotethan, line 1659: 🔵 **[INFO]** 🧩 projectedPerRequestDecodeTPSAtBatch duplicates projectedPerRequestDecodeTPS logic

💡 _Suggestion:_ Consider refactoring to have projectedPerRequestDecodeTPS call projectedPerRequestDecodeTPSAtBatch with snap.backendRunning to eliminate duplication

coordinator/api/ttft_first_content_test.go
```diff
@@ -0,0 +1,85 @@
...
+	if rec == nil || rec.FinalStatus == "" {
+		t.Fatal("route record not found / not finalized")
+	}
+	if rec.FinalStatus != "success" {
+		t.Fatalf("final_status = %q, want success", rec.FinalStatus)
+	}
+	if rec.ActualTTFTMs <= 0 {
+		t.Fatalf("actual_ttft_ms must be populated and > 0 for a delivered completion, got %f (0/NULL = the FirstContentAt stamp-ordering bug)", rec.ActualTTFTMs)
+	}
+}
```
ethenotethan, line 90: 🔵 **[INFO]** 🧩 Entire test file for single regression test

💡 _Suggestion:_ This single test could be moved into an existing test file rather than creating a new file

coordinator/store/interface.go
```diff
@@ -212,6 +212,23 @@ type InferenceRouteOutcome struct {
...
+	InvalidTTFT bool `json:"-"`
```
ethenotethan, line 231: 🔵 **[INFO]** 🧩 Transient InvalidTTFT field in persistent data structure

💡 _Suggestion:_ Consider moving InvalidTTFT to a separate validation result type rather than embedding transient state in the persistent InferenceRouteOutcome struct

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 1 finding(s)
- 🔵 **[INFO]** `coordinator/registry/ttft_shadow.go:1-289` — New shadow evaluation code lacks input validation on environment variables
  - _Suggestion:_ Add bounds checking for EIGENINFERENCE_TTFT_OCCUPANCY_ALPHA and EIGENINFERENCE_TTFT_DEADLINE_BASE_MS in the shadow evaluation logic

### Performance — 5 finding(s) (4 blocking)
- 🟡 **[MEDIUM]** `coordinator/registry/scheduler.go:1571-1615` — Redundant TPS calculations in projectedPerRequestDecodeTPSAtBatch
  - _Suggestion:_ Cache the solo TPS calculation when bObserved == joinBatch to avoid recomputing the same unwinding formula
- 🔵 **[INFO]** `coordinator/registry/ttft_shadow.go:207-280` — Map allocations in loadedIdleAlternativeExistsLocked grow with provider count
  - _Suggestion:_ Pre-allocate excludeSet and allowedSerials maps with known capacity: make(map[string]struct{}, len(excludeIDs)) and make(map[string]struct{}, len(pr.AllowedProviderSerials))
- 🟡 **[MEDIUM]** `coordinator/api/dispatch.go:320-327` — Synchronous telemetry emission on hot dispatch path
  - _Suggestion:_ Move emitTTFTShadowMetrics call inside the async submitTelemetry block to avoid blocking the dispatch goroutine
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_shadow.go:230-280` — O(n) provider scan for idle alternatives on every shadow evaluation
  - _Suggestion:_ Maintain an index of idle providers per model or use early termination when first idle alternative is found
- 🟡 **[MEDIUM]** `coordinator/api/route_outcome.go:319-355` — Redundant timing calculations in applyPendingRouteTelemetry
  - _Suggestion:_ Cache the DispatchedAt.IsZero() check result and reuse it for both FirstContentAt and FirstChunkAt calculations

### Type_diligence — 2 finding(s) (2 blocking)
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_shadow.go:244` — map[string]struct{} used for set membership without type safety
  - _Suggestion:_ Consider using a typed set type or generic Set[string] to prevent accidental insertion of non-string keys
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_shadow.go:249` — map[string]struct{} used for set membership without type safety
  - _Suggestion:_ Consider using a typed set type or generic Set[string] to prevent accidental insertion of non-string keys

### Additive_complexity — 8 finding(s) (3 blocking)
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_shadow.go:1-289` — Entire new file for shadow evaluation that could be simpler
  - _Suggestion:_ Consider if this 289-line shadow evaluation system could be implemented as a simpler function or method rather than a separate file with multiple types and configuration globals
- 🔵 **[INFO]** `coordinator/api/ttft_shadow_metrics.go:1-79` — Separate file for metrics emission that could be inline
  - _Suggestion:_ These metric emission functions could be methods on Server rather than a separate file, reducing file proliferation
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_shadow.go:81-106` — Multiple global configuration variables for shadow evaluation
  - _Suggestion:_ Consider consolidating ttftAdmissionMode, ttftDeadlineBaseMs, and ttftOccupancyAlpha into a single configuration struct to reduce global state
- 🔵 **[INFO]** `coordinator/api/route_outcome.go:35-49` — String constants that could use existing patterns
  - _Suggestion:_ Consider if these final status constants could reuse existing error classification patterns rather than introducing new string literals
- 🟡 **[MEDIUM]** `coordinator/cmd/coordinator/main.go:895-949` — Complex validation functions for environment variables
  - _Suggestion:_ The validation logic for TTFT configuration could be simplified using a common validation pattern or moved to a configuration package
- 🔵 **[INFO]** `coordinator/registry/scheduler.go:1571-1615` — projectedPerRequestDecodeTPSAtBatch duplicates projectedPerRequestDecodeTPS logic
  - _Suggestion:_ Consider refactoring to have projectedPerRequestDecodeTPS call projectedPerRequestDecodeTPSAtBatch with snap.backendRunning to eliminate duplication
- 🔵 **[INFO]** `coordinator/api/ttft_first_content_test.go:1-85` — Entire test file for single regression test
  - _Suggestion:_ This single test could be moved into an existing test file rather than creating a new file
- 🔵 **[INFO]** `coordinator/store/interface.go:214-231` — Transient InvalidTTFT field in persistent data structure
  - _Suggestion:_ Consider moving InvalidTTFT to a separate validation result type rather than embedding transient state in the persistent InferenceRouteOutcome struct

_16 finding(s) total, 9 blocking. Verdict: REQUEST_CHANGES._

Gajesh2007:
## Round 3 fixes + DRY/cleanliness pushed (`93f28a87`)

Two new Codex P2s fixed, plus the DRY/readability pass. Default behavior stays byte-for-byte (alpha=0, mode=off). `gofmt` / `go vet` / `go test ./...` (21 pkgs) / `go test -race ./registry/... ./api/...` all green.

### New Codex P2 #1 (functional) — first-content fallback could stamp an abandoned attempt
`api/provider.go` ~1914. Retries SHARE one `RequestTiming` pointer (`dispatch.go` sets `Timing: d.timing` on every attempt). An abandoned attempt completing late (provider finished after the coordinator timed it out and retried) could win `RemovePending` and — via the ungated fallback — stamp `FirstContentAt` first; since it is first-write-wins, the real committed retry couldn't replace it, clamping/zeroing its `actual_ttft_ms`.

**Fix:** a per-attempt `contentCommitted` flag (`registry.PendingRequest.MarkContentCommitted` / `ContentCommittedSafe`, guarded by `timingMu`) is set by `commitFirstContent` (dispatch, `dispatch.go`) and the generic first-content stamps (`consumer.go:3660`/`:3730`). The fallback now stamps only when `ContentCommittedSafe()` — strictly the committed attempt — so an abandoned attempt can never corrupt the shared `Timing`.
**Tests:** `api/TestAbandonedAttemptDoesNotStampCommittedTTFT` (abandoned completion must NOT stamp the shared `FirstContentAt`; the committed retry's first-content time wins) + the committed-attempt test updated. Both verified to fail without the fix.

### New Codex P2 #2 (metric accuracy) — idle scan missed post-candidate POOL filters; fixed by REUSING the selector
`registry/ttft_shadow.go` ~283 + `registry/scheduler.go`. The scan kept re-deriving eligibility and missing filters (3rd round: the `AvoidVersion` diverse-version pool + the `MinDecodeTPS` quality floor, both applied AFTER per-provider candidate construction).

**Fix (robust, single source of truth):** extracted `scanCandidatesLocked` — the SINGLE eligibility computation: every per-provider gate (exclude / self-route / allowlist / structural-trait-trust / vision / capacity / TTFT ceiling) **and** the post-candidate pool narrowing (prefer-owner / `AvoidVersion` / `MinDecodeTPS`). `selectBestCandidateScanLocked` now builds the pool via it and only adds cost-ranking; `loadedIdleAlternativeExistsLocked` derives idle alternatives from the *same* pool (any non-winner candidate that is loaded + zero-occupancy). The shadow scan can no longer drift from selection — and the manual gate-mirroring (~45 lines) is gone.
**Tests:** `registry/TestLoadedIdleAlternativeHonorsAvoidVersionPool` + `...HonorsMinDecodeTPS` (fail without the fix). The earlier exclude/allowlist/prefer-owner/vision/capacity idle tests still pass through the unified pool.

```mermaid
flowchart LR
  subgraph Before
    S1[selectBestCandidateScanLocked\nper-provider gates + pool narrowing] --> W1[winner]
    I1[loadedIdleAlternativeExistsLocked\nRE-DERIVES gates piecemeal] -. drifts: misses AvoidVersion/MinDecodeTPS .-> M1[would_redirect_to_idle ✗]
  end
  subgraph After
    SC[scanCandidatesLocked\nSINGLE eligibility source] --> S2[selectBestCandidateScanLocked\n= scan + cost rank] --> W2[winner]
    SC --> I2[loadedIdleAlternativeExistsLocked\n= scan pool, pick loaded-idle] --> M2[would_redirect_to_idle ✓]
  end
```

### DRY / cleanliness
- `projectedPerRequestDecodeTPS(snap)` already delegates to `projectedPerRequestDecodeTPSAtBatch(snap, snap.backendRunning)` (single implementation, from the prior commit) — confirmed, no further change.
- The selector-reuse refactor IS the readability win: one eligibility source, ~45 fewer lines of duplicated gate logic in the shadow scan, plus doc-comments on `candidateScan` / `scanCandidatesLocked` / the `contentCommitted` flag / the occupancy term / the shadow eval / the FirstContentAt ordering. No unrelated churn.

### Declined over-engineering nits (Centaur) — with reasons
- **Consolidate the config globals into a struct** — they mirror the accepted `prefillToDecodeRatio` global pattern.
- **`map[string]struct{}` → generic Set** — `map[T]struct{}` is the idiomatic Go set; a wrapper adds indirection without value.
- **"Split files are over-abstraction" / merge the single-test file** — the repo convention is small, single-responsibility files; `ttft_shadow.go` / `ttft_shadow_metrics.go` / `ttft_first_content_test.go` follow it.
- **`*int` for `CompletionTokens`** — the `CompletionTokensSet` flag is intentional and tested (sticky 0-on-cancel, both stores).
- **Move `InvalidTTFT` out / bool→enum** — single boolean reason today; premature.

### Centaur false positive
- **"Missing env validation"** is inaccurate: `validateTTFTOccupancyAlpha` and `validateTTFTDeadlineBaseMs` (with `[1000,120000]` ms range + the alpha clamp/upper bound) already exist in `cmd/coordinator/main.go` and are table-tested.

_Note: this worktree's `core.hooksPath` resolves to the main repo's default `.git/hooks`, so the repo `.githooks/pre-push` doesn't auto-run; I ran its exact checks (gofmt + full `go test` + `-race`) manually and did **not** use `--no-verify`._

coordinator/registry/scheduler.go
```diff
@@ -1527,7 +1659,7 @@ func projectedPerRequestDecodeTPS(snap routingSnapshot) float64 {
...
```
ethenotethan, line 1659: 🔵 **[INFO]** ⚡ projectedPerRequestDecodeTPSAtBatch recalculates load factor and validation on every call

💡 _Suggestion:_ Cache validated load factor and batch values when they haven't changed to avoid repeated calculations in hot paths

📊 _Score: 2×3 = 6 · Category: repeated_computation_

coordinator/registry/ttft_shadow.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+		// Instantly usable: the model is resident AND no work is occupying the box
+		// (a herded peer is not a spread target). The candidate already passed
+		// every selection gate via scanCandidatesLocked.
+		if c.snapshot.modelLoaded && snapshotOccupancy(c.snapshot) == 0 {
+			return true
+		}
+	}
+	return false
+}
```
ethenotethan, line 230: 🟡 **[MEDIUM]** ⚡ loadedIdleAlternativeExistsLocked iterates through candidate pool and locks each provider individually

💡 _Suggestion:_ Batch provider snapshot operations or pre-filter candidates to reduce per-provider locking overhead in the shadow evaluation path

📊 _Score: 3×3 = 9 · Category: n_plus_one_

coordinator/api/ttft_shadow_metrics.go
ethenotethan, line 79: 🔵 **[INFO]** 🏷️ String literals used for metric tags instead of typed constants

💡 _Suggestion:_ Define typed constants for metric tag values like 'would_shed', 'would_serve', 'true', 'false' to prevent typos and enable compile-time validation

📊 _Score: 2×3 = 6 · Category: stringly_typed_

coordinator/store/interface.go
ethenotethan, line 231: 🔵 **[INFO]** 🏷️ Boolean fields rely on Go zero values without explicit initialization

💡 _Suggestion:_ Consider explicit initialization of boolean fields like CompletionTokensSet and InvalidTTFT in constructors to make intent clear

📊 _Score: 2×2 = 4 · Category: implicit_zero_value_

coordinator/registry/ttft_shadow.go
ethenotethan, line 230: 🟡 **[MEDIUM]** 🧩 Entire shadow evaluation system adds significant complexity for measurement-only functionality

💡 _Suggestion:_ Consider if this shadow system could be simplified to basic metrics collection without the full routing decision simulation

coordinator/api/ttft_shadow_metrics.go
ethenotethan, line 79: 🔵 **[INFO]** 🧩 Metrics emission duplicated across DogStatsD and in-process registry

💡 _Suggestion:_ Extract common metric emission pattern to avoid maintaining two parallel metric systems

coordinator/api/dispatch.go
```diff
@@ -367,6 +372,28 @@ func applyTimingDecomposition(out *store.InferenceRouteOutcome, t *registry.Requ
...
+// it preserves an earlier preamble's first-byte time for dispatch_to_first_chunk_ms).
+func (d *dispatchState) commitFirstContent(pr *registry.PendingRequest, chunk string) {
+	d.firstChunk = chunk
+	pr.MarkFirstChunkArrived()
+	pr.MarkFirstContentArrived()
+	// Mark THIS attempt as the committed one so handleComplete's fallback only
+	// ever stamps FirstContentAt for the attempt that actually delivered content —
+	// never a late-completing abandoned/retried attempt sharing the same Timing.
+	pr.MarkContentCommitted()
+}
```
ethenotethan, line 395: 🔵 **[INFO]** 🧩 commitFirstContent function mixes timing concerns with dispatch state management

💡 _Suggestion:_ Split timing operations from dispatch state changes into separate focused functions

📊 _Score: 2×2 = 4 · Category: misplaced responsibility_

coordinator/store/interface.go
ethenotethan, line 231: 🔵 **[INFO]** 🧩 CompletionTokensSet flag adds complexity to handle edge case that could be solved differently

💡 _Suggestion:_ Consider using a pointer type or sentinel value instead of adding a separate boolean flag

coordinator/cmd/coordinator/main.go
ethenotethan, line 949: 🔵 **[INFO]** 🧩 Complex validation functions for environment variables that could use a simpler approach

💡 _Suggestion:_ Consider using a configuration struct with validation tags instead of individual validation functions

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 1 finding(s)
- 🔵 **[INFO]** `coordinator/registry/timing.go:195-244` — RequestTiming fields lack mutex protection during concurrent access
  - _Suggestion:_ Add mutex protection for all RequestTiming field access or document thread safety requirements

- 🟡 **[MEDIUM]** `coordinator/api/dispatch.go:1055-1102` — Repeated commitFirstContent calls across multiple dispatch paths
  - _Suggestion:_ Extract common chunk processing logic to avoid code duplication and ensure consistent behavior across all dispatch paths
- 🔵 **[INFO]** `coordinator/registry/scheduler.go:1615-1659` — projectedPerRequestDecodeTPSAtBatch recalculates load factor and validation on every call
  - _Suggestion:_ Cache validated load factor and batch values when they haven't changed to avoid repeated calculations in hot paths
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_shadow.go:200-230` — loadedIdleAlternativeExistsLocked iterates through candidate pool and locks each provider individually
  - _Suggestion:_ Batch provider snapshot operations or pre-filter candidates to reduce per-provider locking overhead in the shadow evaluation path
- 🟡 **[MEDIUM]** `coordinator/api/route_outcome.go:319-355` — applyPendingRouteTelemetry performs multiple time calculations and mutex operations
  - _Suggestion:_ Cache timing calculations and reduce mutex lock/unlock cycles by batching timing reads under a single lock

### Type_diligence — 3 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/registry/timing.go:195` — RequestTiming fields accessed without type safety
  - _Suggestion:_ Add type assertions when accessing Timing fields or use a typed interface instead of relying on struct field access patterns
- 🔵 **[INFO]** `coordinator/api/ttft_shadow_metrics.go:42-79` — String literals used for metric tags instead of typed constants
  - _Suggestion:_ Define typed constants for metric tag values like 'would_shed', 'would_serve', 'true', 'false' to prevent typos and enable compile-time validation
- 🔵 **[INFO]** `coordinator/store/interface.go:212-231` — Boolean fields rely on Go zero values without explicit initialization
  - _Suggestion:_ Consider explicit initialization of boolean fields like CompletionTokensSet and InvalidTTFT in constructors to make intent clear

### Additive_complexity — 7 finding(s) (2 blocking)
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_shadow.go:1-230` — Entire shadow evaluation system adds significant complexity for measurement-only functionality
  - _Suggestion:_ Consider if this shadow system could be simplified to basic metrics collection without the full routing decision simulation
- 🔵 **[INFO]** `coordinator/api/ttft_shadow_metrics.go:1-79` — Metrics emission duplicated across DogStatsD and in-process registry
  - _Suggestion:_ Extract common metric emission pattern to avoid maintaining two parallel metric systems
- 🔵 **[INFO]** `coordinator/api/route_outcome.go:61-85` — Multiple final status constants when a simple enum or map would suffice
  - _Suggestion:_ Replace individual constants with a slice or map of valid statuses to reduce maintenance burden
- 🟡 **[MEDIUM]** `coordinator/registry/scheduler.go:414-433` — candidateScan struct wraps what could be a simple return tuple
  - _Suggestion:_ Consider returning multiple values directly instead of wrapping in a struct for a single internal function
- 🔵 **[INFO]** `coordinator/api/dispatch.go:372-395` — commitFirstContent function mixes timing concerns with dispatch state management
  - _Suggestion:_ Split timing operations from dispatch state changes into separate focused functions
- 🔵 **[INFO]** `coordinator/store/interface.go:215-231` — CompletionTokensSet flag adds complexity to handle edge case that could be solved differently
  - _Suggestion:_ Consider using a pointer type or sentinel value instead of adding a separate boolean flag
- 🔵 **[INFO]** `coordinator/cmd/coordinator/main.go:895-949` — Complex validation functions for environment variables that could use a simpler approach
  - _Suggestion:_ Consider using a configuration struct with validation tags instead of individual validation functions

_15 finding(s) total, 6 blocking. Verdict: REQUEST_CHANGES._

PR #494, opened 2026-07-02
fix(terminal-paths): refund observability, chunk-key lifecycle, 499 classification, writer deadline races

Gajesh2007:
## Summary

Terminal-path correctness fixes — the top follow-ups from the #491 review plus the P0 swallowed-refund bug. Four fixes, all about what happens when a request or connection *ends*: money that silently vanished, session keys that outlived their requests, backpressure aborts booked as provider faults, and two writer races at the deadline/shutdown boundary.

Coordinator-only; no wire-format changes; `consumer.go`/`registry.go`/`scheduler.go` untouched (in-flight v0.6.31 work) — deferred sites listed below.

### Behavior — request/connection terminals

```mermaid
flowchart LR
  subgraph Before
    A1[refund credit fails] --> B1["silently dropped<br/>(_ = store.Credit)"]
    C1[request abandoned:<br/>retry / cancel / disconnect / grace-expiry] --> D1["session key stays cached<br/>until 8192 cap-reset (maybe never)"]
    E1[overflow-499 / client cancel<br/>after commit] --> F1["route outcome: provider_error_after_commit<br/>AdmittedButFailed=true — pollutes calibration"]
    G1["write completes just as<br/>deadline expires"] --> H1["reported SUCCESS on a socket the<br/>watchdog killed -> TTFT burned + fake fault;<br/>writeTimedOut latch misattributes next frame;<br/>~50% 'stopped' for successfully-written frames"]
  end
  subgraph After
    A2[refund credit fails] --> B2["ERROR log + billing.refund_failures metric<br/>(alerting hook)"]
    C2[ALL terminal paths] --> D2["key forgotten immediately;<br/>zeroed when on the read-loop goroutine"]
    E2[consumer-cancel terminal] --> F2["partial_success / consumer_cancel_after_commit<br/>AdmittedButFailed=false"]
    G2[deadline boundary] --> H2["CAS handoff: exactly one winner;<br/>lost write returns errProviderWriteTimeout even on nil err;<br/>per-frame attribution; final done-drain before 'stopped'"]
  end
```

### Code — control flow

```mermaid
flowchart LR
  subgraph Before
    a1["cancelDispatch x38 (dispatch.go)<br/>grace expiry, disconnect, retry-reassign"] --> b1["no chunkKeys cleanup"]
    c1["magic literals: 499,<br/>'request cancelled' (2 sites)"] --> d1["postCommitProviderErrorOutcome:<br/>always provider fault"]
    e1["writeFrame: bare Load + sticky<br/>writeTimedOut bool"] --> f1["writeLane: random select between<br/>req.done and w.done"]
  end
  subgraph After
    a2["cancelDispatchAndForget wrapper;<br/>settlement onExpiry forget;<br/>forgetProviderPendingKeys (new<br/>registry/pending_iter.go accessor);<br/>forget-before-reassign"] --> b2["chunkKeyCache: forget (cross-goroutine)<br/>vs forgetAndZero (read-loop only,<br/>zeroing policy documented)"]
    c2["terminal_class.go:<br/>statusClientClosedRequest,<br/>isConsumerCancelTerminal()"] --> d2["postCommitProviderErrorOutcome<br/>short-circuits consumer-cancels"]
    e2["writeDeadline value IS the CAS token<br/>(nonzero->0); watchdog and writeFrame<br/>race to claim; writeTimedOut deleted"] --> f2["awaitWriteResult: non-blocking<br/>req.done drain before stoppedResult"]
  end
```

### 1. Swallowed refund credit (money)
`releaseInitialReservation` dropped consumer refunds on store failure with `_ =`. Now: ERROR log (account/model/amount) + `billing.refund_failures` metric, matching the platform-fee path's documented "never swallow it" policy. No inline retry — this runs on latency-sensitive paths; the metric is the alerting hook.

### 2. `chunkKeys` lifecycle (crypto hygiene — highest-agreement #491 review finding)
Only `handleComplete`/`handleInferenceError` forgot the memoized X25519 shared key; every consumer-side terminal orphaned entries, pinning session keys until a cap-reset that may never fire. Now covered: **retry reassignment** (forget before the `SessionPrivKey` pointer overwrite), **settlement grace expiry**, **provider disconnect** (new `Provider.PendingSessionKeys()` snapshot taken *before* `Disconnect` wipes the map), and **all 38 `cancelDispatch` sites in dispatch.go** via a `cancelDispatchAndForget` wrapper.

**Zeroing policy** (deliberate, documented on the type): keys are zeroed only on read-loop-goroutine forgets (`handleComplete`, `handleInferenceError`, disconnect cleanup — same goroutine as chunk decryption, no concurrent use possible). Cross-goroutine forgets (retry, grace timer) delete without zeroing: zeroing there races an in-flight `box.OpenAfterPrecomputation`, and a corrupted decrypt would `MarkUntrusted` an innocent provider.

### 3. Overflow-499 telemetry classification
The #491 overflow abort was correctly classified for reputation but `postCommitProviderErrorOutcome` still booked it as `provider_error_after_commit` + `AdmittedButFailed=true` — polluting exactly the calibration signals the 499 was meant to protect. Now: shared `terminal_class.go` constants/helper (`statusClientClosedRequest`, `isConsumerCancelTerminal`) replace the magic literals in both `provider.go` and `route_outcome.go`, and consumer-cancel terminals short-circuit to `partial_success`/`consumer_cancel_after_commit`/`AdmittedButFailed=false`. Real 502s still book as provider faults (pinned by test).

### 4. Writer deadline/shutdown races
- **Watchdog TOCTOU**: `writeDeadline`'s value is now the CAS token (nonzero UnixNano → 0). `writeFrame` and the watchdog race to claim it; exactly one wins. A write that lost the claim returns `errProviderWriteTimeout` **even on nil err** (the socket is dead — fail over now, not after the TTFT budget). The sticky `writeTimedOut` latch is deleted; attribution is per-frame.
- **Stopped-vs-success**: when `req.done` and `w.done` were both ready, Go's select returned `errProviderWriterStopped` ~50% of the time for a frame that was on the wire — duplicate dispatch to a second provider. Now a final non-blocking `req.done` drain runs before the stopped fallback. Regression test proven to have teeth: reverting the fix fails 6/10 runs.

## Deferred (blocked on in-flight v0.6.31 files)

- `consumer.go:244/:717/:3305` — same refund-swallow fix
- `consumer.go` cancel/cleanup sites (`cancelDispatch` itself could absorb the forget once unfrozen; `cleanupPending` closures; generic-handler exits — full list in review notes)
- `registry.go` — routing `SendLoadModel`/`SendPrefetchModel`/`SendDesiredModels` through the control lane

## Verification

- `gofmt` clean, `go vet ./...` clean, **full `go test ./...` green** (api 104s)
- `go test ./registry/ -race -count=3` green; race-sensitive writer tests at `-race -count=50`
- New tests: refund-failure metric (counting fail-store), forgetAndZero/forget semantics, per-terminal-path lifecycle (incl. full WS-disconnect end-to-end and settlement expiry), 499 classification (overflow + bare 499 vs real 502), CAS exactly-one-winner (10k concurrent iterations), deadline-boundary never-reports-success-on-dead-socket, stopped-vs-success determinism

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/494"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785559209&installation_id=133247490&pr_number=494&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F494&signature=d1658ce987e35d3d6f84c276efbda157abcae78bc95aed05edf1fb057a20a74e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/api/chunk_key_cache.go
+++ b/coordinator/api/chunk_key_cache.go
@@ -1,8 +1,8 @@
 package api
 
 import (
+	"reflect"
 	"sync"
-	"unsafe"
 
 	"github.com/eigeninference/d-inference/coordinator/internal/e2e"
 )
@@ -110,7 +110,7 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
 	shared := e2e.PrecomputeSharedKey(&pub, priv)
 
 	c.mu.Lock()
-	if _, gone := c.dead[uintptr(unsafe.Pointer(priv))]; gone {
+	if _, gone := c.dead[tombstoneKey(priv)]; gone {
 		// The request hit a terminal path while the lock was dropped for the
 		// compute (or this is a late chunk of an already-forgotten request):
 		// hand the key back for this one decrypt but do NOT cache it — no
@@ -139,7 +139,7 @@ func (c *chunkKeyCache) markDeadLocked(priv *[32]byte) {
 	if c.dead == nil || len(c.dead) >= chunkKeyDeadMax {
 		c.dead = make(map[uintptr]struct{})
 	}
-	c.dead[uintptr(unsafe.Pointer(priv))] = struct{}{}
+	c.dead[tombstoneKey(priv)] = struct{}{}
 }
 
 // forget drops the cached shared key for a finished request WITHOUT zeroing
@@ -210,4 +210,12 @@ func (c *chunkKeyCache) forgetPeer(peerPub string) {
 		}
 	}
 	c.mu.Unlock()
+}
+
+// tombstoneKey converts a priv pointer to its non-pinning map identity. Uses
+// reflect rather than unsafe purely to avoid the unsafe import; the semantics
+// (address as identity, no GC pinning, benign ABA on address reuse — see the
+// dead field comment) are identical.
+func tombstoneKey(priv *[32]byte) uintptr {
+	return reflect.ValueOf(priv).Pointer()
 }
--- a/coordinator/api/chunk_key_cache.go
+++ b/coordinator/api/chunk_key_cache.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"sync"
+	"unsafe"
 
 	"github.com/eigeninference/d-inference/coordinator/internal/e2e"
 )
@@ -66,11 +67,24 @@ type chunkKeyCache struct {
 	m  map[*[32]byte]chunkKeyEntry
 	// dead tombstones privs whose request hit a terminal path (forget /
 	// forgetAndZero). sharedKey refuses to (re-)cache a tombstoned priv.
-	// Bounded by chunkKeyDeadMax with a wholesale drop, like m. Note the
-	// entries pin their 32-byte priv arrays from the GC until that drop —
-	// bounded and secret-free (the SHARED key is what the cache zeroes; the
-	// priv array lives on in its PendingRequest anyway).
-	dead map[*[32]byte]struct{}
+	// Bounded by chunkKeyDeadMax with a wholesale drop, like m.
+	//
+	// Keyed by the priv's ADDRESS (uintptr), not the pointer, so a tombstone
+	// never pins the 32-byte private-key array in the heap after its
+	// PendingRequest is gone — retaining dead key material would undermine
+	// the forget-on-terminal hygiene this cache exists for. The trade is a
+	// benign ABA: after the GC reclaims a tombstoned array, a future
+	// SessionPrivKey allocated at the same address is falsely treated as
+	// dead — its stream still decrypts correctly (sharedKey always returns a
+	// freshly computed key), it just recomputes per chunk until the wholesale
+	// drop clears the stale tombstone. Perf-only, rare, self-healing.
+	//
+	// Why a tombstone rather than refcounting or an owner-managed lifecycle:
+	// the racing readers (late chunks on the provider read loop) are
+	// fire-and-forget with no owner to release against, and the race window
+	// is milliseconds; a lock-only O(1) set that self-expires via the
+	// wholesale drop is the smallest mechanism that closes it.
+	dead map[uintptr]struct{}
 }
 
 type chunkKeyEntry struct {
@@ -96,7 +110,7 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
 	shared := e2e.PrecomputeSharedKey(&pub, priv)
 
 	c.mu.Lock()
-	if _, gone := c.dead[priv]; gone {
+	if _, gone := c.dead[uintptr(unsafe.Pointer(priv))]; gone {
 		// The request hit a terminal path while the lock was dropped for the
 		// compute (or this is a late chunk of an already-forgotten request):
 		// hand the key back for this one decrypt but do NOT cache it — no
@@ -123,9 +137,9 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
 // path can race a late decrypt (see the field comment).
 func (c *chunkKeyCache) markDeadLocked(priv *[32]byte) {
 	if c.dead == nil || len(c.dead) >= chunkKeyDeadMax {
-		c.dead = make(map[*[32]byte]struct{})
+		c.dead = make(map[uintptr]struct{})
 	}
-	c.dead[priv] = struct{}{}
+	c.dead[uintptr(unsafe.Pointer(priv))] = struct{}{}
 }
 
 // forget drops the cached shared key for a finished request WITHOUT zeroing
--- a/coordinator/api/chunk_key_cache.go
+++ b/coordinator/api/chunk_key_cache.go
@@ -23,6 +23,27 @@ const chunkKeyCacheMax = 8192
 // key change (paranoia: reconnect races) invalidates the entry instead of
 // decrypting with a stale shared key.
 //
+// Lifecycle / zeroing policy. Every request-terminal path must drop its entry
+// (orphans are otherwise pinned until the chunkKeyCacheMax wholesale reset),
+// but only SOME of those paths may zero the key material:
+//
+//   - forgetAndZero (delete + zero): ONLY from the provider read-loop
+//     goroutine that performs this request's chunk decryption —
+//     handleComplete, handleInferenceError, and providerReadLoop's disconnect
+//     cleanup. On that goroutine no decrypt with the key can be concurrently
+//     in flight, so zeroing is safe and scrubs the shared secret from the
+//     heap promptly.
+//   - forget (delete only, NO zeroing): every CROSS-GOROUTINE terminal —
+//     dispatch retry key reassignment, the settlement-grace expiry timer,
+//     dispatch-loop cancel/abandon sites. Zeroing there would race a
+//     possibly-in-flight box.OpenAfterPrecomputation on the read loop; a
+//     corrupted decrypt is reported as an invalid encrypted chunk and
+//     triggers MarkUntrusted against an innocent provider. The deleted
+//     array is left intact for the GC.
+//   - The wholesale cap-reset in sharedKey must NOT zero either: the dropped
+//     entries may belong to live streams that will recompute and keep
+//     decrypting with the same private key.
+//
 // The zero value is ready to use.
 type chunkKeyCache struct {
 	mu sync.Mutex
@@ -53,6 +74,10 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
 
 	c.mu.Lock()
 	if c.m == nil || len(c.m) >= chunkKeyCacheMax {
+		// Wholesale reset (abandoned-entry safety net). Deliberately does NOT
+		// zero the dropped keys: entries may belong to live streams whose read
+		// loops are decrypting with them right now (see the zeroing policy on
+		// the type comment).
 		c.m = make(map[*[32]byte]chunkKeyEntry)
 	}
 	c.m[priv] = chunkKeyEntry{peerPub: peerPub, shared: shared}
@@ -60,12 +85,39 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
 	return shared, nil
 }
 
-// forget drops the cached shared key for a finished request.
+// forget drops the cached shared key for a finished request WITHOUT zeroing
+// it. Safe to call from ANY goroutine — this is the variant every
+// cross-goroutine terminal path must use (dispatch retry reassignment,
+// settlement-grace timer, dispatch cancel/abandon): zeroing here could race an
+// in-flight decrypt on the provider read loop (see the type comment).
 func (c *chunkKeyCache) forget(priv *[32]byte) {
 	if priv == nil {
 		return
 	}
 	c.mu.Lock()
 	delete(c.m, priv)
+	c.mu.Unlock()
+}
+
+// forgetAndZero drops the cached shared key for a terminal request AND zeroes
+// the 32-byte array so the shared secret does not linger on the heap.
+//
+// MUST only be called from the goroutine that performs this request's chunk
+// decryption — the owning provider's read loop (handleComplete,
+// handleInferenceError, providerReadLoop's disconnect cleanup). On any other
+// goroutine the zeroing races a possibly-in-flight
+// box.OpenAfterPrecomputation and can corrupt a decrypt (which would
+// MarkUntrusted an innocent provider); use forget there instead.
+func (c *chunkKeyCache) forgetAndZero(priv *[32]byte) {
+	if priv == nil {
+		return
+	}
+	c.mu.Lock()
+	if e, ok := c.m[priv]; ok {
+		if e.shared != nil {
+			*e.shared = [32]byte{}
+		}
+		delete(c.m, priv)
+	}
 	c.mu.Unlock()
 }
--- a/coordinator/api/chunk_key_cache.go
+++ b/coordinator/api/chunk_key_cache.go
@@ -13,6 +13,13 @@ import (
 // X25519 recompute per live stream.
 const chunkKeyCacheMax = 8192
 
+// chunkKeyDeadMax bounds the tombstone set (chunkKeyCache.dead). One tombstone
+// is added per terminated request; when full the set is dropped wholesale. A
+// tombstone only needs to outlive the terminal path's race with a late
+// in-flight decrypt on the provider read loop (milliseconds), so losing old
+// tombstones to the reset is harmless in practice.
+const chunkKeyDeadMax = 8192
+
 // chunkKeyCache memoizes the NaCl box shared key per inference request so the
 // per-token chunk decrypt path performs only the symmetric open, not a fresh
 // X25519 scalar multiplication per chunk (~40-60µs each, serialized on the
@@ -44,10 +51,26 @@ const chunkKeyCacheMax = 8192
 //     entries may belong to live streams that will recompute and keep
 //     decrypting with the same private key.
 //
+// Both forget variants also TOMBSTONE the priv (see dead below): sharedKey
+// drops the lock during the X25519 compute, so without a tombstone a forget
+// racing that window (or a late chunk decrypted just after a cross-goroutine
+// forget) would re-insert an entry for a request no terminal path remains to
+// clean up, pinning the session key until the cap reset. A tombstoned priv is
+// never legitimately re-cached: every dispatch attempt generates fresh session
+// keys, so post-forget decrypts are only late stragglers of an abandoned
+// attempt — they still get a working key, they just recompute per chunk.
+//
 // The zero value is ready to use.
 type chunkKeyCache struct {
 	mu sync.Mutex
 	m  map[*[32]byte]chunkKeyEntry
+	// dead tombstones privs whose request hit a terminal path (forget /
+	// forgetAndZero). sharedKey refuses to (re-)cache a tombstoned priv.
+	// Bounded by chunkKeyDeadMax with a wholesale drop, like m. Note the
+	// entries pin their 32-byte priv arrays from the GC until that drop —
+	// bounded and secret-free (the SHARED key is what the cache zeroes; the
+	// priv array lives on in its PendingRequest anyway).
+	dead map[*[32]byte]struct{}
 }
 
 type chunkKeyEntry struct {
@@ -73,6 +96,15 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
 	shared := e2e.PrecomputeSharedKey(&pub, priv)
 
 	c.mu.Lock()
+	if _, gone := c.dead[priv]; gone {
+		// The request hit a terminal path while the lock was dropped for the
+		// compute (or this is a late chunk of an already-forgotten request):
+		// hand the key back for this one decrypt but do NOT cache it — no
+		// terminal path remains to forget the entry, so it would pin the
+		// session key until the cap reset.
+		c.mu.Unlock()
+		return shared, nil
+	}
 	if c.m == nil || len(c.m) >= chunkKeyCacheMax {
 		// Wholesale reset (abandoned-entry safety net). Deliberately does NOT
 		// zero the dropped keys: entries may belong to live streams whose read
@@ -85,22 +117,37 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
 	return shared, nil
 }
 
+// markDeadLocked tombstones priv so sharedKey never re-caches it. Caller must
+// hold c.mu. The set is dropped wholesale at chunkKeyDeadMax — a tombstone
+// only needs to outlive the milliseconds-scale window in which a terminal
+// path can race a late decrypt (see the field comment).
+func (c *chunkKeyCache) markDeadLocked(priv *[32]byte) {
+	if c.dead == nil || len(c.dead) >= chunkKeyDeadMax {
+		c.dead = make(map[*[32]byte]struct{})
+	}
+	c.dead[priv] = struct{}{}
+}
+
 // forget drops the cached shared key for a finished request WITHOUT zeroing
-// it. Safe to call from ANY goroutine — this is the variant every
-// cross-goroutine terminal path must use (dispatch retry reassignment,
-// settlement-grace timer, dispatch cancel/abandon): zeroing here could race an
-// in-flight decrypt on the provider read loop (see the type comment).
+// it, and tombstones the priv so a racing/late decrypt cannot re-cache it
+// (see the type comment). Safe to call from ANY goroutine — this is the
+// variant every cross-goroutine terminal path must use (dispatch retry
+// reassignment, settlement-grace timer, dispatch cancel/abandon): zeroing
+// here could race an in-flight decrypt on the provider read loop.
 func (c *chunkKeyCache) forget(priv *[32]byte) {
 	if priv == nil {
 		return
 	}
 	c.mu.Lock()
 	delete(c.m, priv)
+	c.markDeadLocked(priv)
 	c.mu.Unlock()
 }
 
 // forgetAndZero drops the cached shared key for a terminal request AND zeroes
-// the 32-byte array so the shared secret does not linger on the heap.
+// the 32-byte array so the shared secret does not linger on the heap. Like
+// forget, it tombstones the priv so a late decrypt cannot re-cache it (see
+// the type comment).
 //
 // MUST only be called from the goroutine that performs this request's chunk
 // decryption — the owning provider's read loop (handleComplete,
@@ -118,6 +165,35 @@ func (c *chunkKeyCache) forgetAndZero(priv *[32]byte) {
 			*e.shared = [32]byte{}
 		}
 		delete(c.m, priv)
+	}
+	c.markDeadLocked(priv)
+	c.mu.Unlock()
+}
+
+// forgetPeer drops every cache entry whose peer public key matches peerPub,
+// WITHOUT zeroing and WITHOUT tombstoning. It is the provider-disconnect
+// catch-all: registry-initiated disconnects (stale eviction, duplicate-serial
+// eviction, forced removal) call Registry.Disconnect directly, which wipes
+// the provider's pending map BEFORE the read-loop cleanup defer can snapshot
+// the session keys — leaving those requests' entries unreachable by any
+// per-request terminal. Sweeping by the provider's public key needs no
+// request state at all.
+//
+// No zeroing and no tombstones because the same peerPub can serve a NEWER
+// session of the same machine (duplicate-serial eviction: the replacement
+// connection reuses the keypair): zeroing could corrupt a live decrypt on the
+// new session's read loop, and a tombstone would permanently disable caching
+// for its live streams. A swept live entry simply recomputes on its next
+// chunk. O(len(m)) ≤ chunkKeyCacheMax under the lock, on rare disconnects.
+func (c *chunkKeyCache) forgetPeer(peerPub string) {
+	if peerPub == "" {
+		return
+	}
+	c.mu.Lock()
+	for priv, e := range c.m {
+		if e.peerPub == peerPub {
+			delete(c.m, priv)
+		}
 	}
 	c.mu.Unlock()
 }
--- a/coordinator/api/dispatch.go
+++ b/coordinator/api/dispatch.go
@@ -990,23 +990,6 @@ func (d *dispatchState) dispatchPrimary() dispatchOutcome {
 	return outcomeProceed
 }
 
-// cancelDispatchAndForget abandons a dispatched attempt (cancelDispatch:
-// remove pending, idle the provider, WS-cancel, refund the attempt's top-up)
-// and drops the attempt's memoized chunk-decryption key. Every dispatch-loop
-// abandon site (speculative losers, failover retries, deadline/client-gone
-// exits) funnels through here so an abandoned attempt cannot pin its session
-// key in the cache until the cap-reset. Plain forget, NO zeroing: these sites
-// run on the consumer/dispatch goroutine, not the provider read loop — the
-// read loop may still be decrypting a late in-flight chunk with this key, and
-// zeroing under it would corrupt that decrypt (and MarkUntrusted an innocent
-// provider). See chunkKeyCache's zeroing policy.
-func (s *Server) cancelDispatchAndForget(provider *registry.Provider, pr *registry.PendingRequest) {
-	s.cancelDispatch(provider, pr)
-	if pr != nil {
-		s.chunkKeys.forget(pr.SessionPrivKey)
-	}
-}
-
 // noteDispatchRetry feeds the inference-error breaker + refund for a pre-commit
 // provider error and, unless held boilerplate was discarded (which emits its own
 // pre-content failover counter), emits the generic retry counter. This is the
@@ -1175,7 +1158,7 @@ func (d *dispatchState) waitFirstChunk() (outcome dispatchOutcome) {
 				select {
 				case errMsg := <-pr.ErrorCh:
 					d.excludeProviders[provider.ID] = struct{}{}
-					s.cancelDispatchAndForget(provider, pr)
+					s.cancelDispatch(provider, pr)
 					d.setLastInferenceError(provider, errMsg)
 					d.lastFailedVersion = failedProviderVersion(provider)
 					d.noteDispatchRetry(provider, pr, errMsg.StatusCode, errMsg.Error, &d.heldChunks)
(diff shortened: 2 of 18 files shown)


ethenotethan (commented):

**Verdict:** COMMENT

### Security — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Performance — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Type_diligence — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Additive_complexity — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

✅ All four passes clean. No issues found.

coordinator/api/dispatch.go
```diff
@@ -920,6 +927,23 @@ func (d *dispatchState) dispatchPrimary() dispatchOutcome {
...
+// exits) funnels through here so an abandoned attempt cannot pin its session
+// key in the cache until the cap-reset. Plain forget, NO zeroing: these sites
+// run on the consumer/dispatch goroutine, not the provider read loop — the
+// read loop may still be decrypting a late in-flight chunk with this key, and
+// zeroing under it would corrupt that decrypt (and MarkUntrusted an innocent
+// provider). See chunkKeyCache's zeroing policy.
+func (s *Server) cancelDispatchAndForget(provider *registry.Provider, pr *registry.PendingRequest) {
+	s.cancelDispatch(provider, pr)
+	if pr != nil {
+		s.chunkKeys.forget(pr.SessionPrivKey)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in c94ed045. Went with the tombstone approach: `forget`/`forgetAndZero` now mark the priv pointer in a bounded tombstone set (`chunkKeyCache.dead`, capped at `chunkKeyDeadMax` with wholesale drop), and `sharedKey` re-checks it under the re-acquired lock before storing. A forget racing the unlocked X25519 compute window — or a late straggler chunk decrypted entirely after a cross-goroutine forget — now gets a working key back for that one decrypt without re-inserting an entry that no terminal path remains to clean up.

This is safe because a forgotten priv is never legitimately re-cached: every dispatch attempt generates fresh session keys (`dispatchPrimary` forgets the old pointer before overwriting it at dispatch.go:911-912), so post-forget decrypts are only stragglers of abandoned attempts — they recompute per chunk instead of caching. Tombstones hold no secret material and only need to outlive the ms-scale race window, so the wholesale cap drop is harmless.

Tests: `TestChunkKeyCacheForgetPreventsRecache`, `TestChunkKeyCacheForgetBeforeFirstUsePreventsCache` (cancel-before-first-chunk case), `TestChunkKeyCacheDeadTombstoneCapResets`; existing concurrent test passes under -race.

coordinator/api/provider.go
```diff
@@ -141,6 +158,13 @@ func (s *Server) providerReadLoop(ctx context.Context, conn *websocket.Conn, pro
 	loopCtx, loopCancel := context.WithCancel(ctx)
 	defer func() {
 		loopCancel()
+		// Drop the memoized chunk-decryption keys of every request still
+		// in-flight on this provider BEFORE Disconnect wipes the pending map
+		// (after the wipe the keys are unreachable and would pin their cache
+		// entries until the cap-reset). Zeroing is safe here: this defer runs
+		// on the provider's read-loop goroutine — the only goroutine that ever
+		// decrypts with these keys — after conn.Read has returned for good.
+		s.forgetProviderPendingKeys(provider)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in c94ed045, via the cache rather than the registry (registry.go is frozen on this branch for the in-flight v0.6.31 work, so hooking every direct `Registry.Disconnect` caller — evictStale, DisconnectDuplicatesBySerial, RemoveProviderBySerial(force) — wasn't an option here, and would also leave future callers unprotected).

Instead, `forgetProviderPendingKeys` now ends with a new `chunkKeys.forgetPeer(provider.PublicKey)` sweep that drops every cache entry stored under this provider's public key (the cache already stores peerPub per entry, so it needs no pending-map state at all). Since `Disconnect`'s `closeWriterNow` always unblocks `conn.Read`, the read-loop defer always runs — including for all three registry-initiated paths — and the sweep catches exactly the entries whose requests were wiped from `pendingReqs` before the per-request snapshot could see them.

The sweep is plain-delete only: no zeroing and no tombstoning, because the same peerPub can serve a newer same-keypair session of the same machine (duplicate-serial eviction) whose read loop may be decrypting with a matching entry right now — a swept live entry just recomputes on its next chunk. The per-request `forgetAndZero` snapshot stays first so the common read-loop-owned path still scrubs the shared secrets.

Test: `TestRegistryInitiatedDisconnectSweepsChunkKeys` (registers over WS with a public key, calls `reg.Disconnect` directly to wipe the pending map first, asserts the orphan is swept and NOT zeroed), plus `TestChunkKeyCacheForgetPeerSweepsWithoutZeroingOrTombstoning`.

coordinator/api/reservations.go
```diff
@@ -97,7 +97,18 @@ func (s *Server) releaseInitialReservation(accountID, model string, amount int64
...
+	// the ERROR log + billing.refund_failures metric is the alerting hook.
+	if err := s.store.Credit(accountID, amount, store.LedgerRefund, "reservation_refund"); err != nil {
+		s.logger.Error("failed to credit reservation refund to consumer",
+			"account_id", accountID,
+			"model", model,
+			"refund_micro_usd", amount,
+			"error", err,
+		)
+		s.ddIncr("billing.refund_failures", tags)
+	}
```
ethenotethan, line 115: 🔵 **[INFO]** 🔒 Refund failure logging exposes account_id in error logs

💡 _Suggestion:_ Consider redacting or hashing account_id in error logs to prevent potential PII exposure

coordinator/api/chunk_key_cache.go
```diff
@@ -52,20 +96,104 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
...
 	if c.m == nil || len(c.m) >= chunkKeyCacheMax {
+		// Wholesale reset (abandoned-entry safety net). Deliberately does NOT
+		// zero the dropped keys: entries may belong to live streams whose read
+		// loops are decrypting with them right now (see the zeroing policy on
+		// the type comment).
 		c.m = make(map[*[32]byte]chunkKeyEntry)
 	}
 	c.m[priv] = chunkKeyEntry{peerPub: peerPub, shared: shared}
 	c.mu.Unlock()
 	return shared, nil
```
ethenotethan, line 131: 🟡 **[MEDIUM]** ⚡ X25519 computation performed while holding mutex lock

💡 _Suggestion:_ Move the e2e.PrecomputeSharedKey call outside the mutex-protected section to avoid blocking other cache operations during the ~40-60µs scalar multiplication

📊 _Score: 3×4 = 12 · Category: blocking_io_

coordinator/api/chunk_key_cache.go
```diff
@@ -23,10 +30,47 @@ const chunkKeyCacheMax = 8192
...
+//   - forget (delete only, NO zeroing): every CROSS-GOROUTINE terminal —
+//     dispatch retry key reassignment, the settlement-grace expiry timer,
+//     dispatch-loop cancel/abandon sites. Zeroing there would race a
+//     possibly-in-flight box.OpenAfterPrecomputation on the read loop; a
+//     corrupted decrypt is reported as an invalid encrypted chunk and
+//     triggers MarkUntrusted against an innocent provider. The deleted
+//     array is left intact for the GC.
+//   - The wholesale cap-reset in sharedKey must NOT zero either: the dropped
+//     entries may belong to live streams that will recompute and keep
+//     decrypting with the same private key.
```
ethenotethan, line 53: 🟡 **[MEDIUM]** 🧩 Complex lifecycle policy with multiple forget variants creates cognitive overhead

💡 _Suggestion:_ Consider consolidating the forget/forgetAndZero distinction into a single method with a boolean parameter, or use a clearer naming convention that makes the safety constraints obvious at call sites

coordinator/api/chunk_key_cache.go
ethenotethan, line 131: 🔵 **[INFO]** 🧩 Tombstone mechanism adds state tracking complexity for race protection

💡 _Suggestion:_ Document whether the tombstone approach is simpler than alternative solutions like reference counting or explicit lifecycle management

📊 _Score: 2×3 = 6 · Category: additive complexity_

coordinator/api/dispatch.go
```diff
@@ -920,6 +927,23 @@ func (d *dispatchState) dispatchPrimary() dispatchOutcome {
...
+	}
+}
```
ethenotethan: 🔵 **[INFO]** 🧩 New cancelDispatchAndForget wrapper adds indirection layer

💡 _Suggestion:_ Consider whether the chunk key cleanup could be integrated directly into cancelDispatch rather than requiring a separate wrapper function

📊 _Score: 2×2 = 4 · Category: additive complexity_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 1 finding(s)
- 🔵 **[INFO]** `coordinator/api/reservations.go:100-111` — Refund failure logging exposes account_id in error logs
  - _Suggestion:_ Consider redacting or hashing account_id in error logs to prevent potential PII exposure

- 🟡 **[MEDIUM]** `coordinator/api/chunk_key_cache.go:96-117` — X25519 computation performed while holding mutex lock
  - _Suggestion:_ Move the e2e.PrecomputeSharedKey call outside the mutex-protected section to avoid blocking other cache operations during the ~40-60µs scalar multiplication

### Additive_complexity — 3 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/chunk_key_cache.go:30-52` — Complex lifecycle policy with multiple forget variants creates cognitive overhead
  - _Suggestion:_ Consider consolidating the forget/forgetAndZero distinction into a single method with a boolean parameter, or use a clearer naming convention that makes the safety constraints obvious at call sites
- 🔵 **[INFO]** `coordinator/api/chunk_key_cache.go:96-117` — Tombstone mechanism adds state tracking complexity for race protection
  - _Suggestion:_ Document whether the tombstone approach is simpler than alternative solutions like reference counting or explicit lifecycle management
- 🔵 **[INFO]** `coordinator/api/dispatch.go:928-945` — New cancelDispatchAndForget wrapper adds indirection layer
  - _Suggestion:_ Consider whether the chunk key cleanup could be integrated directly into cancelDispatch rather than requiring a separate wrapper function

_5 finding(s) total, 2 blocking. Verdict: REQUEST_CHANGES._

coordinator/api/settlement.go
```diff
@@ -101,6 +104,14 @@ func (s *Server) holdForSettlement(pr *registry.PendingRequest) {
...
 	}
 	s.settlements.hold(pr, s.terminalSettleGrace(), func(expired *registry.PendingRequest) {
+		// Grace expired with no provider terminal: the request is abandoned
+		// for good, so drop its memoized chunk-decryption key. This runs on
+		// the AfterFunc timer goroutine — plain forget, NO zeroing, which
+		// would race a late chunk decrypt on the provider read loop (see
+		// chunkKeyCache's zeroing policy). Unconditional: even when the
+		// refund below no-ops (already-finalized dup park), no terminal
+		// handler ever claimed this record to forget the key.
+		s.chunkKeys.forget(expired.SessionPrivKey)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in c94ed045. `holdForSettlement`'s `IsReservationFinalized()` early return now does `s.chunkKeys.forget(pr.SessionPrivKey)` before returning: as you noted, the relay's stream-timeout/provider-incomplete branches finalize via `refundReservedBalance` without a read-loop terminal ever running, and since the record is never parked, no late `handleComplete`/`handleInferenceError` can claim it to drop the key. For records that WERE settled by a read-loop terminal first, the terminal's `forgetAndZero` already ran and this forget is a no-op. Plain forget (no zeroing) since this runs on the consumer goroutine while a late chunk decrypt may be in flight on the provider read loop.

Combined with the tombstone change in the same commit (see the dispatch.go thread), the forget also prevents a late chunk decrypted between this cleanup and `RemovePending` from re-caching the key.

Test: `TestHoldForSettlementFinalizedForgetsChunkKey` — finalized record's key is forgotten immediately, not zeroed, and the record is still not parked/refunded.

coordinator/api/chunk_key_cache.go
```diff
@@ -23,10 +30,47 @@ const chunkKeyCacheMax = 8192
...
 type chunkKeyCache struct {
 	mu sync.Mutex
 	m  map[*[32]byte]chunkKeyEntry
+	// dead tombstones privs whose request hit a terminal path (forget /
+	// forgetAndZero). sharedKey refuses to (re-)cache a tombstoned priv.
+	// Bounded by chunkKeyDeadMax with a wholesale drop, like m. Note the
+	// entries pin their 32-byte priv arrays from the GC until that drop —
+	// bounded and secret-free (the SHARED key is what the cache zeroes; the
+	// priv array lives on in its PendingRequest anyway).
+	dead map[*[32]byte]struct{}
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in f008eee3 — tombstones are now keyed by the priv's address (`uintptr`), so they carry identity without pinning the 32-byte array; dead requests' key material is GC-collectable immediately. The documented trade is a benign ABA (a reused address suppresses caching for the new stream until the wholesale drop — correctness unaffected since `sharedKey` always returns a freshly computed key). The cap-reset test itself demonstrated the effect: its fillers now need `runtime.KeepAlive` to reach the cap.

coordinator/api/chunk_key_cache.go
Gajesh2007: The X25519 compute already runs with the mutex dropped — `sharedKey` unlocks before `e2e.PrecomputeSharedKey` and re-locks only to insert (chunk_key_cache.go: unlock at the miss, compute, re-lock). That lock-drop window is precisely why the tombstone re-check under the re-acquired lock exists. The section under the lock is map ops only.

coordinator/api/chunk_key_cache.go
Gajesh2007: Keeping the two named variants deliberately: the split IS the safety contract — `forgetAndZero` may only run on the owning provider read-loop goroutine (concurrent zeroing races an in-flight `box.OpenAfterPrecomputation` and would MarkUntrusted an innocent provider), while `forget` is safe anywhere. A `forget(priv, zero bool)` hides that constraint at call sites; the method name makes every call site auditable at a glance. The full policy is documented on the type.

coordinator/api/chunk_key_cache.go
Gajesh2007: Documented in f008eee3 (on the `dead` field): the racing readers are fire-and-forget late chunks on the provider read loop with no owner to release against, so refcounting has nothing to hang off; the race window is milliseconds, and a lock-only O(1) set that self-expires via the wholesale drop is the smallest mechanism that closes it.

coordinator/api/dispatch.go
Gajesh2007: Agreed — folding the forget into `cancelDispatch` itself is the end state, but `cancelDispatch` lives in consumer.go, which carries in-flight v0.6.31 work on another branch (frozen for this PR series). The wrapper keeps all 38 dispatch.go sites covered now; absorbing it into `cancelDispatch` is on the documented post-v0.6.31 follow-up list (PR description, Deferred section).

coordinator/api/reservations.go
Gajesh2007: Account IDs here are opaque internal identifiers (Privy DIDs / key-hash labels), not direct PII, and logging them on billing paths matches established practice elsewhere in the coordinator (e.g. provider-account linking logs). Operationally the raw ID is required: a refund-failure alert is only actionable if on-call can locate the account to credit it manually. Happy to revisit if we adopt a repo-wide ID-redaction policy.

coordinator/api/chunk_key_cache.go
ethenotethan: :white_check_mark: Looks addressed in `f008eee` — Extensive documentation was added explaining why tombstones are simpler than reference counting or explicit lifecycle management for this specific race condition scenario.. Resolving.

coordinator/api/chunk_key_cache.go
```diff
@@ -2,6 +2,7 @@ package api
 
 import (
 	"sync"
```
ethenotethan, line 5: 🔵 **[INFO]** 🔒 Import of unsafe package for address-based tombstoning

💡 _Suggestion:_ Consider if the ABA issue described in comments could be avoided with a different design that doesn't require unsafe pointer arithmetic

📊 _Score: 2×2 = 4 · Category: unsafe-usage_

coordinator/api/chunk_key_cache.go
```diff
@@ -52,20 +110,104 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
...
 	if c.m == nil || len(c.m) >= chunkKeyCacheMax {
+		// Wholesale reset (abandoned-entry safety net). Deliberately does NOT
+		// zero the dropped keys: entries may belong to live streams whose read
+		// loops are decrypting with them right now (see the zeroing policy on
+		// the type comment).
 		c.m = make(map[*[32]byte]chunkKeyEntry)
 	}
 	c.m[priv] = chunkKeyEntry{peerPub: peerPub, shared: shared}
 	c.mu.Unlock()
 	return shared, nil
```
ethenotethan, line 131: 🟡 **[MEDIUM]** ⚡ X25519 computation performed under mutex lock

💡 _Suggestion:_ Move the e2e.PrecomputeSharedKey call outside the lock, then reacquire the lock to check tombstones and update the cache. This prevents blocking all cache operations during the ~40-60µs X25519 computation.

📊 _Score: 3×4 = 12 · Category: blocking_io_

coordinator/api/chunk_key_cache.go
```diff
@@ -52,20 +110,104 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
...
+	if priv == nil {
+		return
+	}
+	c.mu.Lock()
+	if e, ok := c.m[priv]; ok {
+		if e.shared != nil {
+			*e.shared = [32]byte{}
+		}
+		delete(c.m, priv)
+	}
```
ethenotethan, line 182: 🔵 **[INFO]** ⚡ forgetPeer iterates over all cache entries under lock

💡 _Suggestion:_ Consider adding early termination if the number of entries to check exceeds a threshold, or document that this is acceptable given the chunkKeyCacheMax bound of 8192 entries.

📊 _Score: 2×3 = 6 · Category: unbounded_loop_

coordinator/api/chunk_key_cache.go
ethenotethan, line 5: 🔵 **[INFO]** 🏷️ unsafe package imported but only used for pointer-to-uintptr conversion

💡 _Suggestion:_ Consider using reflect.ValueOf(priv).Pointer() instead of unsafe.Pointer for the tombstone key to avoid unsafe package dependency

📊 _Score: 2×1 = 2 · Category: unsafe_import_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 1 finding(s)
- 🔵 **[INFO]** `coordinator/api/chunk_key_cache.go:4` — Import of unsafe package for address-based tombstoning
  - _Suggestion:_ Consider if the ABA issue described in comments could be avoided with a different design that doesn't require unsafe pointer arithmetic

### Performance — 2 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/chunk_key_cache.go:110-131` — X25519 computation performed under mutex lock
  - _Suggestion:_ Move the e2e.PrecomputeSharedKey call outside the lock, then reacquire the lock to check tombstones and update the cache. This prevents blocking all cache operations during the ~40-60µs X25519 computation.
- 🔵 **[INFO]** `coordinator/api/chunk_key_cache.go:172-182` — forgetPeer iterates over all cache entries under lock
  - _Suggestion:_ Consider adding early termination if the number of entries to check exceeds a threshold, or document that this is acceptable given the chunkKeyCacheMax bound of 8192 entries.

### Type_diligence — 1 finding(s)
- 🔵 **[INFO]** `coordinator/api/chunk_key_cache.go:4` — unsafe package imported but only used for pointer-to-uintptr conversion
  - _Suggestion:_ Consider using reflect.ValueOf(priv).Pointer() instead of unsafe.Pointer for the tombstone key to avoid unsafe package dependency

_4 finding(s) total, 1 blocking. Verdict: REQUEST_CHANGES._

coordinator/api/dispatch.go
```diff
@@ -920,6 +927,23 @@ func (d *dispatchState) dispatchPrimary() dispatchOutcome {
 	return outcomeProceed
 }
 
+// cancelDispatchAndForget abandons a dispatched attempt (cancelDispatch:
+// remove pending, idle the provider, WS-cancel, refund the attempt's top-up)
+// and drops the attempt's memoized chunk-decryption key. Every dispatch-loop
+// abandon site (speculative losers, failover retries, deadline/client-gone
+// exits) funnels through here so an abandoned attempt cannot pin its session
+// key in the cache until the cap-reset. Plain forget, NO zeroing: these sites
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 6779cb55 — consumer.go is no longer frozen, so this closed properly: the forget moved INTO cancelDispatch itself (wrapper deleted), and every direct RemovePending abort site — including the generic /v1/completions and /v1/messages pre-commit timeout/client-gone/ErrorCh exits you flagged — now forgets via the RemovePending return value (13 sites converted). No abandoned attempt can pin its session key until the cap reset anymore.

coordinator/api/reservations.go
```diff
@@ -97,7 +97,18 @@ func (s *Server) releaseInitialReservation(accountID, model string, amount int64
...
+	// No inline retry — this path can run on latency-sensitive request paths;
+	// the ERROR log + billing.refund_failures metric is the alerting hook.
+	if err := s.store.Credit(accountID, amount, store.LedgerRefund, "reservation_refund"); err != nil {
+		s.logger.Error("failed to credit reservation refund to consumer",
+			"account_id", accountID,
+			"model", model,
+			"refund_micro_usd", amount,
+			"error", err,
+		)
+		s.ddIncr("billing.refund_failures", tags)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 6779cb55 — the failure branch now returns before billing.reservation_refunds/reservation_releases, so refund dashboards only count actual ledger credits. Same treatment applied to all three reservation_extra_refund sites in consumer.go (also newly unfrozen), which additionally skip the ReservedMicroUSD reset on failure so a later settlement path can retry the refund.

coordinator/api/chunk_key_cache.go
Gajesh2007: The compute is not under the mutex — sharedKey unlocks before e2e.PrecomputeSharedKey and re-acquires only for the tombstone check + map insert (chunk_key_cache.go: Unlock precedes the compute, Lock follows it). That lock-drop window is precisely why the tombstone re-check exists. This is the second time this finding has been filed; happy to walk through the locking if the tool keeps flagging it.

coordinator/api/chunk_key_cache.go
Gajesh2007: Addressed in 6779cb55 — the unsafe import is gone: the tombstone key is now derived via reflect.ValueOf(priv).Pointer() with identical non-pinning semantics. On the broader design question: a pointer-keyed (pinning) map was the alternative without ABA, and it's exactly what the previous round flagged as retaining dead key material — the benign, documented ABA is the deliberate trade for immediate GC of key material.

Gajesh2007: Done in 6779cb55 — switched to reflect.ValueOf(priv).Pointer(); unsafe import removed.

coordinator/api/chunk_key_cache.go
Gajesh2007: This is already documented on forgetPeer: 'O(len(m)) ≤ chunkKeyCacheMax under the lock, on rare disconnects' — the map is hard-capped at 8192 entries and the sweep runs only on provider disconnect, so the worst case is a bounded, rare, in-memory scan. Early termination isn't possible: multiple requests can share the peer key, so the sweep must visit every entry.

coordinator/api/chunk_key_cache.go
ethenotethan: :white_check_mark: Looks addressed in `6779cb5` — X25519 computation is now performed outside the mutex lock with proper tombstone checking after reacquiring the lock.. Resolving.

coordinator/api/chunk_key_cache.go
```diff
@@ -52,20 +110,112 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
...
 	if c.m == nil || len(c.m) >= chunkKeyCacheMax {
+		// Wholesale reset (abandoned-entry safety net). Deliberately does NOT
+		// zero the dropped keys: entries may belong to live streams whose read
+		// loops are decrypting with them right now (see the zeroing policy on
+		// the type comment).
 		c.m = make(map[*[32]byte]chunkKeyEntry)
 	}
 	c.m[priv] = chunkKeyEntry{peerPub: peerPub, shared: shared}
 	c.mu.Unlock()
 	return shared, nil
```
ethenotethan, line 131: 🟡 **[MEDIUM]** ⚡ X25519 computation (e2e.PrecomputeSharedKey) performed while holding mutex lock

💡 _Suggestion:_ Move the X25519 computation outside the critical section. Acquire lock, check cache/tombstone, release lock, compute key, re-acquire lock to store result.

📊 _Score: 3×4 = 12 · Category: blocking_io_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🟡 **[MEDIUM]** `coordinator/api/chunk_key_cache.go:110-131` — X25519 computation (e2e.PrecomputeSharedKey) performed while holding mutex lock
  - _Suggestion:_ Move the X25519 computation outside the critical section. Acquire lock, check cache/tombstone, release lock, compute key, re-acquire lock to store result.

_1 finding(s) total, 1 blocking. Verdict: REQUEST_CHANGES._

libs/mlx-swift-lm
```diff
@@ -1 +1 @@
-Subproject commit a71c8018c02c5da6097bfd28c35c2b162c6fe5c6
+Subproject commit e5d0a94460d3ee10316a93690a87ea1a802c6b89
```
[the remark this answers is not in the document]
Gajesh2007: Good catch — fixed in c5effc44. The rebase worktree's submodule was still checked out at the old master pin and a broad `git add` staged the stale gitlink, silently downgrading the engine from the v0.7.x fused-cache pin (a71c8018) to v0.6.30-era e5d0a944. Restored to master's exact pin; this PR is coordinator-only again (verified: `git diff origin/master...HEAD -- libs/` is now empty).

coordinator/api/chunk_key_cache.go
Gajesh2007: This finding contradicts its own resolution: the same tool resolved the identical claim 21 seconds earlier on this same line ('Looks addressed in 6779cb5 — X25519 computation is now performed outside the mutex lock with proper tombstone checking after reacquiring the lock. Resolving.') and then re-filed it. The code does exactly what the suggestion asks — lock, check cache, unlock, compute, re-lock, tombstone-check, store — and has since the original design. No change needed; third time this has been filed.

coordinator/api/chunk_key_cache.go
ethenotethan: :white_check_mark: Looks addressed in `c5effc4` — The unsafe package import has been removed and replaced with reflect package usage for pointer-to-uintptr conversion. Resolving.

coordinator/api/chunk_key_cache.go
```diff
@@ -23,10 +31,60 @@ const chunkKeyCacheMax = 8192
...
+	//
+	// Why a tombstone rather than refcounting or an owner-managed lifecycle:
+	// the racing readers (late chunks on the provider read loop) are
+	// fire-and-forget with no owner to release against, and the race window
+	// is milliseconds; a lock-only O(1) set that self-expires via the
+	// wholesale drop is the smallest mechanism that closes it.
+	dead map[uintptr]struct{}
 }
 
 type chunkKeyEntry struct {
```
ethenotethan, line 90: 🔵 **[INFO]** 🔒 reflect.ValueOf().Pointer() used for memory address extraction

💡 _Suggestion:_ Consider using unsafe.Pointer() with proper documentation of the ABA safety guarantees, or add explicit comments about the security implications of using reflection for pointer arithmetic

📊 _Score: 2×3 = 6 · Category: unsafe-deserialization_

coordinator/api/chunk_key_cache.go
```diff
@@ -52,20 +110,112 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
...
 	c.mu.Lock()
+	if _, gone := c.dead[tombstoneKey(priv)]; gone {
+		// The request hit a terminal path while the lock was dropped for the
+		// compute (or this is a late chunk of an already-forgotten request):
+		// hand the key back for this one decrypt but do NOT cache it — no
+		// terminal path remains to forget the entry, so it would pin the
+		// session key until the cap reset.
+		c.mu.Unlock()
+		return shared, nil
+	}
```
ethenotethan, line 121: 🟡 **[MEDIUM]** ⚡ X25519 computation performed while holding mutex lock

💡 _Suggestion:_ Move the e2e.PrecomputeSharedKey call outside the critical section to avoid blocking other cache operations during the ~40-60µs computation

📊 _Score: 3×4 = 12 · Category: blocking_io_

coordinator/api/chunk_key_cache.go
```diff
@@ -52,20 +110,112 @@ func (c *chunkKeyCache) sharedKey(priv *[32]byte, peerPub string) (*[32]byte, er
...
+// MarkUntrusted an innocent provider); use forget there instead.
+func (c *chunkKeyCache) forgetAndZero(priv *[32]byte) {
+	if priv == nil {
+		return
+	}
+	c.mu.Lock()
+	if e, ok := c.m[priv]; ok {
+		if e.shared != nil {
+			*e.shared = [32]byte{}
+		}
```
ethenotethan, line 180: 🔵 **[INFO]** ⚡ O(n) iteration over cache entries under mutex lock

💡 _Suggestion:_ Consider batching the deletion or using a more efficient data structure if this becomes a bottleneck with large cache sizes

📊 _Score: 2×3 = 6 · Category: unbounded_iteration_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 1 finding(s)
- 🔵 **[INFO]** `coordinator/api/chunk_key_cache.go:89-90` — reflect.ValueOf().Pointer() used for memory address extraction
  - _Suggestion:_ Consider using unsafe.Pointer() with proper documentation of the ABA safety guarantees, or add explicit comments about the security implications of using reflection for pointer arithmetic

### Performance — 2 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/chunk_key_cache.go:110-121` — X25519 computation performed while holding mutex lock
  - _Suggestion:_ Move the e2e.PrecomputeSharedKey call outside the critical section to avoid blocking other cache operations during the ~40-60µs computation
- 🔵 **[INFO]** `coordinator/api/chunk_key_cache.go:172-180` — O(n) iteration over cache entries under mutex lock
  - _Suggestion:_ Consider batching the deletion or using a more efficient data structure if this becomes a bottleneck with large cache sizes

PR #507, opened 2026-07-03
fix(coordinator): capacity-reject routing cooldown — stop black-hole providers absorbing dispatch

Gajesh2007:
# fix(coordinator): capacity-reject routing cooldown — stop black-hole providers absorbing dispatch

## Incident (live)

7 provider boxes (64GB, `gemma-4-26b-8bit`, v0.7.2) reject **100% of dispatched requests** with the capacity string (`token_budget_exhausted`, classified 503) from their **first request after registration** — **~9,000 rejections in 30 minutes, zero successes** — while **~170 healthy boxes** for the same model sit underutilized.

The router keeps selecting them because:

1. Their heartbeats report near-zero active tokens, so the cost-based scheduler (`registry/scheduler.go`) scores them best — they look like ideal instant-TTFT targets.
2. Capacity-class failures are **deliberately** excluded from reputation penalties, the shape-keyed 5xx breaker (`registry/error_cooldown.go` skips 503), and the node-health breaker (`registry/provider_breaker.go` ignores capacity sheds). Sound policy for occasional capacity rejects from a busy box; catastrophic for a box that rejects *everything*. Nothing in the system could ever deroute them.

## Fix

A **capacity-reject routing cooldown** (`registry/capacity_cooldown.go`), distinct from the 5xx error cooldown, keyed per `(provider, model)`:

- **Discriminator — zero interleaved accepts.** Trips on **>= N capacity rejects within window W with NO accept in between** (defaults N=5, W=60s). Any accept — the **first content-bearing chunk** (`commitFirstContent`, so a long generation vouches for a busy box *while* it sheds concurrent dispatches) or a **clean completion** (`noteInferenceSuccess`) — resets the streak, any active cooldown, and the backoff. A busy-but-serving box **can never trip**; a black hole (zero accepts by definition) trips within seconds.
- **Cooldown + re-probe.** Trip quarantines the pair for 120s (env-tunable) via the shared routing gate `providerPassesRoutingGatesLockedEx` — honored by both the dispatch hot path and the `QuickCapacityCheck` preflight, so they cannot drift. After expiry the pair is re-probed; a genuinely-full box that recovered gets traffic back and its first accept clears all state. A still-rejecting pair re-arms on its **first** post-expiry reject (half-open) with exponential backoff: 120s → 240s → 480s → capped at 10 min.
[...]

PR #510, opened 2026-07-03
fix(coordinator): capacity-cooldown follow-ups — half-open probe, generic-path accept, 404 strikes, all-cooled 429

Gajesh2007:
Fast-follow for the four P2 review findings on #507 (capacity-reject cooldown). Coordinator-only; no protocol changes.

1. **Generic-path accept** (`api/consumer.go`): `/v1/completions` and `/v1/messages` now record the capacity-cooldown ACCEPT at first content (both pre-accept and post-accept commit branches), matching the chat path. A long generic stream on a busy box keeps vouching for the (provider, model) pair while it sheds concurrent dispatches — the shed storm can no longer read as the zero-accepts black-hole signature.
2. **True half-open re-probe** (`registry/capacity_cooldown.go`, `registry/scheduler.go`): cooldown entries are now `{expiry, probeAt}` structs. After expiry, exactly ONE reservation claims the probe (under the registry write lock); the gate stays closed for everyone else until the probe's outcome lands — accept clears, reject re-arms with doubled TTL. Stale claims expire after 30s so a lost probe can't wedge the pair. No thundering herd through the post-expiry window.
3. **Cold 404 "model not loaded" strikes** (`api/consumer.go`): 404 joins 429/5xx as strike-eligible, guarded by the zero-interleaved-accepts discriminator — the normal 404→load→serve lifecycle never trips (the accept resets); only a box that 404s forever with zero accepts does. Request-shape 404s ("model not found") never strike.
4. **All-cooled surfaces as capacity** (`registry/scheduler.go`): the candidate scan and `QuickCapacityCheck` preflight count a pair blocked only by the capacity cooldown as a TRANSIENT capacityRejection (via an `ignoreCapacityCooldown` re-check), so an all-cooled model returns **429 + Retry-After** (or queues) instead of a structural "no providers" **503**.

Tests: probe-claim lifecycle, 32-way concurrent single-probe reservation race, generic-path accept commit on both branches, 404 strike discriminator, all-cooled 429 surfacing. `go build ./... && go test ./registry/... ./api/...` green on top of v0.7.3 master.

## Before / After

```mermaid
flowchart TB
  subgraph Before
    A1[cooldown expires] --> B1[gate fully opens]
    B1 --> C1[ALL waiting dispatches rush the box]
    C1 --> D1[box still sick: mass rejects, cooldown re-arms after damage]
[...]

PR #498, opened 2026-07-02
fix(billing): Stripe payouts — recipient agreement for non-US/EEA countries, auto-payout schedule, dead-account recovery

Gajesh2007:
Fixes the three Stripe payout issue clusters reported by providers in `#support`:

1. **AU/NZ/JP withdrawals hard-fail**: `Funds can't be sent to accounts located in XX when the account is under the 'full' service agreement` (mohfarsi9, samuelin0xcc, jukeizunosekkeisya050).
2. **Funds stranded in Stripe**: users see *"Contact Eigen Labs, Inc. to get paid out or to update your payout schedule"* with money sitting in their connected balance for weeks (fdpgrvmm5r €10.57 for 18+ days, chiodinostefano ×2).
3. **Stale `acct_…` wedges the account**: after a user closes their Stripe account, every call fails with `No such destination` forever; no unlink existed (samuelin0xcc, geoffrey, youri).

## Root causes (from code, verified against Stripe docs)

| # | Root cause | Where |
|---|-----------|-------|
| 1 | Every Express account was created under the default **`full` service agreement** with `card_payments`. Stripe only allows platform transfers under `full` within US/CA/UK/EEA/CH; everywhere else requires the **`recipient`** agreement (transfers-only). The agreement is immutable after acceptance. | `billing/stripe_connect.go` |
| 2a | Accounts were created with **`settings[payouts][schedule][interval]=manual`** — Stripe never sweeps the balance out. | `billing/stripe_connect.go` |
| 2b | The withdraw fallback comment claimed "Stripe's default daily auto-payout will move them" — false under 2a — and `payouts.create` was **hardcoded to USD**, which fails for any non-USD bank (EUR users), stranding every cross-currency withdrawal at `transferred`. No reconciliation existed. | `api/stripe_payouts.go` |
| 3 | No detection of `account_invalid` / `No such destination`; onboarding only recreated accounts on country change; no unlink endpoint. | `api/stripe_payouts.go` |

## Behavior: before → after

```mermaid
flowchart TB
  subgraph Before
    A1[POST /withdraw/stripe] --> B1[debit ledger]
    B1 --> C1[transfers.create]
    C1 -- "AU/NZ/JP: full-SA error" --> D1[refund + raw Stripe error<br/>user permanently blocked]
    C1 -- ok --> E1["payouts.create (hardcoded USD)"]
[...]

PR #512, opened 2026-07-04
fix(coordinator): seven routing fixes from prod-DB analysis — TTFT gate calibration, reconnect-proof breakers, dedicated queueing, decode floor, servability

Gajesh2007:
## Summary

Seven routing fixes derived from a full production-DB analysis of Jun 27 – Jul 3 (2.73M routed requests, `inference_routes` / `request_rejections` / `provider_sessions`). The week's dominant losses were self-inflicted by the router: a deterministic TTFT rejection burned the full 64-attempt failover ladder (1.89M junk telemetry rows, 28% of the table), the TTFT estimator over-predicted 2–3× behind a live hard-reject gate (883k rejection rows, 100% `could_have_served`), fault breakers were session-keyed so reconnecting black holes escaped ejection (one box absorbed 13,333 requests at 100% error), dedicated-pool overflow fast-429'd instead of queueing (283k `machine_busy`/week vs 290 queued), 2.0× overcommit diluted gemma decode to p50 8 tok/s against a 15 tok/s floor, and the coordinator admitted ~28k requests/week that providers deterministically rejected on token budget.

| # | Fix | Key change |
|---|---|---|
| 1 | TTFT ladder termination | Deterministic `ttft_too_slow` reservation rejection ends the ladder on any attempt (flag `EIGENINFERENCE_TTFT_TERMINAL_REJECT`, default on) |
| 2 | TTFT estimator calibration | Online per-(model, chip-family) windowed-median actual/predicted ratio, clamp [0.2, 1.5], 50-obs warm-up, kill switch `EIGENINFERENCE_TTFT_CALIBRATION=off`, gauge `routing.ttft_calibration_ratio` |
| 3 | Reconnect-proof breakers | All fault state keyed by stable identity (serial → SE-key → account); survives disconnect; new zero-success capacity streak ejects black holes ≤10 failures; #507 pair cooldown also re-keyed |
| 4 | Dedicated-pool queueing | gemma overflow queues instead of instant 429; queue depth/wait env-tunable (`EIGENINFERENCE_QUEUE_MAX_DEPTH`=32, `EIGENINFERENCE_QUEUE_MAX_WAIT`=120s) |
| 5 | Decode-floor enforcement | Quality-concurrency overcommit 2.0 → 1.2 (+ `EIGENINFERENCE_QUALITY_CONCURRENCY_OVERCOMMIT_BY_MODEL` per-model overrides); warm-pool saturation counts against the quality cap |
| 6 | Authoritative servability | `providerBudgetFits` mirrors provider-side admission (post-load KV budget, padded weights, activation reserve); wired into `freeMemoryAdmits` cold path; gate default ON |
| 7 | Reconnect-churn RCA | Root cause: macOS sleep → 90s eviction sweep (65% of closes); real socket outcome now stamped in `disconnect_reason` (`ws_close_<code>`/`read_error`/`oom_suspected`); report in `docs/reports/2026-07-03-reconnect-churn-root-cause.md` |

## Behavior: before → after

```mermaid
flowchart LR
  subgraph Before
    A1[request] --> B1{TTFT gate<br/>estimate 2-3x actual}
    B1 -->|over deadline| C1[reserve fails →<br/>retry x63 → 429<br/>63 telemetry rows]
    B1 -->|passes| D1{dedicated pool full?}
    D1 -->|yes| E1[instant 429 machine_busy<br/>283k/week]
    D1 -->|no| F1[admit at 2.0x overcommit<br/>gemma p50 8 tok/s]
    F1 --> G1[black-hole box<br/>13k errors, never ejected<br/>breaker reset on reconnect]
  end
  subgraph After
    A2[request] --> B2{TTFT gate<br/>calibrated estimate}
    B2 -->|over deadline| C2[single 429 + Retry-After<br/>1 telemetry row]
    B2 -->|passes| D2{dedicated pool full?}
    D2 -->|yes| E2[queue depth 32 →<br/>drains in seconds]
    D2 -->|no| F2[admit at 1.2x overcommit<br/>projected TPS ≥ floor]
    F2 --> G2[black hole ejected ≤10 failures<br/>state survives reconnect<br/>cold budget misfit never dispatched]
  end
```

## Code: before → after

```mermaid
flowchart LR
  subgraph Before
    a1[dispatch.go run loop] -->|errTTFTTooSlow attempt>0| a2[outcomeRetry x63]
    b1[scheduler.go ttftMsFromSnapshot] --> b2[raw estimate → gate]
    c1[error_cooldown / provider_breaker /<br/>capacity_cooldown] -->|keyed by session UUID| c2[Disconnect wipes state]
    d1[inference_admission.go] -->|IsDedicatedModel| d2[skip queue → 429]
    e1[concurrency_cap.go] --> e2[overcommit 2.0 global]
    f1[freeMemoryAdmits cold path] --> f2[weight-only check]
    g1[provider.go disconnect] --> g2[reason=disconnect 96.9%]
  end
  subgraph After
    a3[dispatch.go run loop] -->|errTTFTTooSlow any attempt| a4[terminal 429, 1 row]
    b3[ttft_calibration.go<br/>observe via commitFirstContent] --> b4[calibrated estimate → gate + preflight]
    c3[faultKeyLocked serial→sekey→acct] --> c4[state survives reconnect<br/>+ capacity streak ejection]
    d3[inference_admission.go] --> d4[dedicated overflow queues<br/>NewRequestQueueFromEnv]
    e3[concurrency_cap.go] --> e4[default 1.2 + per-model map<br/>warm-pool saturation uses quality cap]
    f3[freeMemoryAdmits] --> f4[+ providerBudgetFits<br/>provider-matched budget math]
    g3[provider.go read loop] --> g4[ws_close_code / read_error /<br/>oom_suspected stamped]
  end
```

## Production evidence (7 days)

- 30,175 TTFT-rejected requests → 1,892,407 `inference_routes` rows (62.7× amplification, ~30ms per burned ladder)
- Predicted vs actual TTFT p50: gpt-oss 3303ms vs 992ms; gemma 3048ms vs 1638ms; all 883k `ttft_too_slow` rejections had `could_have_served=true`
- Black holes: 6 providers at 99–100% error absorbing 2.8k–13.3k requests each; fleet median 18 reconnects/machine/week (mean 61, max 724) resetting session-keyed breakers
- Dedicated overflow: 283,178 `machine_busy` 429s vs 290 queued requests all week
- gemma decode: p50 8 tok/s, 81% of requests below the 15 tok/s floor (2.0× overcommit dilution)
- 83k coordinator-admitted → provider-503 dispatches in 3 days (token-budget disagreement)

## Deploy notes (operator)

- **No env changes required to ship** — all defaults are safe and every fix has a kill switch:
  - `EIGENINFERENCE_TTFT_TERMINAL_REJECT=false` restores ladder looping
  - `EIGENINFERENCE_TTFT_CALIBRATION=off` freezes calibration at ratio 1.0
  - `EIGENINFERENCE_SERVABILITY_GATE=false` disables the (now default-on) gate
  - `EIGENINFERENCE_QUALITY_CONCURRENCY_OVERCOMMIT` / `..._BY_MODEL` tune admission density
  - `EIGENINFERENCE_QUEUE_MAX_DEPTH` / `..._MAX_WAIT` tune the queue
- **To activate the warm-pool controller** (already default-active in code): remove `EIGENINFERENCE_WARM_POOL_OBSERVE_ONLY=true` from the prod env.
- Reconnect-churn root cause (macOS sleep vs 90s eviction) needs provider-side/infra follow-up — see `docs/reports/2026-07-03-reconnect-churn-root-cause.md`.

## Test plan

- [x] `cd coordinator && go test ./...` — full suite green (all packages)
- [x] `go test -race ./registry/...` — race-clean
- [x] `gofmt -l` / `go vet` clean on all touched packages
- [x] Regression tests added per fix, each verified to fail without its fix (ladder termination row counts, calibration convergence + gate unblock, breaker reconnect survival + black-hole ejection ≤10, dedicated queue/drain/timeout, overcommit dilution math, provider-budget cold misfit, disconnect-reason stamping)
- [x] Dual review: Bugbot + independent correctness review

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/512"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785717220&installation_id=133247490&pr_number=512&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F512&signature=1c243587a92a573d280e5a6402e87aad1996df28b9944ca3b3ec994a4f65056b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


coordinator/registry/queue.go
```diff
@@ -22,6 +24,8 @@ import (
 	"errors"
 	"sync"
 	"time"
+
+	"github.com/eigeninference/d-inference/coordinator/env"
 )
```
ethenotethan, line 30: 🔵 **[INFO]** ⚡ Queue maps grow unbounded without cleanup mechanism

💡 _Suggestion:_ Add periodic cleanup of stale queue entries or implement size limits on the internal maps to prevent memory leaks from abandoned requests

📊 _Score: 2×3 = 6 · Category: unbounded_allocations_

coordinator/registry/ttft_calibration.go
```diff
@@ -0,0 +1,325 @@
...
+// sweepPendingLocked drops expired predictions; if the map is still at
+// capacity afterwards (pathological reserve flood with no commits), arbitrary
+// entries are dropped so the map stays bounded. Caller holds c.mu.
+func (c *ttftCalibrator) sweepPendingLocked(now time.Time) {
+	for k, p := range c.pending {
+		if now.Sub(p.at) > ttftCalibrationPendingTTL {
+			delete(c.pending, k)
+		}
+	}
+	for k := range c.pending {
```
ethenotethan, line 210: 🟡 **[MEDIUM]** ⚡ Pending predictions map can grow to 8192 entries without bounded cleanup

💡 _Suggestion:_ The sweepPendingLocked function only runs when at capacity. Consider periodic cleanup or lower the trigger threshold to prevent sustained high memory usage

coordinator/api/dispatch_ttft_terminal_test.go
```diff
@@ -0,0 +1,308 @@
...
+	ttftRows, otherRows := 0, []string{}
+	for _, reason := range reasons {
+		if reason == "ttft_too_slow" {
+			ttftRows++
+		} else {
+			otherRows = append(otherRows, reason)
+		}
+	}
+	if ttftRows != 1 || len(otherRows) != 0 {
+		t.Errorf("rejection ledger rows = %v, want exactly one ttft_too_slow", reasons)
```
ethenotethan, line 180: 🔵 **[INFO]** ⚡ Test repeatedly creates providers and runs similar dispatch scenarios

💡 _Suggestion:_ Consider extracting common setup logic to reduce test execution time, though this is test code so impact is minimal

📊 _Score: 1×2 = 2 · Category: repeated_work_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Performance — 4 finding(s) (1 blocking)
- 🔵 **[INFO]** `coordinator/registry/queue.go:24-30` — Queue maps grow unbounded without cleanup mechanism
  - _Suggestion:_ Add periodic cleanup of stale queue entries or implement size limits on the internal maps to prevent memory leaks from abandoned requests
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_calibration.go:184-200` — Pending predictions map can grow to 8192 entries without bounded cleanup
  - _Suggestion:_ The sweepPendingLocked function only runs when at capacity. Consider periodic cleanup or lower the trigger threshold to prevent sustained high memory usage
- 🔵 **[INFO]** `coordinator/registry/health_ejection.go:477-510` — Health ejection maps only sweep when exceeding 2048 entries
  - _Suggestion:_ Consider more frequent sweeping or lower thresholds to prevent memory buildup from churned provider identities
- 🔵 **[INFO]** `coordinator/api/dispatch_ttft_terminal_test.go:100-180` — Test repeatedly creates providers and runs similar dispatch scenarios
  - _Suggestion:_ Consider extracting common setup logic to reduce test execution time, though this is test code so impact is minimal

_4 finding(s) total, 1 blocking. Verdict: REQUEST_CHANGES._

coordinator/registry/ttft_calibration.go
Gajesh2007: Fixed in 2e9dcc6f: added an opportunistic TTL sweep — once the pending map exceeds 1,024 entries, expired predictions are reaped at most once per minute (in addition to the existing at-capacity sweep), so steady-state memory tracks live traffic instead of plateauing at the 8,192 cap. Regression test: `TestTTFTCalibratorPendingSweepsBelowCap`.

coordinator/registry/queue.go
Gajesh2007: Addressed in 2e9dcc6f: `PopNextFresh` and `cleanStaleLocked` now delete the model key when its queue empties, so the per-model map only holds live queues. Note the entries themselves were already bounded (lazy stale sweeps on Enqueue/QueuedModels, per-waiter maxWait timers, per-model maxSize cap) and model ids are catalog-bounded — this removes the residual empty-slice retention.

coordinator/api/dispatch_ttft_terminal_test.go
Gajesh2007: Acknowledged — leaving as-is: each test intentionally builds an isolated registry/dispatch fixture because they pin different ladder behaviors (terminal vs kill-switch looping), and shared setup would couple their env-flag state. Test-only cost, ~seconds.

Gajesh2007:
Addressed the automated review (commit 2e9dcc6f):

- **ttft_calibration pending map (MEDIUM, blocking)** — fixed: opportunistic TTL sweep once the map exceeds 1,024 entries (at most once/min), in addition to the existing at-capacity sweep. Steady-state memory now tracks live traffic instead of plateauing at the 8,192 cap. Regression test `TestTTFTCalibratorPendingSweepsBelowCap`.
- **queue maps (INFO)** — fixed: `PopNextFresh`/`cleanStaleLocked` delete a model's key when its queue empties. The entries themselves were already bounded (lazy stale sweeps, per-waiter maxWait, per-model maxSize).
- **health-ejection sweep thresholds (INFO)** — intentionally NOT changed: stable-identity fault state persisting across reconnects is the entire point of this fix (the prod black hole survived precisely because state was reaped/reset too eagerly). The 2048 threshold is ~4× the real fleet size (512 machines by serial), entries are ~100 bytes, and the sweep never reaps currently-ejected or recently-active identities. Reaping churned identities faster would reopen the reconnect-reset exploit.
- **test setup duplication (INFO)** — leaving as-is: the ladder tests deliberately use isolated fixtures because they pin opposite behaviors of the same env flag (terminal vs kill-switch looping); sharing setup would couple their env state.

Registry suite race-clean and api suite green after the changes.

coordinator/registry/queue.go
ethenotethan: :white_check_mark: Looks addressed in `2e9dcc6` — Added cleanup mechanism that deletes empty queue entries from the map when queues become empty. Resolving.

coordinator/registry/ttft_calibration.go
ethenotethan: :white_check_mark: Looks addressed in `2e9dcc6` — Added ttftCalibrationSweepThreshold (1024) and ttftCalibrationSweepInterval (1 minute) constants to enable periodic cleanup before reaching the 8192 capacity limit.. Resolving.

coordinator/api/dispatch.go
```diff
@@ -392,6 +412,7 @@ func (d *dispatchState) commitFirstContent(pr *registry.PendingRequest, chunk st
 	// ever stamps FirstContentAt for the attempt that actually delivered content —
```
ethenotethan, line 412: 🔵 **[INFO]** ⚡ observeTTFTCalibration call on hot path may block on registry operations

💡 _Suggestion:_ Consider making observeTTFTCalibration async or ensure it's non-blocking since it's called from commitFirstContent on the dispatch goroutine

📊 _Score: 2×3 = 6 · Category: blocking_io_

coordinator/registry/ttft_calibration.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	c.pending[ttftPendingKey(requestID, attempt)] = ttftPendingPrediction{
+		model: model,
+		chip:  chip,
+		rawMs: rawMs,
+		at:    now,
+	}
+}
+
+// sweepPendingLocked drops expired predictions; if the map is still at
```
ethenotethan, line 200: 🟡 **[MEDIUM]** ⚡ Unbounded map growth in notePrediction without proper cleanup

💡 _Suggestion:_ The pending map can grow to ttftCalibrationMaxPending (8192) and sweep logic only runs when threshold is exceeded. Consider more frequent cleanup or lower thresholds to prevent memory spikes

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

- 🔵 **[INFO]** `coordinator/api/dispatch.go:412` — observeTTFTCalibration call on hot path may block on registry operations
  - _Suggestion:_ Consider making observeTTFTCalibration async or ensure it's non-blocking since it's called from commitFirstContent on the dispatch goroutine
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_calibration.go:184-200` — Unbounded map growth in notePrediction without proper cleanup
  - _Suggestion:_ The pending map can grow to ttftCalibrationMaxPending (8192) and sweep logic only runs when threshold is exceeded. Consider more frequent cleanup or lower thresholds to prevent memory spikes
- 🔵 **[INFO]** `coordinator/registry/queue.go:162-179` — PopNextFresh performs O(n) queue cleanup on every call
  - _Suggestion:_ Consider batching stale cleanup or using a more efficient data structure to avoid scanning the entire queue on each pop operation

coordinator/registry/ttft_calibration.go
Gajesh2007: The map is hard-bounded: 8,192 entries × ~100 bytes ≈ 1 MB worst case, and since 2e9dcc6f expired entries are additionally reaped once the map crosses 1,024 (at most once/min), so steady-state tracks live traffic. A lower trigger would add a full-map scan to the reserve hot path for no material memory win.

coordinator/api/dispatch.go
Gajesh2007: Measured cheap and intentionally synchronous: observeTTFTCalibration takes only the calibrator's leaf RWMutex (never the registry lock) for a map insert + cached-median window write, plus a DogStatsD gauge (UDP, non-blocking). The dispatch goroutine already holds pr.Timing here, and commitFirstContent runs once per request — same cost class as the adjacent RecordCapacityAccept fast path. Making it async would add a goroutine per request for no contention win.

coordinator/registry/registry.go
```diff
@@ -972,13 +978,23 @@ func (p *Provider) HardUntrustEpoch() uint64 {
...
+		snapshot := *result
+		p.AttestationResult = &snapshot
+	}
+	// Re-derive the stable identity while p.mu is held, then bind it OUTSIDE
+	// p.mu (bindStableFaultKey takes r.mu; the established order is r.mu →
+	// p.mu, so taking r.mu here while holding p.mu could deadlock). Binding at
+	// attestation time is what re-attaches a reconnecting machine's fault
+	// state (breakers/cooldowns keyed by serial/SE-key) to its fresh session
+	// id BEFORE it becomes routable — public routing requires attestation.
+	id, stableID, r := p.ID, stableProviderIdentityLocked(p), p.registry
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 103cb0a3. Root cause confirmed: verifyProviderAttestation calls SetAttestationResult before checking result.Valid, and attestation.Verify copies the blob's SerialNumber into the result pre-verification. stableProviderIdentityLocked now requires ar.Valid before using serial/SE-key — this single gate covers the bind path, the routing-gate ejection check, GetProviderStableIdentity (the serve-outcome feed), and the disconnect cache. Invalid re-attestation unbinds (session-key fallback). Account fallback kept: AccountID comes from the provider's own authenticated token, never the blob. We chose Valid-gating over MDA-gating because VerificationResult has no MDA field (MDA lands on the Provider minutes later — keying on it would flip fault keys mid-session), and a Valid-but-spoofed serial isn't exploitable: fault state only accumulates via dispatch, dispatch requires hardware trust, and both hardware paths cross-check the attested serial. Regression tests: TestInvalidAttestationDoesNotBindStableFaultKey, TestStableProviderIdentityLocked_Precedence.

coordinator/registry/capacity_cooldown.go
```diff
@@ -193,27 +195,36 @@ func (r *Registry) RecordCapacityReject(providerID, modelID string) (tripped boo
...
-	if !hasStrikes && !hasCooldown && !hasTrips {
+	if !hasStrikes && !hasCooldown && !hasTrips && !hasNodeStreak {
 		return
 	}
 	r.mu.Lock()
+	key = capacityRejectKey{ProviderID: r.faultKeyLocked(providerID), ModelID: modelID}
 	delete(r.capacityRejectStrikes, key)
 	delete(r.capacityCooldowns, key)
 	delete(r.capacityCooldownTrips, key)
+	delete(r.healthEjectionCapacityStreaks, key.ProviderID)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 103cb0a3: RecordCapacityAccept now also deletes healthEjectionTrips + healthEjectionLastTripCapacity when the last trip was capacity-shaped, so an accepting half-open probe can't be insta-re-ejected by one concurrent shed. Fault-shaped trip state is deliberately preserved (first content says nothing about fault behavior; clean-completion success remains the fault-recovery signal, keeping exponential backoff against flappers). Active ejections are not lifted early — in-flight content on an ejected node is possible and early un-ejection would weaken the quarantine. Tests: TestHealthEjection_FirstContentDisarmsCapacityHalfOpen (+ counter-case TestHealthEjection_FirstContentPreservesFaultHalfOpen).

coordinator/api/provider.go
```diff
@@ -191,6 +251,25 @@ func (s *Server) providerReadLoop(ctx context.Context, conn *websocket.Conn, pro
...
+			// no-op. Skipped when:
+			//   - provider == nil: never registered, so no session row exists
+			//     (writing would fabricate a zero-duration row);
+			//   - ctx.Err() != nil: coordinator shutdown — the next instance's
+			//     startup reconcile labels these "coordinator_restart";
+			//   - the registry no longer has the provider: registry.Disconnect
+			//     already ran (stale eviction, duplicate-serial kick) and owns
+			//     the reason for that path.
+			if provider != nil && ctx.Err() == nil && s.registry.GetProvider(providerID) != nil {
+				s.closeSessionWithReason(providerID, sessionDisconnectReason(closeStatus, oomSuspected))
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 103cb0a3: the read-loop exit now flips the provider to StatusOffline (pure in-memory write under p.mu, instant) BEFORE the synchronous session-reason stamp, so a dead socket is unroutable for the entire durable write — a stalled DB delays only that connection's own teardown. The stamp still runs before the deferred registry.Disconnect, so first-close-wins keeps the specific reason; pending-request flush and disconnect-cache behavior unchanged. Regression test TestProviderUnroutableBeforeSessionStamp holds the store write 300ms and asserts the provider is unroutable when the stamp arrives (fails without the fix).

coordinator/api/inference_admission.go
```diff
@@ -236,16 +236,19 @@ func (s *Server) runInferenceAdmission(w http.ResponseWriter, r *http.Request, p
...
+		// Dedicated-family models (e.g. Gemma 4) queue like every other model.
+		// They used to fast-429 here (f28e89a9: a TTFT-SLA caution against
+		// waiting on a dedicated slot), but the wait is bounded by the queue's
+		// maxWait and drains fire fleet-wide on every request completion and
+		// heartbeat — across a large dedicated pool a slot frees within
+		// seconds, while each fast 429 was an uptime-visible shed to
+		// OpenRouter. The drain path (ReserveProviderEx) still applies the
+		// dedicated-box routing gate, so a queued request only ever lands on a
+		// dedicated provider.
+		if s.queueBeforeShedEnabled() {
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 103cb0a3, drain-side: drainQueuedRequestsForModels now fails a waiter immediately with ErrQueueTTFTTooSlow when the reserve rejection is purely TTFT (TTFTRejections>0, zero capacity rejections, zero candidates — deterministic for that pass), and both API waiters translate it into the standard ttft_too_slow 429 + Retry-After (stage=queue telemetry). Mixed capacity+TTFT rejections keep waiting (a busy fast box freeing can still serve); owner-scoped waiters are never TTFT-failed; hard-reject-off is unaffected. The preflight side was verified already-correct: quickCapacityCheck only reports bestTTFT for capacity-passing candidates, so a pure-TTFT-unservable fleet sheds before queueing. Tests: TestQueuedDedicatedRequestFailsFastOnDrainTTFTReject (0.18s vs 10s maxWait), TestDrainFailsPureTTFTRejectedWaiterPromptly, TestDrainMixedRejectionKeepsWaiting, TestDrainDoesNotTTFTFailOwnerScopedWaiters, TestDrainSoftGateUnaffectedByTTFT.

Gajesh2007:
All four Codex P2s addressed in 103cb0a3 (details in each thread): valid-only stable-identity binding, capacity half-open disarm at first content, offline-before-stamp disconnect ordering, and prompt ttft_too_slow 429 for pure-TTFT queue drains. Also fixed the CI lint failure (dead hasConcurrencyHeadroomForModelLocked). Full coordinator suite green, registry race-clean, golangci-lint 0 issues. The two Blacksmith e2e failures are the known Qwen3.5 testbed MLX broadcast_shapes crash (provider process dies under concurrency; unrelated to this PR — same failure signature as on #498/#507).

coordinator/registry/queue.go
```diff
@@ -342,5 +437,12 @@ func (q *RequestQueue) cleanStaleLocked(model string) {
 			fresh = append(fresh, req)
 		}
 	}
+	// Drop the key entirely when nothing survives so the per-model map tracks
+	// live queues only (model ids are catalog-bounded, but no reason to retain
+	// empty entries).
+	if len(fresh) == 0 {
+		delete(q.queues, model)
+		return
+	}
```
ethenotethan, line 446: 🔵 **[INFO]** ⚡ Potential O(n) map deletion pattern in cleanStaleLocked

💡 _Suggestion:_ Consider pre-allocating fresh slice capacity based on queue length to avoid repeated slice growth when most entries are fresh

📊 _Score: 2×3 = 6 · Category: inefficient_data_structure_

ethenotethan (commented):

**Verdict:** COMMENT

### Performance — 1 finding(s)
- 🔵 **[INFO]** `coordinator/registry/queue.go:437-446` — Potential O(n) map deletion pattern in cleanStaleLocked
  - _Suggestion:_ Consider pre-allocating fresh slice capacity based on queue length to avoid repeated slice growth when most entries are fresh

_1 finding(s) total, 0 blocking. Verdict: COMMENT._

coordinator/registry/registry.go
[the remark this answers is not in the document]
Gajesh2007: Fixed in c87f3b66: new Provider.RebindStableFaultKey() (derive under p.mu, bind outside — same lock discipline as SetAttestationResult) is called right after AccountID assignment in the registration flow, so account-fallback providers bind acct:<id> and fault state survives reconnects; bindStableFaultKey's first-bind migration carries any pre-linkage session-keyed strikes across. Invalid re-attestation after linkage keeps the acct: binding (stableProviderIdentityLocked falls back to AccountID on invalid results); a later valid attestation upgrades to serial:/sekey: with state migration. The only other production AccountID site (RestoreProviderState) is reachable solely from the valid-attestation path where serial/sekey outrank acct — rebind there would be a no-op. Tests: TestAccountLinkageBindsStableFaultKey, TestInvalidReattestationKeepsAccountBinding, TestAccountLinkedFaultStateSurvivesReconnect (real WS registration path, both sessions).

coordinator/registry/health_ejection.go
```diff
@@ -118,6 +157,190 @@ func (r *Registry) GetProviderStableIdentity(providerID string) string {
...
+			delete(r.inferenceErrorCooldowns, k)
+		}
+	}
+
+	// Node-health breaker.
+	if w, ok := r.providerOutcomes[oldKey]; ok {
+		if _, exists := r.providerOutcomes[newKey]; !exists {
+			r.providerOutcomes[newKey] = w
+		}
+		delete(r.providerOutcomes, oldKey)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in c87f3b66: migrateFaultStateLocked now MERGES health windows instead of keeping the destination — chronological two-pointer merge of both rings' outcomes (destination wins timestamp ties), truncated to the newest ring-size entries, with consecFail recomputed from the merged tail. Applied to both providerOutcomes and healthEjectionWindows. A mid-streak identity enrichment (sekey: → populated serial:) now yields the true combined consecutive streak. Tests: TestFaultStreakSurvivesIdentityEnrichmentMidStreak (3 serial-keyed + 2 sekey-keyed faults → merged consecFail=5 → next fault trips; fails with the old keep-destination policy) and TestProviderHealthWindowMerge (5 subtests: empty src/dst, interleave, trailing-success reset, overflow + head correctness).

coordinator/api/servability_gate.go
```diff
@@ -18,11 +19,33 @@ import (
...
+// per-provider admission gap (coordinator admit → provider token-budget 503) is
+// exactly what this gate sheds, so an off-by-default no longer matches reality.
+// Only an explicit parseable false disables it.
+func (s *Server) servabilityGateEnabled() bool {
+	if s.servabilityGate {
+		return true
+	}
+	v := os.Getenv("EIGENINFERENCE_SERVABILITY_GATE")
+	if v == "" {
+		return true
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in c87f3b66: the tier-2 fleet shed is now STRUCTURAL — PredictServable compares prompt+max_tokens against each provider's budget CEILING (resident: activeTokenBudgetMax as reported, no live subtraction; cold: the optimistic post-load estimate), so a busy-but-fitting fleet falls through to the capacity/queue-before-shed path and only never-fits requests 429 early. The helper was renamed (liveStructuralBudget → structural ceiling vs liveRemainingBudget) so the names match semantics; providerBudgetFits deliberately KEEPS live-remaining semantics for per-provider admission in freeMemoryAdmits, where a rejection is a capacity rejection that correctly queues. Old live-subtract test replaced with structural-contract tests: TestPredictServableTransientFullnessFallsThroughToQueue, TestPredictServableStructuralImpossibilityShedsEarly (+ existing fail-open tests unchanged and passing).

Gajesh2007:
Rebased onto latest master (b4e0d45a — includes #510's capacity-cooldown probe-claim rework, merged semantically with this PR's stable fault keying: the probe gate and claim now resolve through faultKeyLocked). All three round-2 Codex P2s addressed in c87f3b66 — details in each thread. Full coordinator suite green, registry race-clean, golangci-lint 0 issues.

coordinator/registry/queue.go
```diff
@@ -162,6 +254,9 @@ func (q *RequestQueue) PopNextFresh(model string) *QueuedRequest {
 		req := queue[0]
 		queue = queue[1:]
 		q.queues[model] = queue
+		if len(queue) == 0 {
```
ethenotethan, line 257: 🔵 **[INFO]** ⚡ Queue cleanup creates unbounded slice growth in hot path

💡 _Suggestion:_ Pre-allocate fresh slice with estimated capacity based on queue size to avoid repeated allocations during cleanup: fresh := make([]*QueuedRequest, 0, len(queue))

📊 _Score: 2×3 = 6 · Category: unbounded_allocations_

coordinator/registry/ttft_calibration.go
```diff
@@ -0,0 +1,335 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if len(c.pending) >= ttftCalibrationMaxPending ||
+		(len(c.pending) > ttftCalibrationSweepThreshold && now.Sub(c.lastSweep) > ttftCalibrationSweepInterval) {
+		c.sweepPendingLocked(now)
+	}
+	c.pending[ttftPendingKey(requestID, attempt)] = ttftPendingPrediction{
+		model: model,
+		chip:  chip,
+		rawMs: rawMs,
```
ethenotethan, line 195: 🟡 **[MEDIUM]** ⚡ Calibrator sweep creates new map on every call when at capacity

💡 _Suggestion:_ Reuse existing map by clearing entries in-place instead of creating new map: for k := range c.pending { delete(c.pending, k); if len(c.pending) < ttftCalibrationMaxPending { break } }

coordinator/registry/provider_breaker.go
```diff
@@ -95,6 +97,62 @@ func (w *providerHealthWindow) record(ok bool, now time.Time) {
...
+	}
+	a, b := w.chronological(), src.chronological()
+	merged := make([]providerHealthOutcome, 0, len(a)+len(b))
+	i, j := 0, 0
+	for i < len(a) && j < len(b) {
+		if !a[i].ts.After(b[j].ts) {
+			merged = append(merged, a[i])
+			i++
+		} else {
+			merged = append(merged, b[j])
```
ethenotethan, line 135: 🔵 **[INFO]** ⚡ Chronological slice creation and copying in merge operation

💡 _Suggestion:_ Consider using ring buffer indices directly instead of creating intermediate slices, or implement merge without full materialization of chronological order

📊 _Score: 2×3 = 6 · Category: inefficient_data_structures_

coordinator/registry/ttft_calibration.go
```diff
@@ -0,0 +1,335 @@
...
+func clampTTFTCalibrationRatio(ratio float64) float64 {
+	if math.IsNaN(ratio) || ratio <= 0 {
+		return 1.0
+	}
+	if ratio < ttftCalibrationRatioMin {
+		return ttftCalibrationRatioMin
+	}
+	if ratio > ttftCalibrationRatioMax {
+		return ttftCalibrationRatioMax
+	}
```
ethenotethan, line 153: 🟡 **[MEDIUM]** ⚡ Median recalculated on every sample addition with full sort

💡 _Suggestion:_ Use a more efficient data structure like two heaps (min/max) or maintain sorted insertion to avoid O(n log n) sort on every update

coordinator/registry/queue.go
```diff
@@ -42,10 +54,20 @@ type QueuedRequest struct {
 	doneOnce   sync.Once
 
 	// Decision captures the cost breakdown of the routing decision that
-	// dispatched this queued request. Populated by drainQueuedRequestsForModels
-	// just before ResponseCh is signaled, so consumers can emit the same
-	// metrics they would for an immediate (non-queued) selection.
+	// dispatched (or terminally failed) this queued request. Populated by
```
ethenotethan, line 57: 🟡 **[MEDIUM]** 🏷️ Missing type assertion check in FailureReason field usage

💡 _Suggestion:_ Add type assertion check when reading FailureReason: if req.FailureReason != nil { return nil, req.FailureReason }. The field is written before channel send but read after receiving nil, so channel ordering should protect access, but explicit checking would be safer.

📊 _Score: 3×3 = 9 · Category: missing_type_assertion_

coordinator/registry/ttft_calibration.go
```diff
@@ -0,0 +1,335 @@
...
+func TTFTCalibrationRatio(model, chip string) float64 {
+	ttftCalibration.mu.RLock()
+	defer ttftCalibration.mu.RUnlock()
+	return ttftCalibration.learnedRatioLocked(model, chip)
+}
+
+// ResetTTFTCalibration clears all calibration state. Test hook.
+func ResetTTFTCalibration() {
+	ttftCalibration.reset()
+}
```
ethenotethan, line 335: 🟡 **[MEDIUM]** 🧩 New 335-line TTFT calibration system with complex state management for a single use case

💡 _Suggestion:_ Consider if this complex online learning system is necessary vs simpler constant adjustments or if it could be extracted to a separate service

coordinator/registry/stable_fault_key_test.go
```diff
@@ -0,0 +1,617 @@
...
+		{500, "panic: index out of range", false},
+		{400, "token budget", false},
+		{429, "queue full", false},
+	}
+	for _, c := range cases {
+		if got := isNodeCapacityRejectStrike(c.code, c.err); got != c.want {
+			t.Errorf("isNodeCapacityRejectStrike(%d, %q)=%v, want %v", c.code, c.err, got, c.want)
+		}
+	}
+}
```
ethenotethan, line 617: 🔵 **[INFO]** 🧩 617-line test file that may be testing implementation details rather than behavior

💡 _Suggestion:_ Review if all test scenarios are necessary or if some test internal implementation details that could be simplified

📊 _Score: 2×3 = 6 · Category: dead code_

coordinator/registry/health_ejection.go
```diff
@@ -118,6 +157,194 @@ func (r *Registry) GetProviderStableIdentity(providerID string) string {
...
+	// on the FIRST bind (strikes recorded pre-attestation live under the
+	// faultKeyLocked session fallback), or from the previous identity on a
+	// rebind (e.g. sekey: → serial: after MDA enrichment). Without this, a
+	// machine near quarantine sheds its history at the exact moment its
+	// identity improves. No-op on the common re-attestation with an unchanged
+	// identity.
+	old := r.faultKeyBySession[sessionID]
+	if old == "" {
+		old = sessionID
+	}
```
ethenotethan, line 194: 🟡 **[MEDIUM]** 🧩 Complex fault-state migration system with multiple map types and merge logic

💡 _Suggestion:_ Consider if the fault state migration could be simplified or if some of the complexity is necessary for the reconnect-proofing feature

📊 _Score: 3×3 = 9 · Category: over-abstraction_

coordinator/registry/registry.go
```diff
@@ -1293,12 +1325,35 @@ type Registry struct {
...
+	// ignores (prod: 13,333 "token_budget"-shaped 503s, 100% error rate, 90
+	// min, never ejected). Any success clears the streak, so a busy-but-
+	// serving box can never trip. See health_ejection.go.
+	healthEjectionCapacityStreaks map[string]capacityStreak
+	// healthEjectionLastTripCapacity records whether an identity's most recent
+	// ejection came from the capacity streak (true) or the fault path (false).
+	// The capacity branch's half-open instant re-arm applies only to capacity
+	// trips: a single capacity shed — legitimate for a healthy-but-full box —
+	// must not re-arm a FAULT ejection whose cooldown just expired.
+	healthEjectionLastTripCapacity map[string]bool
```
ethenotethan, line 1340: 🔵 **[INFO]** 🧩 Multiple new map fields added to Registry struct for fault tracking

💡 _Suggestion:_ Consider if these maps could be consolidated into a single fault state structure to reduce the number of separate maps to maintain

📊 _Score: 2×3 = 6 · Category: additive state_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Performance — 4 finding(s) (2 blocking)
- 🔵 **[INFO]** `coordinator/registry/queue.go:257` — Queue cleanup creates unbounded slice growth in hot path
  - _Suggestion:_ Pre-allocate fresh slice with estimated capacity based on queue size to avoid repeated allocations during cleanup: fresh := make([]*QueuedRequest, 0, len(queue))
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_calibration.go:183-195` — Calibrator sweep creates new map on every call when at capacity
  - _Suggestion:_ Reuse existing map by clearing entries in-place instead of creating new map: for k := range c.pending { delete(c.pending, k); if len(c.pending) < ttftCalibrationMaxPending { break } }
- 🔵 **[INFO]** `coordinator/registry/provider_breaker.go:119-135` — Chronological slice creation and copying in merge operation
  - _Suggestion:_ Consider using ring buffer indices directly instead of creating intermediate slices, or implement merge without full materialization of chronological order
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_calibration.go:149-153` — Median recalculated on every sample addition with full sort
  - _Suggestion:_ Use a more efficient data structure like two heaps (min/max) or maintain sorted insertion to avoid O(n log n) sort on every update

### Type_diligence — 1 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/registry/queue.go:57` — Missing type assertion check in FailureReason field usage
  - _Suggestion:_ Add type assertion check when reading FailureReason: if req.FailureReason != nil { return nil, req.FailureReason }. The field is written before channel send but read after receiving nil, so channel ordering should protect access, but explicit checking would be safer.

### Additive_complexity — 4 finding(s) (2 blocking)
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_calibration.go:1-335` — New 335-line TTFT calibration system with complex state management for a single use case
  - _Suggestion:_ Consider if this complex online learning system is necessary vs simpler constant adjustments or if it could be extracted to a separate service
- 🔵 **[INFO]** `coordinator/registry/stable_fault_key_test.go:1-617` — 617-line test file that may be testing implementation details rather than behavior
  - _Suggestion:_ Review if all test scenarios are necessary or if some test internal implementation details that could be simplified
- 🟡 **[MEDIUM]** `coordinator/registry/health_ejection.go:88-194` — Complex fault-state migration system with multiple map types and merge logic
  - _Suggestion:_ Consider if the fault state migration could be simplified or if some of the complexity is necessary for the reconnect-proofing feature
- 🔵 **[INFO]** `coordinator/registry/registry.go:1325-1340` — Multiple new map fields added to Registry struct for fault tracking
  - _Suggestion:_ Consider if these maps could be consolidated into a single fault state structure to reduce the number of separate maps to maintain

_9 finding(s) total, 5 blocking. Verdict: REQUEST_CHANGES._

coordinator/registry/health_ejection.go
```diff
@@ -64,6 +88,21 @@ func healthEjectionEnabled() bool {
 // hardware serial → SE public key → account id), or "" when none is available
 // (un-attestable → never ejected, fail-open).
```
ethenotethan, line 89: 🟡 **[MEDIUM]** 🔒 Attestation validation bypass in stable identity derivation

💡 _Suggestion:_ The code only checks ar.Valid for serial/SE key but not for account fallback. An invalid attestation with a spoofed serial could bind fault state under 'serial:<victim>' if the Valid check is bypassed elsewhere. Ensure Valid is checked consistently across all attestation usage.

📊 _Score: 3×3 = 9 · Category: auth-bypass_

coordinator/registry/ttft_calibration.go
```diff
@@ -0,0 +1,335 @@
...
+	ttftCalibrationSweepInterval  = time.Minute
+)
+
+// ttftCalibrationEnabled is the live-read kill switch:
+// EIGENINFERENCE_TTFT_CALIBRATION=off (or false/0) makes the apply path return
+// ratio 1.0. Mirrors decodeFloorUseFleetMedian's live-env pattern.
+func ttftCalibrationEnabled() bool {
+	v := strings.TrimSpace(env.EnvOr(env.EnvPrefix+"_TTFT_CALIBRATION", "on"))
+	return !strings.EqualFold(v, "off") && v != "false" && v != "0"
+}
```
ethenotethan, line 96: 🔵 **[INFO]** 🔒 Environment variable parsing without validation

💡 _Suggestion:_ The ttftCalibrationEnabled() function reads EIGENINFERENCE_TTFT_CALIBRATION from environment and compares string values without proper validation. Consider using a whitelist of allowed values and proper boolean parsing.

📊 _Score: 2×3 = 6 · Category: input-validation_

coordinator/registry/servability.go
```diff
@@ -124,28 +131,67 @@ func snapshotStructuralBudget(snap routingSnapshot) (budget int64, known bool) {
...
+// freeMemoryAdmits), where a "doesn't fit right now" is a capacity rejection
+// that queues. It must NOT feed the fleet-level servability shed, which is
+// structural-only (see PredictServable) — a live-remaining fleet 429 would shed
+// a merely-busy fleet ahead of the queue.
+func liveRemainingBudget(snap routingSnapshot) (budget int64, known bool) {
 	if snap.activeTokenBudgetMax > 0 {
 		rem := snap.activeTokenBudgetMax - snap.activeTokenBudgetUsed - snap.queuedTokenBudget
 		if rem < 0 {
 			rem = 0
 		}
```
ethenotethan, line 150: 🔵 **[INFO]** 🔒 Potential integer overflow in budget calculations

💡 _Suggestion:_ The budget calculations in liveRemainingBudget() perform arithmetic on int64 values from user-controlled sources without overflow checks. Consider using math.MaxInt64 bounds checking or safer arithmetic operations.

📊 _Score: 2×2 = 4 · Category: input-validation_

coordinator/registry/queue.go
```diff
@@ -162,6 +254,9 @@ func (q *RequestQueue) PopNextFresh(model string) *QueuedRequest {
 		req := queue[0]
 		queue = queue[1:]
 		q.queues[model] = queue
+		if len(queue) == 0 {
+			delete(q.queues, model)
```
ethenotethan, line 258: 🔵 **[INFO]** ⚡ Queue cleanup creates unbounded slice growth in hot path

💡 _Suggestion:_ Pre-allocate fresh slice with estimated capacity based on queue size to avoid repeated allocations during cleanup: fresh := make([]*QueuedRequest, 0, len(queue))

📊 _Score: 2×3 = 6 · Category: unbounded_allocations_

coordinator/registry/ttft_calibration.go
```diff
@@ -0,0 +1,335 @@
...
+	now := time.Now()
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if len(c.pending) >= ttftCalibrationMaxPending ||
+		(len(c.pending) > ttftCalibrationSweepThreshold && now.Sub(c.lastSweep) > ttftCalibrationSweepInterval) {
+		c.sweepPendingLocked(now)
+	}
+	c.pending[ttftPendingKey(requestID, attempt)] = ttftPendingPrediction{
+		model: model,
+		chip:  chip,
```
ethenotethan, line 194: 🟡 **[MEDIUM]** ⚡ Calibrator sweep creates new map on every call when at capacity

💡 _Suggestion:_ Reuse existing map by deleting entries in-place instead of creating new map: for k, p := range c.pending { if shouldDelete { delete(c.pending, k) } }

coordinator/registry/queue.go
```diff
@@ -30,6 +34,14 @@ var ErrQueueFull = errors.New("request queue is full")
 // ErrQueueTimeout is returned when a queued request times out waiting for a provider.
```
ethenotethan, line 34: 🟡 **[MEDIUM]** 🏷️ Missing type assertion check for ErrQueueTTFTTooSlow error handling

💡 _Suggestion:_ Add proper type assertion checking when handling the new error type. Consider using errors.Is() pattern or add explicit type checks where this error is handled to ensure type safety.

📊 _Score: 3×3 = 9 · Category: missing_type_assertion_

coordinator/registry/ttft_calibration.go
ethenotethan, line 335: 🟡 **[MEDIUM]** 🧩 New TTFT calibration system adds significant complexity with multiple maps, windows, and state management

💡 _Suggestion:_ Consider if this complex calibration system could be simplified or if the existing constant-based approach could be improved with simpler tuning

coordinator/registry/stable_fault_key_test.go
ethenotethan, line 617: 🔵 **[INFO]** 🧩 617-line test file with complex test scenarios may indicate the underlying fault key system is overly complex

💡 _Suggestion:_ Review if the stable fault key binding system could be simplified to reduce the test complexity needed to verify it

coordinator/registry/concurrency_cap.go
```diff
@@ -43,6 +85,60 @@ func (r *Registry) SetQualityConcurrencyCap(enabled bool, overcommit, floorTPS f
...
+}
+
+// qualityCapOvercommitForModelLocked resolves the overcommit for a model: the
+// per-model override when one exists for the resolved build id, else the global
+// value. Caller holds r.mu.
+func (r *Registry) qualityCapOvercommitForModelLocked(model string) float64 {
+	if v, ok := qualityCapOvercommitByModel[strings.ToLower(model)]; ok {
+		return v
+	}
+	return r.qualityCapOvercommit
```
ethenotethan, line 139: 🔵 **[INFO]** 🧩 Per-model overcommit configuration adds parsing complexity and another configuration dimension

💡 _Suggestion:_ Evaluate if per-model overcommit overrides are necessary or if the global default change achieves the same production goals with less configuration surface

coordinator/api/dispatch_ttft_terminal_test.go
```diff
@@ -0,0 +1,308 @@
...
+	if status != http.StatusTooManyRequests {
+		t.Fatalf("status = %d, want 429; body = %s", status, body)
+	}
+	if !strings.Contains(body, "attempt(s)") {
+		t.Errorf("kill switch must restore the legacy dispatch-exhausted response; body = %s", body)
+	}
+	if got := settleTTFT429Routes(t, st); got <= 1 {
+		t.Errorf("ttft_429 route rows = %d, want the legacy multi-attempt ladder (> 1)", got)
+	}
+}
```
ethenotethan, line 308: 🔵 **[INFO]** 🧩 308-line test file for TTFT terminal rejection suggests the dispatch logic may be becoming complex

💡 _Suggestion:_ Consider if the TTFT rejection logic could be extracted to a simpler, more testable component

📊 _Score: 2×2 = 4 · Category: test-complexity_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 3 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/registry/health_ejection.go:74-89` — Attestation validation bypass in stable identity derivation
  - _Suggestion:_ The code only checks ar.Valid for serial/SE key but not for account fallback. An invalid attestation with a spoofed serial could bind fault state under 'serial:<victim>' if the Valid check is bypassed elsewhere. Ensure Valid is checked consistently across all attestation usage.
- 🔵 **[INFO]** `coordinator/registry/ttft_calibration.go:94-96` — Environment variable parsing without validation
  - _Suggestion:_ The ttftCalibrationEnabled() function reads EIGENINFERENCE_TTFT_CALIBRATION from environment and compares string values without proper validation. Consider using a whitelist of allowed values and proper boolean parsing.
- 🔵 **[INFO]** `coordinator/registry/servability.go:131-150` — Potential integer overflow in budget calculations
  - _Suggestion:_ The budget calculations in liveRemainingBudget() perform arithmetic on int64 values from user-controlled sources without overflow checks. Consider using math.MaxInt64 bounds checking or safer arithmetic operations.

### Performance — 4 finding(s) (2 blocking)
- 🔵 **[INFO]** `coordinator/registry/queue.go:254-258` — Queue cleanup creates unbounded slice growth in hot path
  - _Suggestion:_ Pre-allocate fresh slice with estimated capacity based on queue size to avoid repeated allocations during cleanup: fresh := make([]*QueuedRequest, 0, len(queue))
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_calibration.go:184-194` — Calibrator sweep creates new map on every call when at capacity
  - _Suggestion:_ Reuse existing map by deleting entries in-place instead of creating new map: for k, p := range c.pending { if shouldDelete { delete(c.pending, k) } }
- 🔵 **[INFO]** `coordinator/registry/provider_breaker.go:54-66` — Ring buffer merge performs O(n) slice operations in chronological()
  - _Suggestion:_ Optimize chronological() to avoid slice allocation by using indices directly or implement a more efficient ring merge algorithm
- 🟡 **[MEDIUM]** `coordinator/registry/scheduler.go:2252-2270` — Queue drain loop repeatedly calls ReserveProviderEx for same model
  - _Suggestion:_ Cache routing decision or provider selection logic within the drain loop to avoid redundant scheduling computations for the same model state

### Type_diligence — 1 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/registry/queue.go:34` — Missing type assertion check for ErrQueueTTFTTooSlow error handling
  - _Suggestion:_ Add proper type assertion checking when handling the new error type. Consider using errors.Is() pattern or add explicit type checks where this error is handled to ensure type safety.

### Additive_complexity — 5 finding(s) (2 blocking)
- 🟡 **[MEDIUM]** `coordinator/registry/ttft_calibration.go:1-335` — New TTFT calibration system adds significant complexity with multiple maps, windows, and state management
  - _Suggestion:_ Consider if this complex calibration system could be simplified or if the existing constant-based approach could be improved with simpler tuning
- 🔵 **[INFO]** `coordinator/registry/stable_fault_key_test.go:1-617` — 617-line test file with complex test scenarios may indicate the underlying fault key system is overly complex
  - _Suggestion:_ Review if the stable fault key binding system could be simplified to reduce the test complexity needed to verify it
- 🟡 **[MEDIUM]** `coordinator/registry/health_ejection.go:149-194` — migrateFaultStateLocked adds complex state migration logic across multiple map types with merge policies
  - _Suggestion:_ Consider if fault state migration could use a simpler approach, perhaps with a single unified fault state structure rather than migrating across multiple specialized maps
- 🔵 **[INFO]** `coordinator/registry/concurrency_cap.go:85-139` — Per-model overcommit configuration adds parsing complexity and another configuration dimension
  - _Suggestion:_ Evaluate if per-model overcommit overrides are necessary or if the global default change achieves the same production goals with less configuration surface
- 🔵 **[INFO]** `coordinator/api/dispatch_ttft_terminal_test.go:1-308` — 308-line test file for TTFT terminal rejection suggests the dispatch logic may be becoming complex
  - _Suggestion:_ Consider if the TTFT rejection logic could be extracted to a simpler, more testable component

_13 finding(s) total, 6 blocking. Verdict: REQUEST_CHANGES._

PR #523, opened 2026-07-08
fix(routing): gray-box budget clamp + capacity-503 rate penalty (postmortem layers 3-4)

Gajesh2007:
# PR body — `fix/budget-clamp-gray-box` @ `da7957ca`

> **Suggested title:** `fix(routing): gray-box budget clamp + capacity-503 rate penalty`
> **Base:** `master` (branched from `76447bc0`) · 1 commit · worktree `.worktrees/budget-clamp`
> **Deploy batch:** **A** (before the shed lift, before the v0.7.5 provider release)

## Problem

Postmortem **layers 3 and 4** (`docs/reports/2026-07-06-gemma4-serving-failure-postmortem.md`): during the Jul 6 open-pool window, mixed boxes co-resident with gpt-oss served ~60% of their gemma dispatches while failing the other ~40% with capacity-shaped 503s — **11,581 provider 503s in 6 hours** ("gray boxes").

- **Layer 3 — breakers blind by design:** every existing capacity breaker (pair cooldown in `capacity_cooldown.go`, node capacity streak in `health_ejection.go`) requires *zero interleaved accepts*; each accept reset the streak, so a box failing 40% of dispatches was never tripped.
- **Layer 4 — heartbeat trust:** the provider's live KV gate rejects while its last heartbeat still advertises a ~1.4%-used token budget (`active_token_budget_used` ~72k of ~5.2M at admit time). The heartbeat budget is stale-*optimistic*; coordinator admission believed it and kept dispatching.

## What changed (per commit)

One commit, `da7957ca`, two complementary mechanisms — both keyed by the **stable fault identity** (`faultKeyLocked`: serial → SE-key → account → session), migrated on identity rebind, never cleared on `Disconnect`:

1. **Budget clamp** (`coordinator/registry/budget_clamp.go`, new): on the first capacity-shaped 503 for a (provider, model) pair (same classified `RecordCapacityReject` entry point that feeds the cooldown), admission stops believing the pair's heartbeat budget — the slot is treated **FULL** (`freeMemoryAdmits` budget branch rejects; `liveRemainingBudget` in `servability.go` reports zero live headroom; the *structural* ceiling stays raw so the transient clamp can never feed the fleet-level structural 429). **Release requires provider-side proof, whichever lands later:** (a) a heartbeat delivered *strictly after* the clamp whose raw budget shows ≥ **1024 tokens** of headroom, and (b) an accept for the pair after the clamp. **Fail-open:** 5-minute TTL, so a zero-traffic pair can never strand; a re-reject re-arms. Scoped to budget-reporting slots only (`activeTokenBudgetMax > 0`) — legacy providers keep the existing threshold-5 pair cooldown and are never one-shot clamped.
2. **Capacity-503 rate penalty** (`coordinator/registry/capacity_rate.go`, new): a 5-minute sliding window of capacity-503s AND accepts per (identity, model) with **no accept-triggered reset** (that reset *was* the blindness). Above a 0.25 reject rate with ≥8 windowed outcomes, `buildCandidateWithReason` adds `rate × EIGENINFERENCE_CAPACITY_RATE_PENALTY_MS` (default 15,000 ms) to the candidate's cost — proportional deration, never ejection; fail-open selection untouched; the penalty decays as outcomes age out. One served request counts exactly one accept outcome: the completion-time accept is deduped against the commit-time accept via the new `RecordCapacityAcceptOutcome(provider, model, !pr.ContentCommittedSafe())` (`api/consumer.go`).

Observability: `RoutingDecision` gains `CapacityRateMs` and `CapacityRejectRate` (in-memory only — telemetry rows keep RAW budgets, **no DB schema change, wire protocol unchanged**). Both trackers are migrated in `migrateFaultStateLocked` (`health_ejection.go`) and bounded by the sibling >1024 opportunistic sweeps.

## Before / After

```mermaid
flowchart TB
  subgraph Before["Before — gray box invisible (layers 3+4)"]
    A1[dispatch to mixed box] --> B1{provider live KV gate}
    B1 -- "40%: capacity 503" --> C1[RecordCapacityReject<br/>streak +1]
    B1 -- "60%: serves" --> D1[RecordCapacityAccept<br/>streak RESET to 0]
    D1 --> E1[breaker never trips]
    C1 --> E1
    E1 --> F1[admission re-reads heartbeat budget:<br/>~1.4% used → believes it]
    F1 --> A1
  end
  subgraph After["After — clamp + rate penalty"]
    A2[dispatch to mixed box] --> B2{provider live KV gate}
    B2 -- "first capacity 503" --> C2[RecordCapacityReject →<br/>recordBudgetClampLocked +<br/>recordCapacityRateRejectLocked]
    C2 --> D2["slot treated FULL:<br/>freeMemoryAdmits=false,<br/>liveRemainingBudget=0"]
    D2 --> E2{"release proof?<br/>(a) fresher heartbeat ≥1024 headroom<br/>AND (b) accept after clamp"}
    E2 -- no --> F2[stays clamped, ≤5 min TTL fail-open]
    E2 -- yes --> G2[released; re-reject re-arms]
    B2 -- serves --> H2[accept counted ONCE in rate window,<br/>NO streak reset]
    H2 --> I2["rate >0.25 @ ≥8 outcomes →<br/>cost += rate × 15000ms<br/>(derated, never ejected)"]
  end
```

Code flow delta: `RecordCapacityReject` → *(new)* `recordBudgetClampLocked` + `recordCapacityRateRejectLocked`; `RecordCapacityAccept` → wrapper over *(new)* `RecordCapacityAcceptOutcome` → `noteBudgetClampAcceptLocked` + deduped `recordCapacityRateAcceptLocked`; `snapshotProviderLockedEx` / `quickCapacityCheck` → *(new)* `snap.budgetClamped` via `budgetClampActiveLocked`; `freeMemoryAdmits` + `liveRemainingBudget` consume it; `buildCandidateWithReason` → *(new)* `capacityRatePenaltyLocked` cost term.

## New envs

| Env | Default | Semantics | Runbook flip |
|---|---|---|---|
| `EIGENINFERENCE_BUDGET_CLAMP` | `true` | Kill switch for the budget clamp; `false`/`0` restores stale-heartbeat admission | Active from Deploy A; leave default. Kill switch only if clamp misbehaves |
| `EIGENINFERENCE_BUDGET_CLAMP_TTL_SECONDS` | `300` | Fail-open bound: a clamp never outlives `clampedAt + TTL` | Leave default |
| `EIGENINFERENCE_CAPACITY_RATE_PENALTY_MS` | `15000` | Penalty scale: `cost += rate × this` once rate >0.25 with ≥8 outcomes; `0`/negative disables the tracker entirely | Active from Deploy A; leave default |

(Window 5 min, threshold 0.25, min-sample 8, release headroom 1024 tokens are code constants, not envs.)

## Test evidence

`coordinator/registry/budget_clamp_test.go` (533 lines) + `capacity_rate_test.go` (268 lines):

- **`TestGrayBoxSelectionShareDropsWithFix`** — the prod gray-box scenario (40% 503s with interleaved accepts): selection share drops with the fix; monopoly preserved with both kill switches off (pins the old behavior).
- Clamp lifecycle: `TestBudgetClampBlocksAdmissionOnFirstCapacityReject`, `TestBudgetClampReleaseRequiresFreshHeartbeatAndAccept` (both orders), `TestBudgetClampNoMeaningfulHeadroomDoesNotRelease`, `TestBudgetClampTTLFailOpen`, `TestBudgetClampReArmResetsReleaseProof`, `TestBudgetClampLegacyBudgetlessProviderNotClamped`, `TestBudgetClampKillSwitch`.
- Rate window: `TestCapacityRateAcceptsDoNotResetRejects` (the core semantic), `TestCapacityRateOutcomeDedupe` (one outcome per served request), `TestCapacityRateBelowMinSampleNoPenalty`, `TestCapacityRateAtThresholdNoPenalty`, `TestCapacityRateDecays`, `TestCapacityRateNeverClosesRouting` (sole-candidate fail-open), `TestCapacityRateKillSwitch`.
- Identity: `TestBudgetClampSurvivesReconnect`, `TestCapacityRateSurvivesReconnect`, `TestBudgetClampAndRateStateMigrateOnFirstBind`; `TestBudgetClampAndRateConcurrentAccess` (-race hammer).

Suite status: full `registry` (-race) + `api` suites green, golangci-lint clean (Warden session); **re-verified 2026-07-07 by ship-prep: `go test ./registry/... ./api/...` ok in the worktree**.

## Review status

Built and tested by the **Warden** session (JOURNAL.md row, 2026-07-06): registry -race + api suites green, golangci-lint clean. Committed locally, not yet pushed — orchestrator review on the PR itself.

## Rollout notes

- **Deploy batch A** (with coord-a). Must precede any gemma shed lift.
- Nothing dormant: clamp + rate penalty are **active immediately at deploy** (defaults on). They only affect pairs that emit capacity-503s, so today's healthy fleet is byte-identical.
- Post-deploy verify with the postmortem gray-box SQL (appendix): capacity-503 share from previously-gray boxes must drop; watch clamp arm/release counts and `CapacityRejectRate` on routing decisions.
- Rollback: env kill switches (`EIGENINFERENCE_BUDGET_CLAMP=false`, `EIGENINFERENCE_CAPACITY_RATE_PENALTY_MS=0`) — no code revert needed.
- Merge-order note: conflicts with coord-a in `registry/scheduler.go` (both touch `snapshotProviderLockedEx`/`freeMemoryAdmits`); whichever merges second takes a small rebase. No conflict with coord-b.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/523"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786081377&installation_id=133247490&pr_number=523&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F523&signature=0db4d06810f9d1fea3760aa8caa8a56388d339330b3833b581a09b43db0f9adf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -306,7 +306,19 @@ func (s *Server) noteInferenceError(providerID string, pr *registry.PendingReque
 	if (statusCode == http.StatusTooManyRequests || statusCode == http.StatusNotFound ||
 		statusCode >= http.StatusInternalServerError) &&
 		isCapacityRejectStrike(errStr) {
-		if s.registry.RecordCapacityReject(providerID, pr.Model) {
+		// A cold "model not loaded" miss is benign warm-up lifecycle, not
+		// capacity dishonesty. It still feeds the black-hole cooldown (a box
+		// that 404s forever with zero accepts is a black hole), but it must NOT
+		// derate the pair's gray-box capacity-503 RATE (capacity_rate.go) — that
+		// window has no accept-reset, so counting a healthy box's normal reloads
+		// would penalize it. Genuine capacity/token-budget 503s derate.
+		var tripped bool
+		if isColdModelMissRejection(errStr) {
+			tripped = s.registry.RecordCapacityRejectLifecycle(providerID, pr.Model)
+		} else {
+			tripped = s.registry.RecordCapacityReject(providerID, pr.Model)
+		}
+		if tripped {
 			s.ddIncr(metricCapacityCooldownTripped, []string{"provider_id:" + providerID, "model:" + pr.Model})
 			s.logger.Warn("capacity-reject cooldown tripped: provider+model capacity-rejecting with zero interleaved accepts — routing will skip the pair until the cooldown expires",
 				"provider_id", providerID,
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -311,11 +311,20 @@ func (s *Server) noteInferenceError(providerID string, pr *registry.PendingReque
 		// that 404s forever with zero accepts is a black hole), but it must NOT
 		// derate the pair's gray-box capacity-503 RATE (capacity_rate.go) — that
 		// window has no accept-reset, so counting a healthy box's normal reloads
-		// would penalize it. Genuine capacity/token-budget 503s derate.
+		// would penalize it. A "batch token budget" reject that classifyRejection
+		// proves REQUEST-deterministic (provider budget not below the model
+		// context ⇒ the binding term was the fleet-wide context) indicts the
+		// request, not the provider: it counts a cooldown strike only — arming
+		// the one-shot clamp or the no-reset rate window off a single oversized
+		// prompt would gate/derate a healthy pair. Genuine capacity/token-budget
+		// 503s feed everything.
 		var tripped bool
-		if isColdModelMissRejection(errStr) {
+		switch {
+		case isColdModelMissRejection(errStr):
 			tripped = s.registry.RecordCapacityRejectLifecycle(providerID, pr.Model)
-		} else {
+		case s.isRequestShapeBatchBudgetReject(providerID, pr.Model, errStr):
+			tripped = s.registry.RecordCapacityRejectRequestShape(providerID, pr.Model)
+		default:
 			tripped = s.registry.RecordCapacityReject(providerID, pr.Model)
 		}
 		if tripped {
@@ -336,6 +345,41 @@ func (s *Server) noteInferenceError(providerID string, pr *registry.PendingReque
 // inference-error breaker) and routing.provider_breaker_open (node health) so
 // black-hole trips are independently alertable.
 const metricCapacityCooldownTripped = "routing.capacity_cooldown_tripped"
+
+// isRequestShapeBatchBudgetReject reports whether a capacity-class rejection
+// is a "batch token budget" reject that classifyRejection proves REQUEST-
+// deterministic: the rejecting provider's reported token budget is not below
+// the model's context window, so the provider's admission cap
+// min(context, budget) was the CONTEXT — the prompt is too big fleet-wide and
+// every provider rejects it identically. Such a reject must arm neither the
+// one-shot budget clamp nor the no-reset capacity-503 rate window
+// (RecordCapacityRejectRequestShape). When the reported budget IS below the
+// context, the binding term may have been this node's memory-pressured KV
+// budget — a genuine provider-specific capacity signal — and the reject feeds
+// the full gray-box state (same discrimination the dispatch failover uses:
+// classifyRejection in inference_failure_class.go, DAR-347).
+//
+// Inputs mirror the dispatch path exactly: providerBudget from the provider's
+// last heartbeat (ReportedTokenBudgetMaxForModel), modelContext from the model
+// registry record. Called only inside the isCapacityRejectStrike branch, so
+// explicit context-overflow strings never reach it (they never strike at all).
+// The string gate keeps the two lookups off every other rejection.
+func (s *Server) isRequestShapeBatchBudgetReject(providerID, model, errStr string) bool {
+	e := strings.ToLower(strings.TrimSpace(errStr))
+	e = strings.ReplaceAll(e, "’", "'")
+	if !strings.Contains(e, "batch token budget") {
+		return false
+	}
+	var providerBudget int64
+	if p := s.registry.GetProvider(providerID); p != nil {
+		providerBudget = p.ReportedTokenBudgetMaxForModel(model)
+	}
+	modelContext := 0
+	if rec, err := s.store.GetModelRegistryRecord(model); err == nil && rec != nil {
+		modelContext = rec.MaxContextLength
+	}
+	return classifyRejection("", errStr, providerBudget, modelContext) == rejectionDeterministicUnservable
+}
 
 // noteInferenceSuccess clears the inference-error strike state for the serving
 // provider-model pair on a clean completion (streaming relay ended without a
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -336,8 +336,12 @@ func (s *Server) noteInferenceSuccess(pr *registry.PendingRequest) {
 	// A clean completion is an ACCEPT for the capacity-reject cooldown: clear
 	// the pair's reject streak, any active capacity cooldown, and the re-trip
 	// backoff. Belt-and-braces with the commit-time accept (commitFirstContent)
-	// and the only accept signal on paths that never stream content.
-	s.registry.RecordCapacityAccept(pr.ProviderID, pr.Model)
+	// and the only accept signal on paths that never stream content. For the
+	// capacity-503 RATE window (capacity_rate.go) one served request must
+	// count exactly ONE outcome, so this completion-time accept only counts
+	// there when no commit-time accept already did (ContentCommittedSafe is
+	// stamped immediately before every commit-time RecordCapacityAccept).
+	s.registry.RecordCapacityAcceptOutcome(pr.ProviderID, pr.Model, !pr.ContentCommittedSafe())
 	// A clean completion proves the node is healthy — close its node-health
 	// breaker (and reset the exponential backoff) if it had tripped.
 	if _, closed := s.registry.RecordProviderOutcome(pr.ProviderID, true, 200, ""); closed {
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -338,10 +338,16 @@ func (s *Server) noteInferenceSuccess(pr *registry.PendingRequest) {
 	// backoff. Belt-and-braces with the commit-time accept (commitFirstContent)
 	// and the only accept signal on paths that never stream content. For the
 	// capacity-503 RATE window (capacity_rate.go) one served request must
-	// count exactly ONE outcome, so this completion-time accept only counts
-	// there when no commit-time accept already did (ContentCommittedSafe is
-	// stamped immediately before every commit-time RecordCapacityAccept).
-	s.registry.RecordCapacityAcceptOutcome(pr.ProviderID, pr.Model, !pr.ContentCommittedSafe())
+	// count exactly ONE outcome, so this completion-time accept re-offers the
+	// outcome only when the commit-time accept did not actually RECORD one
+	// (RateOutcomeCountedSafe — stamped from RecordCapacityAccept's return at
+	// every commit site). Keying on the recorded outcome rather than on
+	// ContentCommittedSafe covers the stream that committed BEFORE the pair's
+	// first windowed reject (nothing recorded then — accepts only store while
+	// a reject is in-window) but was still serving during the rejects: it
+	// enters the denominator here instead of vanishing and letting a few
+	// later 503s overstate a healthy pair's reject rate.
+	s.registry.RecordCapacityAcceptOutcome(pr.ProviderID, pr.Model, !pr.RateOutcomeCountedSafe())
 	// A clean completion proves the node is healthy — close its node-health
 	// breaker (and reset the exponential backoff) if it had tripped.
 	if _, closed := s.registry.RecordProviderOutcome(pr.ProviderID, true, 200, ""); closed {
@@ -3751,7 +3757,9 @@ func (s *Server) handleGenericInference(w http.ResponseWriter, r *http.Request, 
 			// generic path has no preamble filter, so this is the first chunk of
 			// ANY kind rather than strictly first content — fine as an accept
 			// signal: a black hole capacity-rejects, it never emits a chunk.)
-			s.registry.RecordCapacityAccept(provider.ID, pr.Model)
+			if s.registry.RecordCapacityAccept(provider.ID, pr.Model) {
+				pr.MarkRateOutcomeCounted()
+			}
 			committed = true
 		} else {
 			select {
@@ -3824,7 +3832,9 @@ func (s *Server) handleGenericInference(w http.ResponseWriter, r *http.Request, 
 				// capacity-cooldown ACCEPT at first content for the same reason.
 				pr.MarkFirstContentArrived()
 				pr.MarkContentCommitted()
-				s.registry.RecordCapacityAccept(provider.ID, pr.Model)
+				if s.registry.RecordCapacityAccept(provider.ID, pr.Model) {
+					pr.MarkRateOutcomeCounted()
+				}
 				committed = true
 			} else {
 				select {
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -258,8 +258,11 @@ func (s *Server) refundProviderExtra(pr *registry.PendingRequest) {
 //
 // It emits the cool-down metric on the inference-error transition and the
 // provider_breaker_open metric on the node-health transition into quarantine.
-// errStr is the provider's error message ("" for synthetic timeouts).
-func (s *Server) noteInferenceError(providerID string, pr *registry.PendingRequest, statusCode int, errStr string) {
+// errStr is the provider's error message and errReason its structured
+// InferenceErrorMessage.ErrorReason ("" for synthetic timeouts and legacy
+// providers) — the reason feeds the gray-box request-shape classification the
+// same way the dispatch failover trusts it (classifyRejection P1).
+func (s *Server) noteInferenceError(providerID string, pr *registry.PendingRequest, statusCode int, errStr, errReason string) {
 	if providerID == "" || pr == nil {
 		return
 	}
@@ -322,7 +325,7 @@ func (s *Server) noteInferenceError(providerID string, pr *registry.PendingReque
 		switch {
 		case isColdModelMissRejection(errStr):
 			tripped = s.registry.RecordCapacityRejectLifecycle(providerID, pr.Model)
-		case s.isRequestShapeBatchBudgetReject(providerID, pr.Model, errStr):
+		case s.isRequestShapeBatchBudgetReject(providerID, pr.Model, errStr, errReason):
 			tripped = s.registry.RecordCapacityRejectRequestShape(providerID, pr.Model)
 		default:
 			tripped = s.registry.RecordCapacityReject(providerID, pr.Model)
@@ -347,27 +350,33 @@ func (s *Server) noteInferenceError(providerID string, pr *registry.PendingReque
 const metricCapacityCooldownTripped = "routing.capacity_cooldown_tripped"
 
 // isRequestShapeBatchBudgetReject reports whether a capacity-class rejection
-// is a "batch token budget" reject that classifyRejection proves REQUEST-
-// deterministic: the rejecting provider's reported token budget is not below
-// the model's context window, so the provider's admission cap
-// min(context, budget) was the CONTEXT — the prompt is too big fleet-wide and
-// every provider rejects it identically. Such a reject must arm neither the
-// one-shot budget clamp nor the no-reset capacity-503 rate window
+// is PROVEN request-deterministic by classifyRejection: a "batch token budget"
+// reject from a provider whose reported token budget is not below the model's
+// context window (the admission cap min(context, budget) was the CONTEXT — the
+// prompt is too big fleet-wide), or an explicit request_exceeds_context
+// structured reason. Such a reject must arm neither the one-shot budget clamp
+// nor the no-reset capacity-503 rate window
 // (RecordCapacityRejectRequestShape). When the reported budget IS below the
 // context, the binding term may have been this node's memory-pressured KV
 // budget — a genuine provider-specific capacity signal — and the reject feeds
 // the full gray-box state (same discrimination the dispatch failover uses:
 // classifyRejection in inference_failure_class.go, DAR-347).
 //
-// Inputs mirror the dispatch path exactly: providerBudget from the provider's
-// last heartbeat (ReportedTokenBudgetMaxForModel), modelContext from the model
-// registry record. Called only inside the isCapacityRejectStrike branch, so
-// explicit context-overflow strings never reach it (they never strike at all).
-// The string gate keeps the two lookups off every other rejection.
-func (s *Server) isRequestShapeBatchBudgetReject(providerID, model, errStr string) bool {
+// Inputs mirror the dispatch path exactly: the structured errReason
+// (InferenceErrorMessage.ErrorReason — a provider that says
+// request_exceeds_node_budget / capacity_busy is TRUSTED over the stale
+// heartbeat-budget heuristic, so a stale snapshot that still reads >= context
+// cannot misroute a genuine node-capacity failure away from the gray-box
+// trackers), providerBudget from the provider's last heartbeat
+// (ReportedTokenBudgetMaxForModel), and modelContext from the model registry
+// record. Called only inside the isCapacityRejectStrike branch, so explicit
+// context-overflow STRINGS never reach it (they never strike at all). The
+// cheap gate keeps the two lookups off every other rejection.
+func (s *Server) isRequestShapeBatchBudgetReject(providerID, model, errStr, errReason string) bool {
 	e := strings.ToLower(strings.TrimSpace(errStr))
 	e = strings.ReplaceAll(e, "’", "'")
-	if !strings.Contains(e, "batch token budget") {
+	reason := strings.ToLower(strings.TrimSpace(errReason))
+	if !strings.Contains(e, "batch token budget") && reason != "request_exceeds_context" {
 		return false
 	}
 	var providerBudget int64
@@ -378,7 +387,7 @@ func (s *Server) isRequestShapeBatchBudgetReject(providerID, model, errStr strin
 	if rec, err := s.store.GetModelRegistryRecord(model); err == nil && rec != nil {
 		modelContext = rec.MaxContextLength
 	}
-	return classifyRejection("", errStr, providerBudget, modelContext) == rejectionDeterministicUnservable
+	return classifyRejection(errReason, errStr, providerBudget, modelContext) == rejectionDeterministicUnservable
 }
 
 // noteInferenceSuccess clears the inference-error strike state for the serving
@@ -436,9 +445,9 @@ func (s *Server) noteInferenceSuccess(pr *registry.PendingRequest) {
 // so arms where cancelDispatch did refund are safe, and a failed pre-commit
 // attempt never reaches settlement (its channels are closed and it is neither
 // pending nor parked), so this can never double-credit against a settle.
-func (s *Server) noteDispatchProviderError(provider *registry.Provider, pr *registry.PendingRequest, statusCode int, errStr string, held *[]string) (discardedHeld bool) {
+func (s *Server) noteDispatchProviderError(provider *registry.Provider, pr *registry.PendingRequest, statusCode int, errStr, errReason string, held *[]string) (discardedHeld bool) {
 	if provider != nil {
-		s.noteInferenceError(provider.ID, pr, statusCode, errStr)
+		s.noteInferenceError(provider.ID, pr, statusCode, errStr, errReason)
 	}
 	s.refundProviderExtra(pr)
 	if held == nil || len(*held) == 0 {
@@ -1657,7 +1666,7 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r 
 				case errMsg, ok := <-pr.ErrorCh:
 					if ok && errMsg.Error != "" {
 						s.refundReservedBalance(pr, "provider_error:"+pr.RequestID)
-						s.noteInferenceError(pr.ProviderID, pr, errMsg.StatusCode, errMsg.Error)
+						s.noteInferenceError(pr.ProviderID, pr, errMsg.StatusCode, errMsg.Error, errMsg.ErrorReason)
 						s.ddIncr("inference.in_band_error", []string{"model:" + pr.Model, "reason:provider_error"})
 						s.updateInferenceRouteOutcomeForPending(pr, postCommitProviderErrorOutcome(pr, errMsg))
 						errData, _ := json.Marshal(map[string]any{
@@ -1801,7 +1810,7 @@ func (s *Server) handleStreamingResponseWithFirstChunk(w http.ResponseWriter, r 
 				continue
 			}
 			s.refundReservedBalance(pr, "provider_error:"+pr.RequestID)
-			s.noteInferenceError(pr.ProviderID, pr, errMsg.StatusCode, errMsg.Error)
+			s.noteInferenceError(pr.ProviderID, pr, errMsg.StatusCode, errMsg.Error, errMsg.ErrorReason)
 			s.ddIncr("inference.in_band_error", []string{"model:" + pr.Model, "reason:provider_error"})
 			s.updateInferenceRouteOutcomeForPending(pr, postCommitProviderErrorOutcome(pr, errMsg))
 			errData, _ := json.Marshal(map[string]any{
@@ -1892,7 +1901,7 @@ func (s *Server) handleResponsesStreamingResponseWithFirstChunk(w http.ResponseW
 				continue
 			}
 			s.refundReservedBalance(pr, "provider_error:"+pr.RequestID)
-			s.noteInferenceError(pr.ProviderID, pr, errMsg.StatusCode, errMsg.Error)
+			s.noteInferenceError(pr.ProviderID, pr, errMsg.StatusCode, errMsg.Error, errMsg.ErrorReason)
 			s.ddIncr("inference.in_band_error", []string{"model:" + pr.Model, "reason:provider_error"})
 			s.updateInferenceRouteOutcomeForPending(pr, postCommitProviderErrorOutcome(pr, errMsg))
 			emitter.emitError("provider_error", errMsg.Error)
@@ -1934,7 +1943,7 @@ func (s *Server) handleNonStreamingResponseWithFirstChunk(w http.ResponseWriter,
 				case errMsg, ok := <-pr.ErrorCh:
 					if ok && errMsg.Error != "" {
 						s.refundReservedBalance(pr, "provider_error:"+pr.RequestID)
-						s.noteInferenceError(pr.ProviderID, pr, errMsg.StatusCode, errMsg.Error)
+						s.noteInferenceError(pr.ProviderID, pr, errMsg.StatusCode, errMsg.Error, errMsg.ErrorReason)
 						s.updateInferenceRouteOutcomeForPending(pr, preResponseProviderErrorOutcome(pr, errMsg))
 						statusCode := errMsg.StatusCode
 						if statusCode == 0 {
@@ -2064,7 +2073,7 @@ func (s *Server) handleNonStreamingResponseWithFirstChunk(w http.ResponseWriter,
 				continue
 			}
 			s.refundReservedBalance(pr, "provider_error:"+pr.RequestID)
-			s.noteInferenceError(pr.ProviderID, pr, errMsg.StatusCode, errMsg.Error)
+			s.noteInferenceError(pr.ProviderID, pr, errMsg.StatusCode, errMsg.Error, errMsg.ErrorReason)
 			s.updateInferenceRouteOutcomeForPending(pr, preResponseProviderErrorOutcome(pr, errMsg))
 			statusCode := errMsg.StatusCode
 			if statusCode == 0 {
@@ -3825,7 +3834,7 @@ func (s *Server) handleGenericInference(w http.ResponseWriter, r *http.Request, 
 				s.sendProviderCancel(provider, requestID)
 				refundExtra()
 				refundReservation()
-				s.noteInferenceError(provider.ID, pr, errMsg.StatusCode, errMsg.Error)
+				s.noteInferenceError(provider.ID, pr, errMsg.StatusCode, errMsg.Error, errMsg.ErrorReason)
 				s.updateInferenceRouteOutcomeForPending(pr, preCommitProviderErrorOutcome(pr, errMsg))
 				statusCode := errMsg.StatusCode
 				if statusCode == 0 {
@@ -3844,7 +3853,7 @@ func (s *Server) handleGenericInference(w http.ResponseWriter, r *http.Request, 
 		s.sendProviderCancel(provider, requestID)
 		refundExtra()
 		refundReservation()
-		s.noteInferenceError(provider.ID, pr, errMsg.StatusCode, errMsg.Error)
+		s.noteInferenceError(provider.ID, pr, errMsg.StatusCode, errMsg.Error, errMsg.ErrorReason)
 		s.updateInferenceRouteOutcomeForPending(pr, preCommitProviderErrorOutcome(pr, errMsg))
 		statusCode := errMsg.StatusCode
 		if statusCode == 0 {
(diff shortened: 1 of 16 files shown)


coordinator/registry/budget_clamp.go
```diff
@@ -0,0 +1,207 @@
...
+// cannot shed the clamp; migrated on identity rebind (migrateFaultStateLocked)
+// and deliberately NOT cleared by Disconnect. Map is bounded by the same
+// opportunistic >1024 sweep as the sibling cooldown maps. All state is guarded
+// by r.mu; the routing-path reads happen under r.mu held in either mode.
+const (
+	// envBudgetClamp is the kill switch. Default ON; set to false/0 to restore
+	// pre-clamp behavior (stale-heartbeat admission).
+	envBudgetClamp = "EIGENINFERENCE_BUDGET_CLAMP"
+	// envBudgetClampTTLSecs caps how long a clamp can hold without release —
+	// the fail-open bound. Default 300s.
```
ethenotethan, line 73: 🔵 **[INFO]** 🔒 Environment variable EIGENINFERENCE_BUDGET_CLAMP_TTL_SECONDS not validated for negative values

💡 _Suggestion:_ Add validation to ensure TTL is positive before converting to time.Duration

coordinator/api/consumer.go
```diff
@@ -336,8 +336,12 @@ func (s *Server) noteInferenceSuccess(pr *registry.PendingRequest) {
 	// A clean completion is an ACCEPT for the capacity-reject cooldown: clear
 	// the pair's reject streak, any active capacity cooldown, and the re-trip
 	// backoff. Belt-and-braces with the commit-time accept (commitFirstContent)
-	// and the only accept signal on paths that never stream content.
-	s.registry.RecordCapacityAccept(pr.ProviderID, pr.Model)
+	// and the only accept signal on paths that never stream content. For the
+	// capacity-503 RATE window (capacity_rate.go) one served request must
+	// count exactly ONE outcome, so this completion-time accept only counts
+	// there when no commit-time accept already did (ContentCommittedSafe is
+	// stamped immediately before every commit-time RecordCapacityAccept).
```
ethenotethan: 🟡 **[MEDIUM]** ⚡ Redundant ContentCommittedSafe() call in hot path

💡 _Suggestion:_ Cache the result of pr.ContentCommittedSafe() since it's called twice in the same function and the state won't change between calls

📊 _Score: 2×4 = 8 · Category: repeated work_

ethenotethan: 🟡 **[MEDIUM]** 🏷️ pr.ContentCommittedSafe() called without type assertion check

💡 _Suggestion:_ Add type assertion check: if committed, ok := pr.ContentCommittedSafe(); ok && !committed { ... } or ensure ContentCommittedSafe() method handles nil receiver safely

📊 _Score: 4×3 = 12 · Category: missing_type_assertion_

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 1 finding(s)
- 🔵 **[INFO]** `coordinator/registry/budget_clamp.go:67-72` — Environment variable EIGENINFERENCE_BUDGET_CLAMP_TTL_SECONDS not validated for negative values
  - _Suggestion:_ Add validation to ensure TTL is positive before converting to time.Duration

- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:343` — Redundant ContentCommittedSafe() call in hot path
  - _Suggestion:_ Cache the result of pr.ContentCommittedSafe() since it's called twice in the same function and the state won't change between calls

### Type_diligence — 1 finding(s) (1 blocking)
- 🟡 **[MEDIUM]** `coordinator/api/consumer.go:343` — pr.ContentCommittedSafe() called without type assertion check
  - _Suggestion:_ Add type assertion check: if committed, ok := pr.ContentCommittedSafe(); ok && !committed { ... } or ensure ContentCommittedSafe() method handles nil receiver safely

_3 finding(s) total, 2 blocking. Verdict: REQUEST_CHANGES._

coordinator/registry/budget_clamp.go
ethenotethan: :white_check_mark: Looks addressed in `98fc1ae` — Added validation to ensure TTL is positive by checking if cfg.TTL <= 0 and resetting to default if invalid.. Resolving.

coordinator/api/consumer.go
ethenotethan: :white_check_mark: Looks addressed in `98fc1ae` — The code no longer calls ContentCommittedSafe() method, replacing it with RateOutcomeCountedSafe() which appears to be a safer alternative.. Resolving.

coordinator/registry/budget_clamp.go
```diff
@@ -0,0 +1,263 @@
...
```
ethenotethan, line 73: 🔵 **[INFO]** 🔒 Environment variable TTL parsing lacks bounds validation

💡 _Suggestion:_ Add validation to ensure TTL is within reasonable bounds (e.g., 1 second to 24 hours) to prevent extremely large or zero values that could cause operational issues

📊 _Score: 2×2 = 4 · Category: missing-input-validation_

coordinator/registry/budget_clamp.go
```diff
@@ -0,0 +1,263 @@
...
+			}
+		}
+	}
+	if prev, ok := r.budgetClamps[key]; ok {
+		budgetReported = budgetReported || prev.budgetReported
+	}
+	r.budgetClamps[key] = &budgetClampEntry{clampedAt: now, budgetReported: budgetReported}
+}
+
+// providerReportsTokenBudgetLocked reports whether the provider's CURRENT
```
ethenotethan, line 170: 🟡 **[MEDIUM]** ⚡ Opportunistic sweep creates O(n) allocation on every capacity reject when map exceeds 1024 entries

💡 _Suggestion:_ Consider using a separate cleanup goroutine or batching cleanup operations to avoid blocking the hot path

coordinator/registry/capacity_rate.go
```diff
@@ -0,0 +1,200 @@
...
+// cooldown sweeps — churned session-keyed identities are never re-keyed).
+func sweepCapacityRateMapLocked(m map[capacityRejectKey][]time.Time, now time.Time) {
+	if len(m) <= 1024 {
+		return
+	}
+	for key, outcomes := range m {
+		if len(outcomes) == 0 || now.Sub(outcomes[len(outcomes)-1]) >= capacityRateWindow {
+			delete(m, key)
+		}
+	}
```
ethenotethan, line 147: 🟡 **[MEDIUM]** ⚡ Map sweep iterates entire map on every record operation when size exceeds 1024

💡 _Suggestion:_ Use background cleanup or limit sweep frequency to avoid O(n) work on request hot path

coordinator/registry/capacity_rate.go
```diff
@@ -0,0 +1,200 @@
...
+// appendWindowedOutcome slides the window (keeps only in-window timestamps)
+// and appends the new outcome, reusing the backing array.
+func appendWindowedOutcome(outcomes []time.Time, now time.Time) []time.Time {
+	kept := outcomes[:0]
+	for _, ts := range outcomes {
+		if now.Sub(ts) < capacityRateWindow {
+			kept = append(kept, ts)
+		}
+	}
+	return append(kept, now)
```
ethenotethan, line 133: 🔵 **[INFO]** ⚡ appendWindowedOutcome creates new slice on every call, potentially copying large arrays

💡 _Suggestion:_ Pre-allocate slice capacity based on expected window size or use circular buffer for better memory efficiency

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Security — 1 finding(s)
- 🔵 **[INFO]** `coordinator/registry/budget_clamp.go:67-72` — Environment variable TTL parsing lacks bounds validation
  - _Suggestion:_ Add validation to ensure TTL is within reasonable bounds (e.g., 1 second to 24 hours) to prevent extremely large or zero values that could cause operational issues

- 🟡 **[MEDIUM]** `coordinator/registry/budget_clamp.go:140-160` — Opportunistic sweep creates O(n) allocation on every capacity reject when map exceeds 1024 entries
  - _Suggestion:_ Consider using a separate cleanup goroutine or batching cleanup operations to avoid blocking the hot path
- 🟡 **[MEDIUM]** `coordinator/registry/capacity_rate.go:137-147` — Map sweep iterates entire map on every record operation when size exceeds 1024
  - _Suggestion:_ Use background cleanup or limit sweep frequency to avoid O(n) work on request hot path
- 🔵 **[INFO]** `coordinator/registry/capacity_rate.go:123-133` — appendWindowedOutcome creates new slice on every call, potentially copying large arrays
  - _Suggestion:_ Pre-allocate slice capacity based on expected window size or use circular buffer for better memory efficiency

_4 finding(s) total, 2 blocking. Verdict: REQUEST_CHANGES._

coordinator/api/consumer.go
Gajesh2007: The premise was off — `noteInferenceSuccess` called `ContentCommittedSafe()` once (the other match at consumer.go:342-343 is the doc comment; provider.go:1892 is a different function/goroutine). As of 98fc1aef the call site is gone anyway: the completion-time accept now keys on `RateOutcomeCountedSafe()` (see the Codex thread on this line for why). One guarded bool read per completion either way — nothing repeats, so there is nothing to cache.

coordinator/api/consumer.go
```diff
@@ -336,8 +336,12 @@ func (s *Server) noteInferenceSuccess(pr *registry.PendingRequest) {
...
+	s.registry.RecordCapacityAcceptOutcome(pr.ProviderID, pr.Model, !pr.ContentCommittedSafe())
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 98fc1aef: `RecordCapacityAccept`/`RecordCapacityAcceptOutcome` now return whether the rate outcome was actually stored; all three commit sites stamp the request (`MarkRateOutcomeCounted`) when the offer recorded, and `noteInferenceSuccess` re-offers at completion only when the commit-time offer recorded nothing (`!RateOutcomeCountedSafe()` replaces `!ContentCommittedSafe()`). A stream that commits before the pair's first windowed reject and completes during the reject window now enters the denominator exactly once, and a commit-recorded request cannot double-count. Regression: `TestCapacityRatePreRejectCommitCountsAtCompletion` (8 rejects + a pre-reject-committed stream → rate 8/9 over 9 samples, plus the double-count guard).

coordinator/registry/scheduler.go
```diff
@@ -1118,6 +1148,18 @@ func (r *Registry) snapshotProviderLockedEx(p *Provider, model string, traits Re
...
+	// Gray-box budget clamp (budget_clamp.go): when a capacity-503 has proven
+	// the pair's live gate is rejecting, admission must not believe the
+	// stale-optimistic heartbeat budget. Only budget-reporting slots are
+	// clamped; p.LastHeartbeat is when the CURRENT BackendCapacity was
+	// delivered (Heartbeat stamps both in one critical section), which is what
+	// the release-freshness check compares against the clamp time. p.mu and
+	// r.mu are both held here (see lock discipline above).
+	if snap.activeTokenBudgetMax > 0 {
+		rawRemaining := snap.activeTokenBudgetMax - snap.activeTokenBudgetUsed - snap.queuedTokenBudget
+		snap.budgetClamped = r.budgetClampActiveLocked(p.ID, model, p.LastHeartbeat, rawRemaining, now)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 98fc1aef: `budgetClampEntry` now records — sticky across re-arms — whether the pair was budget-reporting when clamped (`providerReportsTokenBudgetLocked` at arm time), and `budgetClampActiveLocked` holds such a pair through any budgetless snapshot, including the reconnect window where `BackendCapacity == nil` until the first heartbeat, until the post-clamp heartbeat release proof or the TTL. Both snapshot sites now evaluate the clamp unconditionally and `freeMemoryAdmits` checks `budgetClamped` before the budget branch, so the budgetless window can no longer fall through to the legacy memory-admission path. Never-budget-reporting legacy pairs keep their documented exemption (a single 503 still cannot gate them). Regression: `TestBudgetClampHoldsAcrossReconnectBeforeFirstHeartbeat` — verified to fail against the pre-fix semantics.

coordinator/registry/budget_clamp.go
Gajesh2007: Declining the upper bound: zero/negative values already fall back to the 5-minute default (`loadBudgetClampConfig`: `if cfg.TTL <= 0 { cfg.TTL = defaultBudgetClampTTL }`). The TTL is an operator-set fail-open bound read once at coordinator construction; a deliberately large value only lengthens the MAXIMUM hold for a pair that keeps capacity-rejecting, and both release conditions (post-clamp heartbeat with headroom + accept) still lift the clamp early. An arbitrary cap like 24h would silently override an explicit operator choice without protecting anything.

coordinator/registry/budget_clamp.go
Gajesh2007: Declining: this is the same opportunistic >1024 sweep every sibling breaker map already uses (error_cooldown.go; the strike/cooldown sweeps in `RecordCapacityReject` itself), and it runs on the capacity-REJECT path (a provider 503) — not the serving hot path. It triggers only past 1024 tracked pairs, deletes TTL-expired entries in place (map iterate + delete, zero allocation), and is exactly what bounds session-keyed identity churn. A background cleanup goroutine would add lifecycle and locking for a path that fires orders of magnitude less often than serves.

coordinator/registry/capacity_rate.go
Gajesh2007: Declining — same pattern parity as the clamp-map sweep: `sweepCapacityRateMapLocked` fires only past 1024 pairs, and only on the reject path or the accept-with-rejects-in-window path (the degraded-pair case — `RecordCapacityAcceptOutcome`'s read-lock fast path returns before any of this for healthy pairs, so the serve hot path never sees it). Iterate + delete allocates nothing. Mirrors the sibling cooldown-map sweeps that already shipped.

coordinator/registry/capacity_rate.go
Gajesh2007: Declining as a false positive: `appendWindowedOutcome` reuses the backing array — `kept := outcomes[:0]` is the filter-in-place idiom — so the steady path creates no new slice; an allocation only happens when append must grow past existing capacity, and each pair's slice is bounded by the 5-minute window. A circular buffer would be more code for the same bound.

Gajesh2007:
Review round addressed @ethenotethan — ready for re-review.

- **Fixed** (98fc1aef): both Codex findings — the budget clamp now holds through a reconnect's pre-heartbeat window (budgetless snapshots no longer shed onto the legacy memory path; legacy never-budget-reporting pairs keep their exemption), and pre-reject-committed streams now enter the capacity-rate denominator exactly once (`RecordCapacityAccept` returns whether the outcome recorded; completion re-offers only when it didn't). Regression tests for both, verified failing pre-fix.
- **Declined with evidence** (per-thread replies): TTL bounds (zero/negative already clamped to default), the ContentCommittedSafe double-call/type-assertion findings (single guarded bool read; no assertion exists), and the >1024 sweep / appendWindowedOutcome allocation findings (sibling-map pattern parity; filter-in-place reuses the backing array).
- CI: the E2E `TestIntegration_ConcurrentRequests` failure is the known provider-side MLX `broadcast_shapes` batching crash — it reproduces on master @ 76447bc0 (run 28758695595), unrelated to this branch.

coordinator/registry/budget_clamp.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	if !budgetReported {
+		// No live budget snapshot to judge release against: hold iff the pair
+		// was budget-reporting when clamped (legacy pairs stay exempt).
+		return e.budgetReported
+	}
+	released := e.acceptedSince &&
+		heartbeatAt.After(e.clampedAt) &&
+		rawBudgetRemaining >= budgetClampReleaseMinHeadroomTokens
+	return !released
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in a85c0fce: `budgetClampActiveLocked` now early-returns `false` whenever `e.budgetReported == false` — before the budgeted release branch — so a clamp armed on a budgetless reject (cold `404 model not loaded`, or a first dispatch before the first capacity heartbeat) can never gate, even after a later heartbeat starts reporting `active_token_budget_max > 0`. Your stranding mechanism was exactly right: the budgeted branch would evaluate release with `acceptedSince=false` while the clamp itself blocked the dispatch that could produce the accept proof, holding a never-lied box until TTL. The sticky-hold reconnect fix is untouched (it lives in the `e.budgetReported == true` + budgetless-snapshot case), and a subsequent genuine reject while budget-reporting re-arms with `budgetReported=true` via the sticky-or and gates normally. Regression: `TestBudgetClampArmedBudgetlessDoesNotGateAfterBudgetAppears` — armed budgetless → budget heartbeat arrives → must stay admittable; verified to fail pre-fix (clamp re-activated until TTL).

coordinator/api/consumer.go
```diff
@@ -336,8 +336,18 @@ func (s *Server) noteInferenceSuccess(pr *registry.PendingRequest) {
...
+	// count exactly ONE outcome, so this completion-time accept re-offers the
+	// outcome only when the commit-time accept did not actually RECORD one
+	// (RateOutcomeCountedSafe — stamped from RecordCapacityAccept's return at
+	// every commit site). Keying on the recorded outcome rather than on
+	// ContentCommittedSafe covers the stream that committed BEFORE the pair's
+	// first windowed reject (nothing recorded then — accepts only store while
+	// a reject is in-window) but was still serving during the rejects: it
+	// enters the denominator here instead of vanishing and letting a few
+	// later 503s overstate a healthy pair's reject rate.
+	s.registry.RecordCapacityAcceptOutcome(pr.ProviderID, pr.Model, !pr.RateOutcomeCountedSafe())
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in a85c0fce: `handleComplete`'s `consumerGone` branch (parked settlement claimed — the streaming handler returned on `r.Context().Done()`) now re-offers the served completion with `RecordCapacityAcceptOutcome(pr.ProviderID, pr.Model, !pr.RateOutcomeCountedSafe())` — the same stamp-keyed bookkeeping `noteInferenceSuccess` does on clean completion, which indeed never runs on this path. A stream that committed before the pair's first windowed reject (commit-time offer recorded nothing) and whose consumer disconnected now enters the denominator at the provider terminal; a commit that already recorded passes `countRateOutcome=false` and cannot double-count. Keyed on `pr.ProviderID` so it matches the commit-time accept and reject keys. Regression: `TestCapacityRateParkedCompletionCountsAtHandleComplete` — commit pre-reject, park via `holdForSettlement`, complete via `handleComplete` → samples 2, rate 0.5; verified to fail pre-fix (samples stuck at 1, the served stream vanished).

coordinator/registry/capacity_cooldown.go
```diff
@@ -167,11 +167,22 @@ func (r *Registry) RecordCapacityReject(providerID, modelID string) (tripped boo
...
+
+	// Gray-box trackers ride the SAME classified entry point but have their own
+	// kill switches, independent of the cooldown threshold: the budget clamp
+	// stops admission believing the pair's stale heartbeat budget immediately
+	// (budget_clamp.go), and the rate window accumulates the reject side of the
+	// capacity-503 rate (capacity_rate.go — accepts deliberately do NOT reset
+	// it, unlike the strike streak below).
+	clampKey := capacityRejectKey{ProviderID: r.faultKeyLocked(providerID), ModelID: modelID}
+	r.recordBudgetClampLocked(clampKey, r.providerReportsTokenBudgetLocked(providerID, modelID), now)
+	r.recordCapacityRateRejectLocked(clampKey, now)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in a85c0fce: cold-load misses no longer derate. `RecordCapacityReject` is split over a shared impl — the existing name stays the derating entry point (cooldown + clamp + rate window), and new `RecordCapacityRejectLifecycle` feeds the black-hole cooldown (a forever-404 box with zero accepts must still trip) and, harmlessly, the clamp (budgetless-armed clamps never gate, per the sibling thread) but skips `recordCapacityRateRejectLocked` entirely. `noteInferenceError` classifies with `isColdModelMissRejection` ("not loaded" / "no model loaded" — mirrors the `capacityClassMarkers` cold-miss vocabulary, evaluated only after `isCapacityRejectStrike`, so request-shape "model not found" 404s remain excluded) and routes cold misses to the lifecycle path; genuine token-budget/queue/headroom 503s still derate. Regressions: `TestCapacityRateLifecycleRejectFeedsCooldownNotRate` (registry: lifecycle rejects trip the cooldown, rate window stays 0/0) and `TestCapacityRateExcludesColdModelMiss` (api: 8 cold 404-then-serve rounds through `noteInferenceError` → rate 0/0, while genuine 503s accumulate); the api test fails pre-fix at rate 0.50 over 16 samples — exactly the ~50% warm-up penalty you predicted.

coordinator/registry/budget_clamp.go
```diff
@@ -0,0 +1,284 @@
...
+		// and THAT gates.
+		return false
+	}
+	if !budgetReported {
+		// Armed while budget-reporting, but the current snapshot carries no
+		// budget (reconnect before the first heartbeat, or the model's slot
+		// missing from the live report): hold — reconnecting must not shed the
+		// clamp onto the legacy memory-admission path (the stable fault key
+		// exists precisely so it can't).
+		return true
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 79a592d3: released clamp entries are now DELETED, so nothing lingers to revive. `dropInactiveBudgetClampSnapshotLocked` removes an entry once it can no longer gate — fully released (accept proof + strictly-fresher heartbeat with ≥1024 raw headroom), TTL-expired, or budgetless-armed — called from both write-lock opportunities: the accept path (`RecordCapacityAcceptOutcome`, covering heartbeat-then-accept order) and a new heartbeat sweep (`releaseBudgetClampsOnHeartbeat` in `Heartbeat`, covering accept-then-heartbeat with NO further traffic — exactly your reconnect-before-TTL scenario). The sweep evaluates the heartbeat's own stamped time and report rather than re-reading `r.providers`, so a disconnect racing the sweep can't void the release proof either. Unreleased clamps still hold through the budgetless pre-heartbeat window (`TestBudgetClampHoldsAcrossReconnectBeforeFirstHeartbeat` green, unchanged). Regressions: `TestBudgetClampReleasedEntryDoesNotReviveOnReconnect` (both release orders → disconnect → budgetless reconnect must admit; fails pre-fix on the lingering entry) and `TestBudgetClampHeartbeatReleaseSurvivesDisconnectRace`.

coordinator/api/consumer.go
```diff
@@ -306,7 +306,19 @@ func (s *Server) noteInferenceError(providerID string, pr *registry.PendingReque
...
+		// capacity dishonesty. It still feeds the black-hole cooldown (a box
+		// that 404s forever with zero accepts is a black hole), but it must NOT
+		// derate the pair's gray-box capacity-503 RATE (capacity_rate.go) — that
+		// window has no accept-reset, so counting a healthy box's normal reloads
+		// would penalize it. Genuine capacity/token-budget 503s derate.
+		var tripped bool
+		if isColdModelMissRejection(errStr) {
+			tripped = s.registry.RecordCapacityRejectLifecycle(providerID, pr.Model)
+		} else {
+			tripped = s.registry.RecordCapacityReject(providerID, pr.Model)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 79a592d3: `noteInferenceError` now classifies "batch token budget" rejects with the same discrimination the dispatch failover uses — `isRequestShapeBatchBudgetReject` feeds `classifyRejection` the rejecting provider's `ReportedTokenBudgetMaxForModel` and the registry record's `MaxContextLength`, and a `rejectionDeterministicUnservable` verdict (budget not below context ⇒ the binding term was the fleet-wide context) routes to new `RecordCapacityRejectRequestShape`: cooldown strikes ONLY — no one-shot clamp, no no-reset rate outcome — so a single oversized prompt can no longer clamp or derate a healthy pair. Node-pressured rejects (reported budget < context) keep feeding the full gray-box state. The cooldown strike is retained deliberately: a budget-misreporting box (huge advertised budget ≥ context) rejects NORMAL prompts with exactly this string and classifies deterministic too — the cooldown's zero-interleaved-accepts discriminator is what makes strikes safe for healthy pairs, a safety the clamp/rate lack by design. Regression: `TestCapacityRejectDeterministicBatchBudgetArmsNoGrayBoxState` (budget 5M ≥ ctx 131072 → no clamp, no rate, 5 strikes still trip; budget 90k < ctx → clamp + rate as before) — fails pre-fix at the clamp assertion.

coordinator/registry/capacity_cooldown.go
```diff
@@ -276,35 +320,76 @@ func (r *Registry) claimCapacityProbeLocked(providerID, modelID string, now time
...
+	// needs the write lock when the pair has rejects in its rate window
+	// (recordCapacityRateAcceptLocked is what builds the denominator, and a
+	// zero-reject pair's rate is 0 regardless — capacityRatePenaltyLocked
+	// fast-exits on rejects==0 — so skipping the append loses nothing).
 	r.mu.RLock()
 	key := capacityRejectKey{ProviderID: r.faultKeyLocked(providerID), ModelID: modelID}
 	_, hasStrikes := r.capacityRejectStrikes[key]
 	_, hasCooldown := r.capacityCooldowns[key]
 	_, hasTrips := r.capacityCooldownTrips[key]
+	_, hasClamp := r.budgetClamps[key]
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in 79a592d3: inactive entries are deleted at both write-lock opportunities, so map presence again means "possibly gating" and the healthy fast path recovers. `dropInactiveBudgetClampSnapshotLocked` removes released / TTL-expired / budgetless-armed entries; it runs on every accept for a pair with an entry (`RecordCapacityAcceptOutcome`) and on every heartbeat whose report covers a clamped model (`releaseBudgetClampsOnHeartbeat` — read-locked probe, write lock only when an entry exists). After one transient reject + recovery, the FIRST post-release accept (or the release heartbeat itself) deletes the entry; subsequent accepts return to the read-lock fast path once the pair's 5-minute rate-reject window decays (that residue is bounded by design, unlike the indefinite clamp entry). The >1024 opportunistic sweep remains as the churned-identity backstop, consistent with the sibling maps. Regression: `TestBudgetClampInactiveEntriesAreDeleted` — TTL-expired and budgetless-armed entries dropped on the next accept, released entries dropped by the heartbeat sweep, and an UNRELEASED entry proven to survive its accepts (the gate must not weaken); fails pre-fix on the lingering-entry assertions.

Gajesh2007:
Two on-the-record notes from the round-3 independent review (PASS, non-blocking — logged here to avoid another push cycle):

1. **Follow-up**: `api/consumer.go:381` passes `reason=""` into the deterministic-reject classifier while the dispatch path threads `msg.ErrorReason`. No live divergence today; once providers ship `request_exceeds_node_budget` (DAR-347), thread `errMsg.ErrorReason` through so both paths classify identically.
2. **Accepted bounded blind spot**: on a ≥DAR-347 provider, a node-pressured 'batch token budget' reject under a stale-optimistic heartbeat (reported budget ≥ model context) arms no clamp/rate for up to one heartbeat interval; the next heartbeat re-syncs and subsequent rejects feed everything. Mirrors `classifyRejection`'s documented LIMITATION; the incident's dominant reject strings are unaffected, and misreporting boxes still accrue cooldown strikes throughout.

---
**UPDATE (c7770544):** Item 1 is now addressed — `errMsg.ErrorReason` is threaded through `noteInferenceError` into the gray-box classifier (`classifyRejection(errReason, …)`), matching the dispatch path; item 2's blind spot narrows accordingly for ≥DAR-347 providers (a structured node-budget reason now overrides the stale snapshot). See the resolved round-4 review thread on `api/consumer.go`.

ethenotethan (commented):

**Verdict:** COMMENT

### Security — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Performance — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Type_diligence — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Additive_complexity — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

✅ All four passes clean. No issues found.

coordinator/registry/budget_clamp.go
```diff
@@ -0,0 +1,386 @@
...
+	// dropping TTL-expired clamps once it grows.
+	if len(r.budgetClamps) > 1024 {
+		for k, e := range r.budgetClamps {
+			if !now.Before(e.clampedAt.Add(r.budgetClampCfg.TTL)) {
+				delete(r.budgetClamps, k)
+			}
+		}
+	}
+	if prev, ok := r.budgetClamps[key]; ok {
+		budgetReported = budgetReported || prev.budgetReported
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in c7770544: the sticky-OR now inherits `prev.budgetReported` only while the previous entry is still inside its TTL (`recordBudgetClampLocked`: `ok && now.Before(prev.clampedAt.Add(TTL))`). A TTL-expired entry is a clamp cycle that already failed open, so a later budgetless reject (cold 404, pre-heartbeat window) re-arms from the CURRENT budgetless read and stays exempt, exactly as the budgetless-armed rule requires; the sticky-OR keeps protecting a LIVE clamp's budgetless reconnect window (its original purpose). All round-3 deletion/exemption tests stay green. Regression: `TestBudgetClampExpiredEntryDoesNotDonateBudgetState` — budget-armed clamp TTL-expires with the entry lingering, snapshot goes budgetless, a fresh reject re-arms → must not gate; fails pre-fix with the pair blocked for another TTL.

coordinator/registry/capacity_cooldown.go
```diff
@@ -161,17 +161,90 @@ type capacityRejectKey struct {
...
+// black hole, caught by the zero-interleaved-accepts discriminator) and,
+// harmlessly, the budget clamp (a budgetless-armed clamp never gates — see
+// budgetClampActiveLocked), but it does NOT derate the pair in the gray-box
+// capacity-503 RATE window. That window deliberately has NO accept-reset, so
+// counting a healthy box's normal reload misses would accumulate a false
+// reject rate and penalize it as if its reported budget were dishonest. The
+// api layer routes cold "not loaded"/"no model loaded" rejections here;
+// genuine capacity/token-budget 503s go to RecordCapacityReject and DO derate.
+func (r *Registry) RecordCapacityRejectLifecycle(providerID, modelID string) (tripped bool) {
+	return r.recordCapacityReject(providerID, modelID, false, true)
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in c7770544: `RecordCapacityRejectLifecycle` no longer touches the clamp at all (`armClamp=false` — same shared impl as the request-shape path). You were right that the budgetless-keyed exemption was only conditionally harmless: a provider that idle-unloads after its last heartbeat still shows the slot budget in the stale snapshot, so `providerReportsTokenBudgetLocked` read true and a routine re-warm 404 (the normal 1h idle-unload + lazy-reload cycle, e.g. gemma) armed a REAL gating clamp — and with dispatch blocked, `acceptedSince` could never be proven, stranding the pair until TTL. The lifecycle classification now takes precedence over snapshot budget presence: a cold miss is a statement about model residency, never about budget honesty. Decision on the cooldown, per the round-3 rationale: lifecycle misses STILL feed it — the forever-404 box (never loads, zero accepts) is a genuine black hole and the zero-interleaved-accepts discriminator makes strikes free for the normal 404→load→serve cycle (first accept clears the streak). Documented in the function comment. Regression: `TestBudgetClampLifecycleMissWithStaleBudgetSnapshotDoesNotClamp` — lifecycle 404 on a stale budget-reporting snapshot → no clamp entry, admission proceeds, cooldown still trips on forever-404, rate stays 0; fails pre-fix at the clamp assertion.

coordinator/api/consumer.go
```diff
@@ -325,6 +346,41 @@ func (s *Server) noteInferenceError(providerID string, pr *registry.PendingReque
...
+	}
+	var providerBudget int64
+	if p := s.registry.GetProvider(providerID); p != nil {
+		providerBudget = p.ReportedTokenBudgetMaxForModel(model)
+	}
+	modelContext := 0
+	if rec, err := s.store.GetModelRegistryRecord(model); err == nil && rec != nil {
+		modelContext = rec.MaxContextLength
+	}
+	return classifyRejection("", errStr, providerBudget, modelContext) == rejectionDeterministicUnservable
```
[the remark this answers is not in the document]
Gajesh2007: Fixed in c7770544 (this also closes the follow-up logged on the PR in comment [#523 (comment)](https://github.com/Layr-Labs/d-inference/pull/523#issuecomment-4919442466), item 1): `InferenceErrorMessage.ErrorReason` is now threaded through `noteInferenceError` (and the `noteDispatchProviderError`/`noteDispatchRetry` wrappers) into `isRequestShapeBatchBudgetReject`, which passes it to `classifyRejection(errReason, errStr, …)` exactly as the dispatch failover does. A DAR-347 provider declaring `request_exceeds_node_budget`/`capacity_busy` is trusted over the stale heartbeat-budget heuristic — a stale snapshot still reading ≥ context can no longer misroute a genuine node-capacity failure to the strike-only path, so it clamps and derates as it should; an explicit `request_exceeds_context` is request-deterministic regardless of string or budget; reasonless legacy rejects keep the existing heuristic unchanged. Regression: `TestGrayBoxClassificationTrustsStructuredReason` (all three cases) — fails pre-fix with the node-budget reject skipping the clamp.

ethenotethan (commented):

**Verdict:** COMMENT

### Security — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Performance — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Type_diligence — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Additive_complexity — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

✅ All four passes clean. No issues found.

ethenotethan (commented):

**Verdict:** COMMENT

### Security — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Performance — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Type_diligence — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

### Additive_complexity — ⚠️ SKIPPED
_Error: OpenRouter HTTP 402: {"error":{"message":"Insufficient credits. Add more using https://openrouter.ai/settings/credits","code":402}}_

✅ All four passes clean. No issues found.

PR #558, opened 2026-07-19
feat(provider): make Gemma 4 MTP production-default

Gajesh2007:
## Outcome

Supported Gemma 4 targets now use verified multi-token prediction by default. A normal provider start/load can fetch, verify, bind, and publish an MTP-active engine in one lifecycle; if the assistant arrives after a target-only slot is already serving, the provider promotes that retained target automatically at an idle boundary without an operator restart.

Every optional failure remains fail-open: the existing target-only engine keeps serving, memory/grants are restored, and MTP is never reported active until the concrete runtime confirms it.

## Behavior and code flow: Before

```mermaid
flowchart LR
  subgraph Behavior_Before[Behavior: Before]
    B1[Provider starts] --> B2[Gemma target loads]
    B2 --> B3{Assistant already cached?}
    B3 -- No --> B4[Publish target-only slot]
    B4 --> B5[Download assistant in background]
    B5 --> B6[Operator restarts provider]
    B6 --> B7[MTP can activate]
    B3 -- Yes --> B7
  end

  subgraph Code_Before[Code: Before]
    C1[backend.mtp default false] --> C2[Beta enable workflow]
    C2 --> C3[SpecDec prepare checks local cache]
    C3 --> C4[Prefetch task has no slot callback]
    C4 --> C5[ModelSlot remains target-only]
  end
```

## Behavior and code flow: After

```mermaid
flowchart LR
  subgraph Behavior_After[Behavior: After]
    A1[Provider establishes liveness] --> A2[Demanded Gemma load]
    A2 --> A3[Fetch fresh spec_dec metadata]
    A3 --> A4{Verified assistant available?}
    A4 -- Cached --> A5[Atomically reserve target plus assistant]
    A4 -- Download --> A5
    A5 --> A6[Load and bind assistant]
    A6 --> A7{Runtime active and headroom safe?}
    A7 -- Yes --> A8[Publish MTP-active slot]
    A7 -- No --> A9[Publish target-only fallback]
    A4 -- Failure --> A9
    A9 --> A10[Bounded retry and periodic rediscovery]
    A10 --> A11[Verified artifact-ready event]
    A11 --> A12[Wait for outer and engine quiescence]
    A12 --> A13[Build fenced replacement over retained target]
    A13 --> A14{Replacement active and safe?}
    A14 -- Yes --> A15[Atomically swap; retire old residency]
    A14 -- No --> A9
  end

  subgraph Code_After[Code: After]
    D1[BackendSettings default-on policy] --> D2[SpecDecArtifactFunnel prepareForLoad]
    D2 --> D3[Generation-ordered verified resolver]
    D3 --> D4[ProviderLoop MTP promotion policy]
    D4 --> D5[Atomic GlobalKV reservation]
    D5 --> D6[Re-slice gate and unregistered replacement build]
    D6 --> D7[EngineV2 runtime activation check]
    D7 --> D8[Runtime registration and ModelSlot swap]
    D8 --> D9[Old bridge/assistant release before survivor growth]
  end
```

## Default policy and exact opt-out precedence

1. `DARKBLOOM_CBV2_MTP=0` has highest precedence and forces target-only behavior.
2. `darkbloom start` persists that allowlisted variable into the launchd plist, so watchdog, login, and ordinary restart relaunches retain the rollback.
3. Explicit `[backend] mtp = false` is the persistent config opt-out.
4. Explicit `mtp = true` and an absent key are default-on. Existing configs without the key migrate without a rewrite; deliberate old `false` values remain off.
5. Non-Gemma and unsupported model families remain target-only and never enter assistant resolution.

MTP was removed from `darkbloom beta`; the remaining beta registry no longer exposes or mutates it.

## One-start verified assistant lifecycle

- Cold Gemma loads fetch fresh catalog identity before model-load/re-slice locks.
- `spec_dec` metadata must bind immutable prefix, manifest digest, config digest, size, count, roles, and revision.
- Cached bytes are cryptographically revalidated without CDN traffic; empty caches use bounded, deduplicated download and immutable publication.
- Mandatory target residency is reserved first. Optional assistant growth is then admitted atomically against all live unified-memory promises.
- The assistant is loaded with its own quantization, bound to the exact serving target, and included in slot sizing and fleet re-slicing.
- Active MTP forces contiguous KV because transactional sliding-window rollback is required.
- Publication happens only after runtime `mtp.active=true` and the post-build headroom guard passes.

Optional fleet artifact work does not block daemon liveness, local endpoint startup, or coordinator registration. Only a demanded Gemma cold load awaits its bounded preparation.

## Automatic late-download promotion

- Artifact completion is generation-ordered; stale catalog results cannot overtake newer revisions.
- Fast retries are bounded, then low-frequency rediscovery continues so long outages do not require a restart.
- Promotion signals deduplicate, retain the newest pending revision, and cap construction attempts per immutable artifact identity.
- The provider waits for both outer reservations and concrete engine queues to become quiescent.
- Capacity reports `reloading` while promotion fences new work.
- The retained target container is reused; target weights are not downloaded again.
- A replacement is built unregistered, runtime activation and headroom are verified, then runtime registration and `ModelSlot` swap occur atomically.
- The old bridge, assistant, and cache residency are retired before the re-slice gate is released or survivor grants grow.

Unload, update drain, and shutdown cancel and await owned promotion/preparation work. Cancellation does not consume a promotion attempt.

## Fail-open and observability guarantees

Stable low-cardinality reasons cover metadata, manifest, digest, download timeout/cancellation, memory, re-slice floor, construction, headroom, busy timeout, and inactive-engine fallbacks. Logs report policy, active state, source/revision, download+verification duration, assistant bytes, real promotion attempt/result/duration, and cumulative proposed/accepted/emitted counters. No prompts or token IDs are emitted.

## Verification

Passed on the final reviewed diff:

- `swift test --no-parallel --filter MTPResliceFallbackTests` — 15 lifecycle/promotion tests.
- `swift test --no-parallel --filter SpecDecArtifactFunnelTests` — 14 ordering/backoff/shutdown tests.
- `swift test --no-parallel --filter SpecDecResolverTests` — 22 live-isolated HTTP/filesystem verification tests.
- `swift test --no-parallel --filter ProviderMTPFactoryTests` — 10 factory/ownership tests.
- `swift test --no-parallel --filter SpecDecCapacity` — 18 memory/accounting tests.
- `swift test --no-parallel --filter MTPConfig` — 15 policy/config/non-Gemma tests.
- `swift test --no-parallel --filter standalonePendingLoadReservesWeightsBeforeAwaitingContainer` — standalone atomic pending-load path.
- `swift test --no-parallel --filter StartupPreloader` — 7 preload tests.
- `swift test --no-parallel --filter CBv2MTP` in `libs/mlx-swift-lm` — 75 engine tests.
- `swift test --no-parallel --filter Gemma4MTP` in `libs/mlx-swift-lm` — 17 Gemma MTP tests.
- Supervised real local QAT target + QAT-4bit assistant load/bind passed via `scripts/run-mtp-benchmark.py`; final launch fingerprint `25c093a0045df9e218fde66be2a2a7f7f40e5fcdd379dfec9bcb3e042eb8a717`.
- `swift build -c release --product darkbloom` passed.
- `git diff --check`, IDE diagnostics, commit hook, and pre-push hook passed.
- Two independent same-model lifecycle/memory reviews passed the final current diff after all findings were fixed.

Existing performance evidence retained by the implementation contract: QAT-4bit assistant fixed-k=1 production sweeps measured positive aggregate decode TPS at B=1/B=2/B=4 on M4 Max and M5 Max, while unsafe/unprofitable shapes select depth zero.

## Rollout and rollback

No rollout, release, tag, deploy, catalog write, or production mutation is included here.

Roll out by shipping the provider normally; supported catalog targets with valid `metadata.spec_dec` converge automatically. Observe MTP active state, fallback reason, assistant bytes, promotion results, acceptance, committed goodput, memory, and TTFT before broad fleet expansion.

Emergency per-machine rollback:

```bash
darkbloom stop
DARKBLOOM_CBV2_MTP=0 darkbloom start
```

The launchd plist retains the opt-out across later restarts. A persistent config rollback is:

```toml
[backend]
mtp = false
```

## Known limitations

- Production MTP remains target-authoritative greedy decode; stochastic sampling, structured constraints, and unsupported request traits use ordinary target-only decoding.
- MTP uses contiguous KV; paged sliding-window speculative transactions remain a separately gated future capability.
- Promotion waits for an idle/drain-safe boundary and is capped at three construction attempts per immutable assistant identity; a new revision/hash resets the ledger.
- The monolithic provider `swift test` process has a pre-existing cross-suite quiescence interaction after thousands of passing tests (reproduced outside affected suites). All affected suites, the isolated stall suite, release build, real QAT load/bind, commit hook, and pre-push aggregate gate pass.

Made with [Cursor](https://cursor.com)

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/558"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787039305&installation_id=133247490&pr_number=558&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F558&signature=f103bec8d9ea35fdb2fb2ad2e03a0c90d0666cfd1e89627ec97e12c28451b803"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
+++ b/provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
@@ -1,0 +1,282 @@
+import Foundation
+import MLX
+
+private struct SpecDecPromotionFailure: Error {
+    let reason: MTPFallbackReason
+}
+
+extension ProviderLoop {
+    static let specDecPromotionIdleTimeout: Duration = .seconds(300)
+    static let specDecPromotionPollInterval: Duration = .milliseconds(250)
+
+    func runSpecDecPromotion(
+        modelId: String,
+        artifact: SpecDecArtifact,
+        generation: UInt64,
+        attempt: Int,
+        taskID: UUID
+    ) async {
+        let started = ContinuousClock.now
+        logger.info(
+            "mtp: promotion model=\(modelId) attempt=\(attempt) result=waiting_for_idle "
+                + "source=\(artifact.source.rawValue) revision=\(artifact.revision) "
+                + "assistant_bytes=\(artifact.residentBytes)")
+
+        while true {
+            guard promotionMayContinue(
+                modelId: modelId, generation: generation, taskID: taskID)
+            else {
+                restoreCancelledPromotionAttempt(
+                    modelId: modelId, artifact: artifact)
+                finishSpecDecPromotionTask(modelId: modelId, taskID: taskID)
+                return
+            }
+            let observedBridge = modelSlots[modelId]?.engineV2
+            let engineBusy: Bool
+            if let observedBridge {
+                engineBusy = !(await observedBridge.isQuiescentForReplacement())
+            } else {
+                engineBusy = true
+            }
+            // Re-read actor-owned admission state AFTER the bridge await. A
+            // request may have entered while the actor was reentrant.
+            let remoteBusy = requestToModel.values.contains(modelId)
+            let localBusy = hasLocalReservation(modelId)
+            if !remoteBusy, !localBusy, !engineBusy, !isLoadingAny,
+                modelSlots[modelId]?.engineV2 === observedBridge
+            {
+                break
+            }
+            if ContinuousClock.now - started >= Self.specDecPromotionIdleTimeout {
+                logSpecDecPromotionResult(
+                    modelId: modelId, artifact: artifact,
+                    result: "fallback",
+                    reason: .promotionBusyTimeout,
+                    attempt: attempt,
+                    started: started)
+                finishSpecDecPromotionTask(modelId: modelId, taskID: taskID)
+                return
+            }
+            do {
+                try await taskSleep(Self.specDecPromotionPollInterval)
+            } catch {
+                restoreCancelledPromotionAttempt(
+                    modelId: modelId, artifact: artifact)
+                finishSpecDecPromotionTask(modelId: modelId, taskID: taskID)
+                return
+            }
+        }
+
+        guard promotionMayContinue(
+            modelId: modelId, generation: generation, taskID: taskID),
+            let originalSlot = modelSlots[modelId]
+        else {
+            restoreCancelledPromotionAttempt(
+                modelId: modelId, artifact: artifact)
+            finishSpecDecPromotionTask(modelId: modelId, taskID: taskID)
+            return
+        }
+        let originalBridge = originalSlot.engineV2
+        modelsPromotingSpecDec.insert(modelId)
+        isLoadingAny = true
+        await originalBridge.beginRecoveryReload()
+        let reservationID = "mtp-promotion:\(modelId)"
+        var gateHeld = false
+        var replacementBundle: ProviderEngineBundle?
+        var previousGrants: [ExistingSlotGrant] = []
+        var preparedAssistant: ProviderMTPAssistantHandle?
+        var replacementRegistered = false
+
+        do {
+            guard await kvBudget.reserveBytes(
+                requestID: reservationID,
+                bytes: artifact.residentBytes)
+            else {
+                throw SpecDecPromotionFailure(reason: .assistantMemoryUnavailable)
+            }
+
+            let slotLogger = logger
+            let prepared = try await EngineV2SlotFactory.prepareProductionModel(
+                modelId: modelId,
+                isVLM: originalSlot.isVLM,
+                modelDirectory: ModelScanner.resolveLocalPath(modelID: modelId),
+                container: originalSlot.container,
+                specDecPreparation: .init(
+                    artifact: artifact, status: .candidate(artifact)),
+                assistantLoader: engineV2SlotHooks?.assistantLoader
+                    ?? Gemma4ProviderMTPAssistantLoader(),
+                emitTelemetry: engineV2SlotHooks?.emitTelemetry,
+                logInfo: { slotLogger.info($0) },
+                logWarning: { slotLogger.warning($0) })
+            guard let assistant = prepared.assistant, prepared.mtpStatus.active else {
+                throw SpecDecPromotionFailure(
+                    reason: prepared.mtpStatus.reason ?? .assistantLoadFailed)
+            }
+            preparedAssistant = assistant
+
+            guard promotionMayContinue(
+                modelId: modelId, generation: generation, taskID: taskID),
+                modelSlots[modelId]?.engineV2 === originalBridge
+            else { throw CancellationError() }
+
+            await acquireResliceGate()
+            gateHeld = true
+            guard promotionMayContinue(
+                modelId: modelId, generation: generation, taskID: taskID),
+                modelSlots[modelId]?.engineV2 === originalBridge
+            else { throw CancellationError() }
+
+            let sizing = originalSlot.sizing.replacingAuxiliaryWeightBytes(
+                prepared.assistantBytes)
+            previousGrants = await existingSlotGrants(excludingModelId: "")
+            let budget = fleetKVBudgetBytes(
+                extraWeightBytes: sizing.weightsBytes,
+                replacingModelId: modelId)
+            let targets = EngineV2KVSizing.resliceGrants(
+                existing: previousGrants.map(\.slot),
+                newcomer: nil,
+                fleetKVBudgetBytes: budget)
+            guard EngineV2KVSizing.resliceMeetsServiceabilityFloor(
+                targets, fixedCarveBytes: [:])
+            else {
+                throw SpecDecPromotionFailure(reason: .assistantResliceFloor)
+            }
+            for entry in previousGrants {
+                if let target = targets[entry.slot.modelId],
+                    target < entry.previousGrant
+                {
+                    await entry.bridge.updateKVBytesCapacity(target)
+                }
+            }
+
+            let replacement = try await makeEngineV2BundleForSlot(
+                modelId: modelId,
+                modelType: originalSlot.modelType,
+                isVLM: originalSlot.isVLM,
+                modelDirectory: ModelScanner.resolveLocalPath(modelID: modelId),
+                container: originalSlot.container,
+                tokenizer: originalSlot.tokenizer,
+                sizing: sizing,
+                kvBytesCapacity: targets[modelId] ?? 0,
+                specDecPreparation: .init(
+                    artifact: artifact, status: .candidate(artifact)),
+                preparedModel: prepared,
+                cacheEligibleWeightHash: originalSlot.cacheEligibleWeightHash,
+                registerRuntime: false,
+                logConstruction: false)
+            replacementBundle = replacement
+
+            MLX.Memory.clearCache()
+            let postBuildServeable = KVHeadroomProbe.postBuildServeable(
+                kvBackendKind: replacement.bridge.kvBackendKind,
+                pagedPoolBytes: await replacement.bridge.kvBackendPoolBytes())
+            let runtimeActive = engineV2SlotHooks != nil
+                ? replacement.mtpStatus.active
+                : await replacement.bridge.mtpStatusSnapshot().active
+            guard runtimeActive else {
+                throw SpecDecPromotionFailure(reason: .engineInactive)
+            }
+            guard postBuildServeable else {
+                throw SpecDecPromotionFailure(reason: .assistantPostBuildHeadroom)
+            }
+            await replacement.bridge.beginRecoveryReload()
+            guard promotionMayContinue(
+                modelId: modelId, generation: generation, taskID: taskID),
+                modelSlots[modelId]?.engineV2 === originalBridge
+            else { throw CancellationError() }
+
+            await engineV2Runtime.register(
+                modelId: modelId, bridge: replacement.bridge)
+            replacementRegistered = true
+            guard promotionMayContinue(
+                modelId: modelId, generation: generation, taskID: taskID),
+                modelSlots[modelId]?.engineV2 === originalBridge
+            else { throw CancellationError() }
+
+            modelSlots[modelId] = ModelSlot(
+                engineBundle: replacement,
+                container: originalSlot.container,
+                tokenizer: originalSlot.tokenizer,
+                sizing: sizing,
+                cacheEligibleWeightHash: originalSlot.cacheEligibleWeightHash,
+                isVLM: originalSlot.isVLM,
+                modelType: originalSlot.modelType,
+                lastInferenceAt: originalSlot.lastInferenceAt)
+            replacementBundle = nil
+            preparedAssistant = nil
+            await originalBridge.shutdown()
+            originalSlot.engineBundle.releaseAssistant()
+            MLX.Memory.clearCache()
+            await kvBudget.release(requestID: reservationID)
+            for entry in previousGrants {
+                if let target = targets[entry.slot.modelId],
+                    target > entry.previousGrant,
+                    entry.slot.modelId != modelId
+                {
+                    await entry.bridge.updateKVBytesCapacity(target)
+                }
+            }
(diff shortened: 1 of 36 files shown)


provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
```diff
@@ -0,0 +1,282 @@
...
+        var preparedAssistant: ProviderMTPAssistantHandle?
+        var replacementRegistered = false
+
+        do {
+            guard await kvBudget.reserveBytes(
+                requestID: reservationID,
+                bytes: artifact.residentBytes)
+            else {
+                throw SpecDecPromotionFailure(reason: .assistantMemoryUnavailable)
+            }
```
ethenotethan, line 96: 🟡 **[MEDIUM]** ⚡ Memory reservation without release on early return path

💡 _Suggestion:_ Add `await kvBudget.release(requestID: reservationID)` before the early return on line 96 to prevent memory leak when promotion is cancelled

provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
```diff
@@ -0,0 +1,282 @@
...
+                logWarning: { slotLogger.warning($0) })
+            guard let assistant = prepared.assistant, prepared.mtpStatus.active else {
+                throw SpecDecPromotionFailure(
+                    reason: prepared.mtpStatus.reason ?? .assistantLoadFailed)
+            }
+            preparedAssistant = assistant
+
+            guard promotionMayContinue(
+                modelId: modelId, generation: generation, taskID: taskID),
+                modelSlots[modelId]?.engineV2 === originalBridge
```
ethenotethan, line 119: 🟡 **[MEDIUM]** ⚡ Memory reservation without release on early return path

💡 _Suggestion:_ Add `await kvBudget.release(requestID: reservationID)` before throwing CancellationError on line 119 to prevent memory leak when promotion is cancelled

provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
```diff
@@ -0,0 +1,282 @@
...
+            else { throw CancellationError() }
+
+            await acquireResliceGate()
+            gateHeld = true
+            guard promotionMayContinue(
+                modelId: modelId, generation: generation, taskID: taskID),
+                modelSlots[modelId]?.engineV2 === originalBridge
+            else { throw CancellationError() }
+
+            let sizing = originalSlot.sizing.replacingAuxiliaryWeightBytes(
```
ethenotethan, line 129: 🟡 **[MEDIUM]** ⚡ Memory reservation without release on early return path

💡 _Suggestion:_ Add `await kvBudget.release(requestID: reservationID)` before throwing CancellationError on line 129 to prevent memory leak when promotion is cancelled

provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
```diff
@@ -0,0 +1,282 @@
...
+            }
+            guard postBuildServeable else {
+                throw SpecDecPromotionFailure(reason: .assistantPostBuildHeadroom)
+            }
+            await replacement.bridge.beginRecoveryReload()
+            guard promotionMayContinue(
+                modelId: modelId, generation: generation, taskID: taskID),
+                modelSlots[modelId]?.engineV2 === originalBridge
+            else { throw CancellationError() }
+
```
ethenotethan, line 187: 🟡 **[MEDIUM]** ⚡ Memory reservation without release on early return path

💡 _Suggestion:_ Add `await kvBudget.release(requestID: reservationID)` before throwing CancellationError on line 187 to prevent memory leak when promotion is cancelled

provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
```diff
@@ -0,0 +1,282 @@
...
+            await engineV2Runtime.register(
+                modelId: modelId, bridge: replacement.bridge)
+            replacementRegistered = true
+            guard promotionMayContinue(
+                modelId: modelId, generation: generation, taskID: taskID),
+                modelSlots[modelId]?.engineV2 === originalBridge
+            else { throw CancellationError() }
+
```
ethenotethan, line 195: 🟡 **[MEDIUM]** ⚡ Memory reservation without release on early return path

💡 _Suggestion:_ Add `await kvBudget.release(requestID: reservationID)` before throwing CancellationError on line 195 to prevent memory leak when promotion is cancelled

ethenotethan (changes requested):

**Verdict:** REQUEST_CHANGES

### Performance — 5 finding(s) (5 blocking)
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift:94-96` — Memory reservation without release on early return path
  - _Suggestion:_ Add `await kvBudget.release(requestID: reservationID)` before the early return on line 96 to prevent memory leak when promotion is cancelled
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift:117-119` — Memory reservation without release on early return path
  - _Suggestion:_ Add `await kvBudget.release(requestID: reservationID)` before throwing CancellationError on line 119 to prevent memory leak when promotion is cancelled
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift:127-129` — Memory reservation without release on early return path
  - _Suggestion:_ Add `await kvBudget.release(requestID: reservationID)` before throwing CancellationError on line 129 to prevent memory leak when promotion is cancelled
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift:185-187` — Memory reservation without release on early return path
  - _Suggestion:_ Add `await kvBudget.release(requestID: reservationID)` before throwing CancellationError on line 187 to prevent memory leak when promotion is cancelled
- 🟡 **[MEDIUM]** `provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift:193-195` — Memory reservation without release on early return path
  - _Suggestion:_ Add `await kvBudget.release(requestID: reservationID)` before throwing CancellationError on line 195 to prevent memory leak when promotion is cancelled

_5 finding(s) total, 5 blocking. Verdict: REQUEST_CHANGES._

provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
Gajesh2007: Declined: this guard throws only when reserveBytes returned false, so no reservation exists to release. The enclosing catch also performs the single cleanup call on every later failure path.

provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
Gajesh2007: Declined: this throws inside the single enclosing do/catch. The catch unconditionally calls kvBudget.release(requestID:) after assistant/replacement cleanup and grant restoration. Adding a local release would duplicate cleanup.

provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
Gajesh2007: Declined: this throws inside the single enclosing do/catch. The catch unconditionally calls kvBudget.release(requestID:) after assistant/replacement cleanup and grant restoration. Adding a local release would duplicate cleanup.

provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
Gajesh2007: Declined: this throws inside the single enclosing do/catch. The catch unconditionally calls kvBudget.release(requestID:) after shutting down any replacement, releasing assistant ownership, clearing cache, and restoring grants.

provider-swift/Sources/ProviderCore/ProviderLoop+MTPPromotion.swift
Gajesh2007: Declined: this throws inside the single enclosing do/catch after runtime registration. The catch restores the original runtime registration or unregisters, tears down the replacement, restores grants, and unconditionally releases reservationID.

Gajesh2007:
@codex review this

Gajesh2007:
Status for v0.8.0: **deferring, not closing** — and splitting.

**The default flip should not ship in 0.8.0.** The throughput premise behind `mtp = true` is currently negative on the backend the release is named for. The MTP depth controller mis-attributes seed cost — `CBv2MTPRoundDriver.swift:472` computes `attributed = wallTimeNanos &+ claimedSeedCostNanos`, but the early return at :470 never credits the seed's emitted token — so depth-1 looks systematically overpriced and the controller under-speculates. Forcing k=1 past the controller gains +14.7% on contiguous (128.33 vs 111.90 tok/s, gemma-4 26B / M4 Max) but **loses 5.8% on paged** (106.53 vs 113.10), because paged carries an additive ~15.5% per-verify-round penalty. Defaulting the fleet on would default it into a slower configuration on PagedAttention.

**Dismissing the 5 blocking findings from @ethenotethan.** All five "memory reservation without release on early return" sites sit inside a single `do { … } catch { }`, and the catch calls `await kvBudget.release(requestID:)` unconditionally at line 250. The `:96` site throws only when `reserveBytes` returned false, which per `GlobalKVCacheBudget.swift:197-200` means no reservation was created. The suggested local releases would double-release — harmless (`:178` guards `removeValue != nil`) but pure noise. These five alone produced the CHANGES_REQUESTED verdict.

**The genuinely blocking item is the one that got no reply:** Codex's P1 quiescence race at `ProviderLoop+MTPPromotion.swift:45`. The wait loop takes the engine snapshot *first*, then re-reads admission state; a request admitted during the bridge await and then cancelled is removed from `requestToModel` before the engine retires its row, so both observations read idle while the row is live. The identity check at line 47 detects bridge *replacement*, not row *occupancy*, and the actual fence (`modelsPromotingSpecDec.insert`) is at line 81 — after the loop. Fix is to hoist the fence above the wait and order the engine snapshot last.

**Codex's P1 on foreground downloads is now a regression, not just a comment.** Master and integration already ship the background-only funnel it asked for (`SpecDecArtifactFunnel.swift:165-169`). This PR's `prepareForLoad` reintroduces a 120s foreground `resolver.prefetch` ahead of coordinator registration (`ProviderLoop+Serve.swift:118` precedes `:153`).

**Split:** `increaseReservationIfAvailable`, the capacity quiescence accessor, and `MTPResliceFallbackTests` can land independently. The promotion subsystem reworks post-0.8.0 with the race fixed and the foreground download dropped. The config default flip waits on the seed-cost fix plus a paged replay showing default-on is a net win.

MTP stays available as a beta opt-in (`[backend] mtp = true`) throughout.

PR #573, opened 2026-07-25
test(bench): one-command Gemma 4 continuous-batching benchmark + recorded baseline

Gajesh2007:
## Summary

Adds a one-command, fail-closed benchmark for the Gemma 4 26B-A4B QAT 4-bit continuous-batching path, and commits the pre-optimization baseline it compares against.

```bash
make benchmark-gemma-contbatch
make benchmark-gemma-contbatch GEMMA_BENCHMARK_ARGS="--label my-change --iterations 3"
```

One invocation: builds the release provider, refreshes the matching metallib, runs four sweeps through the **production** `EngineV2Factory.makeProductionEngine` entry point, diffs against the committed baseline, and writes timestamped Markdown + JSON to the gitignored `tmp/benchmarks/`.

## Why it fails closed

A benchmark that reports a number for a run that did not actually happen is worse than no benchmark. The runner refuses results on: model-id or matrix mismatch, wrong cardinality, non-positive metrics, decode-token reconstruction mismatch, TTFT iteration mismatch, arrival topology/row/token mismatch, checksum failure, broken output invariance. It recomputes every median it prints rather than trusting the payload.

Two measurement decisions worth reviewing:

- **Arrival patterns share one warmed engine.** A fresh engine per pattern charges asynchronous compiled-decode startup to whichever pattern ran first, which is a measurement artifact, not a property of the arrival topology. Pattern order is rotated across repetitions.
- **Both decode-window and end-to-end TPS are reported.** Staggered arrivals put prefill gaps inside the decode window, so decode-window TPS alone overstates what the workload actually delivered.

## Baseline recorded

`mlx-community/gemma-4-26B-A4B-it-qat-4bit` @ `0e3cbab3`, Apple M4 Max, 40-core GPU, 128 GB.

| Prompt | Raw prefill | Production TTFT |
|---:|---:|---:|
| 128 | 839.9 tok/s | 207.6 ms |
| 512 | 1,260.8 tok/s | 474.9 ms |
| 2,048 | 1,292.8 tok/s | 1,800.4 ms |

Decode aggregate: B1 107.8, B2 160.2, B3 192.3, B4 205.1 tok/s. 4 × 512 burst TTFT 1,881.6 ms. Output tokens were exactly invariant across all four arrival topologies.

Committed in both human-readable and machine-readable form so later runs diff automatically.

## Before / After

```mermaid
flowchart LR
  subgraph Before
    A1[perf question] --> B1[hand-run darkbloom benchmark]
    B1 --> C1[eyeball stdout]
    C1 --> D1[numbers in a scrollback buffer]
    D1 --> E1[no baseline, no invariance check]
  end
  subgraph After
    A2[make benchmark-gemma-contbatch] --> B2[release build + matching metallib]
    B2 --> C2[4 sweeps via production EngineV2]
    C2 --> D2{strict validation}
    D2 -- reject --> F2[non-zero exit, no report]
    D2 -- accept --> G2[diff vs committed baseline]
    G2 --> H2["tmp/benchmarks/*.md + *.json"]
  end
```

```mermaid
flowchart TB
  subgraph Code
    M[Makefile: benchmark-gemma-contbatch] --> S[scripts/benchmark-gemma-contbatch.py]
    S --> R[gemma_contbatch.runner]
    R --> C[config: CLI + matrix]
    R --> P[process: build, metallib, darkbloom]
    R --> V[validation: fail-closed checks]
    R --> A[results: medians + baseline diff]
    A --> O[reporting: markdown + json]
    P --> B1["darkbloom benchmark --sweep"]
    P --> B2["darkbloom benchmark --scheduler-prefill"]
    P --> B3["darkbloom benchmark --arrival-invariance"]
    B3 --> N[ArrivalInvarianceBenchmark.swift NEW]
  end
```

## Notes

- Generated reports land in `tmp/benchmarks/` and are gitignored; only the baseline docs under `docs/reports/` are committed.
- The Python runner is split into focused modules (`config`, `process`, `validation`, `results`, `reporting`, `runner`), each well under 300 lines, per the repo's modularity guidance.
- This PR is measurement only — no serving-path behavior changes.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/573"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787549336&installation_model_id=21733&pr_number=573&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F573&signature=bfc78e924b53f1812ac350e2ed7801e10c900e01a6bb589eb4ddbc5908ecbb2b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 .PHONY: help \
         coordinator-test coordinator-build coordinator-build-linux coordinator \
         prompt-sidecar-format prompt-sidecar-check prompt-sidecar-test prompt-sidecar-build prompt-sidecar \
-        provider-build provider-test provider \
+        provider-build provider-test provider benchmark-gemma-contbatch \
         ui-install ui-build ui-lint ui-test ui \
         e2e-integration e2e-benchmark e2e \
         test build all clean
@@ -51,6 +51,9 @@ provider-test: ## swift test for the Swift provider CLI
 	cd provider-swift && swift test
 
 provider: provider-build provider-test ## Build + test provider
+
+benchmark-gemma-contbatch: ## Build and benchmark Gemma 4 26B continuous batching
+	python3 scripts/benchmark-gemma-contbatch.py $(GEMMA_BENCHMARK_ARGS)
 
 # ---- Console UI (Next.js 16) ----------------------------------------------
 
--- a/docs/reports/2026-07-23-gemma-4-26b-qat4bit-continuous-batching-baseline.json
+++ b/docs/reports/2026-07-23-gemma-4-26b-qat4bit-continuous-batching-baseline.json
@@ -1,0 +1,185 @@
+{
+  "schemaVersion": 1,
+  "name": "gemma-4-26b-qat4bit-continuous-batching-baseline-2026-07-23",
+  "modelID": "mlx-community/gemma-4-26B-A4B-it-qat-4bit",
+  "sourceCommit": "0fb9d2b689e4934caf01c7242c1d2ea992ebeff5",
+  "modelSnapshot": "0e3cbab38ce568cf6e23543010d08d03b731910c",
+  "hardware": {
+    "chipName": "Apple M4 Max",
+    "gpuCores": 40,
+    "memoryGb": 128,
+    "memoryBandwidthGbs": 546
+  },
+  "configuration": {
+    "iterations": 3,
+    "prefillLengths": [128, 512, 2048],
+    "decodePromptTokens": 64,
+    "decodeTokens": 64,
+    "arrivalPromptTokens": 512,
+    "arrivalDecodeTokens": 64,
+    "maxBatch": 4,
+    "compiledDecode": true,
+    "prefixCache": false,
+    "mtp": false,
+    "kvBackend": "contiguous"
+  },
+  "summary": {
+    "prefill": [
+      {
+        "promptTokens": 128,
+        "medianElapsedMs": 152.391916,
+        "medianTokensPerSecond": 839.9395673980503,
+        "samples": [
+          {"elapsedMs": 167.561042, "tokensPerSecond": 763.9007162536027},
+          {"elapsedMs": 152.359334, "tokensPerSecond": 840.1191882343093},
+          {"elapsedMs": 152.391916, "tokensPerSecond": 839.9395673980503}
+        ]
+      },
+      {
+        "promptTokens": 512,
+        "medianElapsedMs": 406.099958,
+        "medianTokensPerSecond": 1260.7733389620296,
+        "samples": [
+          {"elapsedMs": 407.260792, "tokensPerSecond": 1257.1796992429363},
+          {"elapsedMs": 406.099958, "tokensPerSecond": 1260.7733389620296},
+          {"elapsedMs": 405.871041, "tokensPerSecond": 1261.4844329334646}
+        ]
+      },
+      {
+        "promptTokens": 2048,
+        "medianElapsedMs": 1584.189417,
+        "medianTokensPerSecond": 1292.774701069727,
+        "samples": [
+          {"elapsedMs": 1584.269666, "tokensPerSecond": 1292.7092173460828},
+          {"elapsedMs": 1578.109625, "tokensPerSecond": 1297.7552177340026},
+          {"elapsedMs": 1584.189417, "tokensPerSecond": 1292.774701069727}
+        ]
+      }
+    ],
+    "schedulerTTFT": [
+      {
+        "promptTokens": 128,
+        "medianTTFTMs": 207.611875,
+        "samples": [
+          {"ttftMs": 214.070875, "msPerPrefillToken": 1.685597440944882},
+          {"ttftMs": 207.611875, "msPerPrefillToken": 1.6347391732283465},
+          {"ttftMs": 201.863583, "msPerPrefillToken": 1.5894770314960631}
+        ]
+      },
+      {
+        "promptTokens": 512,
+        "medianTTFTMs": 474.925542,
+        "samples": [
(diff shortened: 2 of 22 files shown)


Gajesh2007:
All five findings were legitimate and are fixed in `9127746`. Four of them meant the harness could report a confident number for something it had not actually measured — the exact failure this tooling exists to prevent — so thank you, these were good catches.

**Codex P1 — decode measured once regardless of `--iterations`.** Correct, and the most consequential: the flag only repeated *prefill lengths*, while the decode curve ran once per process and `summarize` copied that single sample and called it a median. I had already reported "3-sample" decode deltas off this path; they were single-sample. Fixed by adding `--decode-iterations` to the Swift sweep (looping repetition-outer/batch-inner so thermal drift is shared across batch sizes rather than dumped on the last one), feeding the derived metrics per-batch medians instead of the first sample, and requiring exactly `--iterations` samples per batch size in validation. Chose in-process repetition over re-invoking the sweep, which would reload 15.6 GB of weights per iteration for no extra signal.

**Codex P1 — model snapshot not pinned.** Correct. `ModelScanner.resolveLocalPath` selects the newest snapshot by mtime, so a fresh pull of the same model ID silently became "an engine delta". The snapshot is now derived from all three raw payloads, *required to agree across them* (they are separate processes), and matched against the baseline.

**Codex P1 — hardware not pinned.** Correct; `baseline["hardware"]` was dead data. Chip, GPU cores, memory and bandwidth are now validated, and I additionally pinned `maxBatch` and `prefillLengths`, which were also unchecked.

**Codex P1 — staggered arrivals drift.** Correct, including the observation that validation trusted the constant. Worse than reported: `burst` had no timing enforcement at all, since `delayMs == 0` skipped the sleep entirely. Arrivals now sleep to an absolute deadline from the scenario start, every submission offset is recorded, samples outside a 5 ms tolerance are discarded and retried, and the run fails if the delivered topology is not the named one. The Python side re-derives and re-checks all of it. Tolerance is one fifth of the tightest inter-arrival gap, and the validator rejects an override that would exceed half that gap.

**Copilot — `compare()` silently drops non-overlapping rows.** Correct, and it was the worst failure mode: the report still called itself a baseline diff while hiding exactly the rows with no counterpart. Now hard-fails symmetrically, including on duplicate keys.

**Copilot — `@unchecked Sendable`.** Correct; `CBv2Engine` is declared `Sendable`, so plain conformance compiles clean under Swift 6 strict concurrency.

**Copilot — `lookup_comparison` annotation.** Correct, widened to `str | int`.

**Known remaining gap, disclosed rather than hidden:** the committed baseline was produced by the pre-fix binary, so its decode figures are single samples and its arrival numbers carry no delivered-topology evidence. A median-of-3 run compared against it is still not like-for-like on decode. The honest fix is re-recording the baseline with the hardened harness against the unmodified engine; I can do that as a follow-up commit here.

Rebased onto current master.

Gajesh2007:
Second round addressed in `a6ed22f`, rebased onto current master (which now includes #574).

**Codex P2 — pin performance overrides before comparing.** Valid, and the most useful of this batch. The runner captured `MLX_METAL_*`/`DARKBLOOM_CBV2_*` into metadata and then never read it. That matters immediately, because the engine now ships per-feature kill switches and A/B-ing them is precisely the workflow that would silently mis-attribute a feature's effect as a code delta. Now wired into the baseline pins.

Worth noting the finding was *understated*: the existing allowlist did not even capture `DARKBLOOM_GEMMA4_PREFILL_*`, `MLX_GATHER_QMM_EXPERT_SLICES` or `DARKBLOOM_EXPERT_TILES_GATEUP`, so pinning alone would not have caught the scenario — the allowlist had to grow too, and every entry was verified against code that actually reads it. An absent baseline block is read as "engine defaults": absent + empty run compares, absent + override refuses.

**Copilot — no tests for the decode median.** Valid. That math now decides every decode number, so it should not have been untested. Added a suite pinning: an outlier first sample is ignored, even counts average the two middle values (matching CPython `statistics.median`, which the Python side applies to the same data), grouping is order-independent, `derived` consumes medians rather than the first observation, and single-iteration output is byte-identical. Each assertion carries a counterfactual that reproduces the old bug when fed raw samples, so the tests provably discriminate.

**Copilot — baseline cites out-of-tolerance arrival error.** Valid, and a genuine contradiction: the doc quoted 5–36 ms while the hardened harness now *rejects* anything over 5 ms. Rather than quietly delete the inconvenient numbers, the doc now states which figures came from the pre-fix harness and that the current harness would refuse to reproduce them.

**Copilot — `assert_finite` ImportError.** Not valid. `validation.py` re-exports it from `checks.py`, so `from .validation import assert_finite` resolves; I confirmed by importing the module rather than reasoning about it. I did switch to importing from `.checks` directly, since relying on a re-export was needlessly indirect.

**Codex P1 — regenerate the baseline with repeated decode samples.** Agreed, and still outstanding. It is the last real gap: the committed baseline's decode rows are single samples and its arrival rows carry no delivered-topology evidence. Re-recording has to happen against the *unmodified* engine, so it must be done before Layr-Labs/mlx-swift-lm#84 merges. Doing that next.

PR #574, opened 2026-07-25
fix(metallib): key the cache on mlx working-tree contents, not just HEAD

Gajesh2007:
## Summary

`scripts/fetch-metallib.sh` keyed its cache on the mlx submodule's **HEAD commit** plus the deployment target. Kernel work happens in the **working tree** long before it is committed, so any uncommitted change to mlx kernels silently reused a metallib built from *different sources* than the host code dispatching into it.

## Why this is a correctness bug, not a cache-hygiene nit

Hit while patching `qmm_t_impl`'s `M` parameter (`const constant int&` → `const int`). The host encoded the argument one way; the cached metallib expected the other. Result: **the GPU wedged at 0% CPU with no error message**, which presented as a hung test suite and cost a debugging cycle chasing a phantom regression.

The worse failure mode is silent: a benchmark run in that state reports confident numbers for kernels that were never built. Any uncommitted-kernel A/B measured through this script was untrustworthy.

## Fix

Fold uncommitted state into the cache key — tracked diffs plus untracked file contents, hashed and appended as a `-w<hash>` suffix:

```
mlx-<sha>-dt<target>.metallib            # clean tree (unchanged key)
mlx-<sha>-w<hash>-dt<target>.metallib    # dirty tree
```

A clean tree hashes to the empty-input digest and keeps its **existing** key, so current caches stay valid and committed builds are byte-for-byte unaffected. A dirty tree now rebuilds once per distinct source state and prints which content hash it keyed on.

## Before / After

```mermaid
flowchart LR
  subgraph Before
    A1[edit mlx kernel source] --> B1["key = HEAD sha + target"]
    B1 --> C1{cache hit?}
    C1 -- yes --> D1[reuse PRE-EDIT metallib]
    D1 --> E1[host/kernel ABI mismatch]
    E1 --> F1[GPU hang at 0% CPU<br/>or silently wrong benchmark]
  end
  subgraph After
    A2[edit mlx kernel source] --> B2["key = HEAD sha + tracked diff<br/>+ untracked contents + target"]
    B2 --> C2{cache hit?}
    C2 -- no, tree changed --> D2[rebuild metallib from actual sources]
    C2 -- yes, identical tree --> E2[reuse correct metallib]
    D2 --> F2[host and kernel agree]
    E2 --> F2
  end
```

## Verification

Reproduced the stale hit (`Using cached metallib for d5a240408508`, hash `2e535253…`) against patched sources, applied the fix, and confirmed the same tree now rebuilds to a different metallib (`6b3099c9…`). Clean-tree runs still resolve to the original cache entry.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/574"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787549365&installation_model_id=21733&pr_number=574&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F574&signature=81cd3d04b63bb5c8192c3349028af33bfbac488af285ef933eb23c859a8910dc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/scripts/fetch-metallib.sh
+++ b/scripts/fetch-metallib.sh
@@ -43,40 +43,65 @@ test -f "$MLX_SRC/mlx/version.h" || {
   exit 1
 }
 
-MLX_SHA="$(git -C "$MLX_SRC" rev-parse HEAD 2>/dev/null || echo nogit)"
-
 # The commit alone is NOT a sufficient cache key: kernel work happens in the
 # WORKING TREE long before it is committed, and serving a metallib built from
 # different sources than the host code that dispatches into it is not a stale
 # cache — it is a hang. (Observed: patching `qmm_t_impl`'s M parameter while
 # reusing the pre-patch metallib wedges the GPU at 0% CPU.) So fold any
 # uncommitted change — tracked diffs AND untracked files — into the key.
-MLX_TREE_HASH="$({
-    git -C "$MLX_SRC" diff HEAD -- 2>/dev/null || true
-    git -C "$MLX_SRC" ls-files --others --exclude-standard 2>/dev/null \
-        | while IFS= read -r f; do
+#
+# `git diff` is normalized for determinism: --binary so binary edits actually
+# change the hash instead of collapsing to "Binary files differ",
+# --no-ext-diff so a user's difftool cannot alter the key, and --no-color so
+# terminal settings cannot either.
+if git -C "$MLX_SRC" rev-parse --is-inside-work-tree >/dev/null 2>&1; then
+    MLX_SHA="$(git -C "$MLX_SRC" rev-parse HEAD 2>/dev/null || echo nogit)"
+    MLX_TREE_HASH="$({
+        git -C "$MLX_SRC" diff HEAD --binary --no-ext-diff --no-color -- 2>/dev/null || true
+        {
+            git -C "$MLX_SRC" ls-files --others --exclude-standard 2>/dev/null || true
+        } | while IFS= read -r f; do
             printf '%s\n' "$f"
             cat "$MLX_SRC/$f" 2>/dev/null || true
         done
-} | shasum -a 256 | cut -d' ' -f1)"
-# Hash of empty input == a clean tree; keep those cache entries keyed exactly
-# as before so existing caches stay valid.
+    } | shasum -a 256 | cut -d' ' -f1)"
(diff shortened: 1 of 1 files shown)


Gajesh2007:
All four findings were legitimate and are fixed in `580b472`.

**Codex P2 — legacy clean-tree entries.** Correct, and it is the sharpest of the four: the previous script wrote the bare `mlx-<sha>-dt<target>` name *even when the tree was dirty*, so a later clean-tree run could adopt a metallib built from modified sources — resurrecting exactly the mismatch this PR exists to prevent. Clean trees now key on `-c2`, retiring those entries. Verified: a clean tree rebuilds rather than reusing the legacy name, and produces the identical content hash `2e535253…`.

**Copilot — non-git checkout.** Correct. A `git` failure made the tree hash collapse to the empty digest, i.e. "clean", so a vendored copy or source tarball reintroduced the original bug. That path now hashes the kernel/header/source contents. Verified a one-byte edit changes the key: `518904a591ec` → `d5edab320bc6`.

**Copilot — `set -euo pipefail` abort + diff determinism.** Correct on both. The `ls-files | while` pipeline is guarded, and `git diff` now runs with `--binary --no-ext-diff --no-color` so binary edits actually alter the hash instead of collapsing to "Binary files differ", and no difftool or color setting can perturb the key.

**Copilot — ambiguous log line.** Fixed; both the build and cache-hit paths now print the full key, e.g. `Using cached metallib mlx-d5a2404…-c2-dt26.2`.

Rebased onto current master.

PR #579, opened 2026-07-25
docs: record the prefill and fleet performance findings

Gajesh2007:
## Summary

Records the findings from shipping v0.7.15 and then measuring production to decide what to do next. Documentation only — no code changes.

Three conclusions contradict what we believed going in, and re-deriving them is expensive, so they are written down with the evidence and the file references.

## The corrections

**"MoE runs at 3 TFLOPS vs 13 achievable" is arithmetically impossible.** At 3.0 TFLOPS the expert stage alone would take 487 ms, but the entire measured 512-token TTFT is 380 ms. Experts actually run at ~9.2 TFLOPS. The gap is exactly tile padding — `inflation(C) = 1 + 256/C`, which predicts 1.500 at C=512 against 1.465 measured. So the remaining MoE win is ~1.11×, not 4×.

**Packing and chunk growth are the same thing to the MoE.** `Gemma4Experts` flattens B and S together, so `[4,512]` and `[1,2048]` are the *same* call to `SwitchGLU` and reach the same kernel. Packing changed `R`, not dispatch — which means it helped concurrent bursts only, and most production chunks run `B=1`.

**Head-of-line blocking from giant prompts causes ~0% of the p90/p99 tail.** `4 rows × 512 tokens` is exactly `maxBatchedTokensPerStep`, so every running row gets its full chunk every step. A 124k-token prompt is a *longer-lived* blocker, not a harder one. The tail is own prompt size: above p99, 74.8% of routes are >20k-token prompts. This kills SJF/size-aware admission as a lever — its entire ceiling is 639 ms of p50.

## What outranks the engine work

Two production findings are each worth more than the entire remaining 1.30× engine ceiling:

- **71 hardware-trusted, ≥36 GB machines with gemma-4 already on disk are barred from gemma traffic** by a catalog-composition rule (`registry/dedicated_models.go:96`), purely because gpt-oss is also installed. Meanwhile gemma queues **22.9%** and gpt-oss queues **0.0%**. Fixable with no code change.
- **Big prompts land on slow silicon** across a 2.9× fleet throughput spread; 39.2% of ≥8k-token traffic goes to sub-800 tok/s chips. Routing them to fast hardware is worth **p90 −16%, p99 −34%**.

Also documented: prefix caching is off by deliberate operator decision, and gemma-4's 26,624-token donation floor means it could never benefit at our prompt lengths even if re-enabled.

## Instrumentation section

Most of the wrong turns in this session came from metrics that do not mean what their names suggest. That section exists so the next person does not repeat them:

- `actual_ttft_ms` is provider-side and **already excludes queue wait**
- `backlog_ms` / `queue_ms` / `pending_ms` are **router scoring penalties, not milliseconds**
- `providers.id` **churns per registration** — 5,381 rows resolve to 348 machines; always dedupe by `serial_number`
- `cache_affinity_key` is **blanked by a DB trigger**
- `telemetry_events` has been **stale since 2026-05-28**
- the main checkout carries a **stale submodule**, so engine analysis there concludes shipped features do not exist

## Verification note

v0.7.15's production impact is recorded as measured, not as claimed: regressing `TTFT = intercept + slope × prompt_tokens` on gemma routes gives **fixed cost −23%, per-token −4.1%** (−22% at 128 tokens, −17% at p50, −8% at p90), which corroborates the synthetic benchmark on independent traffic. The residual confound — 0.7.15 machines running slightly lighter — is stated rather than omitted.

Open items are listed at the end, including the unverified live coordinator config, the unreconstructed prefix-cache incident, and the fact that `tmp/patches/` holds ~900 lines of ported kernel work that **is not in version control**.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/579"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787557900&installation_model_id=21733&pr_number=579&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F579&signature=7eda375e611fc262856bbc53b5a1b509ac7a123ac2f1e92ea24ae1caa37be4bb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/reports/2026-07-25-prefill-and-fleet-performance-findings.md
+++ b/docs/reports/2026-07-25-prefill-and-fleet-performance-findings.md
@@ -1,0 +1,461 @@
+# Prefill and Fleet Performance — Findings, 2026-07-25
+
+Working notes from the session that shipped provider **v0.7.15** (CBv2 prefill
+stack) and then measured the production system to decide what to do next.
+
+Everything here is either measured or read out of the source. Where a number is
+inferred rather than observed, it says so. Several widely-assumed things turned
+out to be wrong; those are called out explicitly, because re-deriving them is
+expensive.
+
+---
+
+## 1. What shipped in v0.7.15
+
+Merged: `d-inference` #573, #574, #576, #577, #578; `mlx-swift-lm` #84
+(`b177c35`). Tagged `v0.7.15`, registered with the coordinator at 07:12 UTC.
+
+| Change | Mechanism |
+|---|---|
+| Prompt-output narrowing | Prompt chunks no longer build `[B, L, 262144]` logits for positions the engine never reads. Intermediate chunks project nothing; the frontier projects one row. |
+| Final-layer tail pruning + last-query | The last decoder layer keeps full attention and every K/V write but evaluates only the frontier row. |
+| Rectangular packed prefill | Equal-length text chunks run as ONE layer-major `[B, L]` forward instead of B separate ones. |
+| Mixed-step prefill quota | `DARKBLOOM_CBV2_MIXED_PREFILL_CAP`, opt-in, default OFF. |
+
+**Kill switches** (no rebuild): `DARKBLOOM_GEMMA4_PREFILL_TAIL_ROWS=0`,
+`DARKBLOOM_GEMMA4_PREFILL_LAST_QUERY=0`, `DARKBLOOM_CBV2_MIXED_PREFILL_CAP=<n>`.
+Prompt-output narrowing and packed prefill have **no env gate** — backing those
+out means reverting the submodule bump.
+
+### Measured impact (production, not benchmark)
+
+Regressing `TTFT = intercept + slope x prompt_tokens` on gemma-4 routes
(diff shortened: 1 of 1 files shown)


Gajesh2007:
Closing — reports are kept local, not committed to the repo.

PR #580, opened 2026-07-25
Guard the three-way telemetry field-allowlist parity

Gajesh2007:
The Go, Swift and TypeScript telemetry field allowlists are mirrored in three places and **nothing enforced that they agree**. Drift is silent in the dangerous direction: add a field to Go only and the provider's client filter strips it before transmission, so the field never exists, nothing errors, and CI stays green.

Four existing guards all miss it — `telemetry_symmetry_test.go` pins event *shape*, `TelemetrySymmetryTests.swift` pins the *kind* set, `telemetry_handlers_test.go` checks seven hardcoded keys Go-side only, and the console-ui test asserts a buffer count without inspecting output.

Scrapes both client mirrors at test time and diffs against the authoritative Go map **in both directions**. The five documented pre-existing gaps (`network_reachable`, `coordinator_url`, `url`, `user_agent`, `route` — `docs/reference/telemetry-schema.md:110`) are enumerated as commented exceptions, with a staleness test so a fixed gap forces its exception to be removed.

WS-7 of the paged migration adds MTP and paged-pool fields across all three mirrors, so this guard has to exist first.

**Verified:** `go test ./api/ -run TestTelemetryAllowlist -v` → 3/3 PASS; `go vet` and `gofmt` clean. Drift detection proven by blanking the exception list — the parity test then failed naming exactly the 7 real key/mirror pairs.

```mermaid
flowchart LR
  subgraph Before
    G1["Go allowlist<br/>+ new field"] --> S1["Swift filter<br/>(not updated)"]
    S1 --> D1["field STRIPPED<br/>before transmission"]
    D1 --> C1["CI green<br/>field silently absent"]
  end
  subgraph After
    G2["Go allowlist<br/>+ new field"] --> T2{"3-way parity test"}
    S2["Swift filter"] --> T2
    W2["TS set"] --> T2
    T2 -->|"mismatch"| F2["FAIL naming the key<br/>and which mirror"]
    T2 -->|"known gap"| E2["documented exception<br/>+ staleness check"]
  end
```

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/580"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787599933&installation_model_id=21733&pr_number=580&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F580&signature=df8eb641a699ff30c51bd7da5cee0e8efbf87ed1d7a88a2f10af3885d39c5413"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/coordinator/api/telemetry_allowlist_parity_test.go
+++ b/coordinator/api/telemetry_allowlist_parity_test.go
@@ -1,0 +1,399 @@
+package api
+
+// Three-way telemetry field-allowlist parity guard.
+//
+// The allowlist is mirrored in three places:
+//
+//	Go   (authoritative, server-side filter) telemetryFieldAllowlist, this package
+//	Swift (client-side pre-filter)           provider-swift/Sources/ProviderCore/Telemetry/TelemetryEvent.swift
+//	TS   (console-ui client-side set)        console-ui/src/lib/telemetry-types.ts
+//
+// Drift is silent in the dangerous direction: add a field to Go alone and the
+// Swift client filter strips it before transmission, so the field simply never
+// exists. Nothing errors, no test fails, and the gap is invisible until someone
+// goes looking for a metric that was never emitted. The pre-existing guards all
+// miss this — coordinator/protocol/telemetry_symmetry_test.go pins event shape,
+// TelemetrySymmetryTests.swift pins the kind set, and
+// TestTelemetryFieldAllowlistHasKnownKeys only checks Go-side existence of seven
+// hardcoded keys.
+//
+// This test parses the two client mirrors out of their source files at test time
+// and asserts set equality against the Go map. The "parser" is deliberately dumb:
+// it locates the set literal by its opening line, then pulls double-quoted string
+// members out of it. It is not a Swift or TypeScript parser and must never grow
+// into one — if either literal stops being a flat list of quoted strings, rewrite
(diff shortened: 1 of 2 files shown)


Gajesh2007:

Both touched files are byte-identical in HEAD. The sole review comment — that `for i := range len(line)` is invalid Go — is **INVALID**: `go.mod` declares `go 1.25.0` and range-over-int has been legal since 1.22. Disproven by execution: `go test ./coordinator/api/ -run TestTelemetryAllowlist -v` → 3/3 PASS.

PR #582, opened 2026-07-25
Let the e2e testbed select a KV backend and concurrency

Gajesh2007:
The e2e harness could not select a KV backend. `suite.go` launched the provider with five env vars and no provider TOML; a grep for `engine_v2|kv_backend|max_concurrent|DARKBLOOM_CBV2` across `e2e/testbed` returned nothing.

Two consequences: e2e could never exercise paged, and the moment `.auto` resolves `.paged` all 14 integration tests would **silently switch backend** with no test edit and no signal.

`DARKBLOOM_CBV2_PAGED_KV` cannot be used for this — it is negative-polarity, able to force paged *off* but never *on*. Selecting paged requires `engine_v2_kv_backend` under `[backend]` in a provider TOML, so the testbed now generates one and passes `--config`. Also exposes `engine_v2_max_concurrent`, since the migration has to exercise B=8.

**Defaults are a strict no-op:** an empty `ProviderConfig{}` emits no TOML, no `--config`, and a byte-identical launch environment.

The generated TOML pins `auto_update=false` and `auto_restart=false` — a testbed provider must not self-update mid-suite nor install the launchd watchdog, which would outlive the test process.

`Stop()` removes `~/.config/darkbloom/provider.toml` only when it was absent at launch **and** its bytes exactly match what we generated. Without that, `migrateConfigIfNeeded` copies any `--config` to the canonical path when it is empty, so on a fresh CI runner one paged run would leave `engine_v2_kv_backend = "paged"` as the machine-wide default and silently flip every later no-config run.

**Verified:** `go build ./...`, `go vet`, `gofmt` clean. Behaviour checked via an overlay-only test: default → no file, no flag; `paged` → the key; `paged`+8 → both; invalid → a named error; env fallback applies and an explicit field beats it.

```mermaid
flowchart LR
  subgraph Before
    S1["testbed suite.go"] -->|"5 env vars only"| P1["provider"]
    P1 --> A1[".auto -> contiguous"]
    A1 --> T1["14 tests inherit<br/>compiled-in default"]
    T1 --> F1["flip is INVISIBLE<br/>in the diff"]
  end
  subgraph After
    S2["testbed + KVBackend<br/>+ MaxConcurrent"] --> G2["generate provider.toml"]
    G2 -->|"--config"| P2["provider"]
    P2 --> B2["engine_v2_kv_backend<br/>= paged"]
    S2 -.->|"empty = no TOML,<br/>no flag, byte-identical"| N2["existing tests unchanged"]
  end
```

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/582"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787599938&installation_model_id=21733&pr_number=582&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F582&signature=4ebbda2546cc6d3643455b55bf6419fcc04bb19c025312c24067667d3c31d9b1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Gajesh2007:

All three files byte-identical. **Carrying forward:** temp StateDir leaked on the two config-error paths (`suite.go:487-489`, `:493-494`); the overlay test the PR body claims was never committed (grep for `BuildProviderTOML|removeMigratedTestbedConfig|ResolveKVBackend` across all `*_test.go` → zero); `max_concurrent` accepted beyond the provider's [1,8] clamp.

The `UserHomeDir` vs `homeDirectoryForCurrentUser` comment is **INVALID** as a live defect — it needs the runner to override HOME, and `suite.go:511` inherits `os.Environ()` unchanged.

PR #584, opened 2026-07-25
fix(memory,registry): make the activation reserve shape-aware and trust real decode measurements

Gajesh2007:
Three coordinator/provider fixes to the memory and routing model. None of them need paged; all of them are needed *by* paged, and they independently unblock B=8 on contiguous.

## 1. The activation reserve was a constant, and the wrong one

`UnifiedMemoryCap` hardcoded a 3 GiB activation reserve regardless of prefill shape. The coordinator's mirror in `servability.go` hardcoded `servabilityActivationReserveGB = 3.0` separately — so a provider-side change could silently desynchronize the two.

The reserve is now a function of prefill shape on both sides.

```mermaid
flowchart LR
  subgraph Before
    A1[model load gate] --> B1["reserve = 3.0 GiB constant"]
    C1[coordinator cold estimate] --> D1["servabilityActivationReserveGB = 3.0 — separate literal"]
    B1 -.->|can desync silently| D1
  end
  subgraph After
    A2[model load gate] --> B2["reserve = f(prefill shape)"]
    C2[coordinator cold estimate] --> B2
  end
```

Downstream effect on a 48 GiB box with 31 GiB of weights serving 100k VLM: KV budget moves 9.200 → 5.906 GiB.

## 2. The decode load factor was fitted against a legacy proxy

`k` was fitted to a `sqrt(memory_bandwidth)` proxy rather than measured decode rates, because the Swift provider never sent a decode rate at all. It does now. Re-fitting against the engine's measured rates moves `k` 0.27 → 0.39 (MAPE 23.4% → 2.9%).

This makes the cap **tighter**, not looser — the B=8 threshold rises 39.3 → 50.1 tok/s. The plan expected re-fitting to unblock B=8; it moves the goalpost away. What actually unblocks it is the coordinator now inferring from real measurements.

## 3. A real solo measurement below the 5-sample floor was discarded

The trust floor required 5 samples before believing a solo rate, so a restart returned all 94 boxes to the no-samples state and fell straight through to the `sqrt(memory_bandwidth)` seed. A relaxed tier (≥1 sample) now sits between the trusted median and the seed. gemma-4 is dedicated, so the non-dedicated guard did not protect it.

Also updates `docs/architecture/operations/routing.md`, which documented a 3-step decode-rate fallback chain that is now 5 steps.

**Verified:** `go build ./...`, `go vet`, `gofmt` clean; registry suites green; goldens recomputed.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/584"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787606601&installation_model_id=21733&pr_number=584&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F584&signature=610c82697b0c06cac5ecf6f990cf3a83eb5fbd3de4d7b0b79570750e820829e9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/coordinator/registry/concurrency_cap.go
+++ b/coordinator/registry/concurrency_cap.go
@@ -28,6 +28,24 @@ import (
 // single-stream decode rate (resolvedDecodeTPS), NEVER the observed-under-load
 // EWMA: the observed rate collapses under the very overload this cap exists to
 // prevent, which would force the cap to 1 — a feedback loop.
+//
+// Raising a backend's own concurrency ceiling therefore buys NOTHING on its
+// own. The provider-reported number is only the `base` operand of the MIN
+// below; the resolved per-model SOLO RATE decides. Inverting the cap math, a
+// provider is granted its full reported N only above
+//
+//	q    = floor((N-1)/overcommit) + 1     # smallest quality batch whose
+//	                                       # ceil(q·overcommit) still reaches N
+//	solo >= floor · (1 + k·q)              # N=8, floor 15, overcommit 1.2:
+//	                                       #   39.3 tok/s at k=0.27
+//	                                       #   50.1 tok/s at k=0.39
+//
+// and a model whose resolved rate sits under that gets the quality batch, not
+// the bump. The rate is what needs fixing when a bump does not land — usually
+// by seeding it (modelSoloTPSSeedEnv), because solo sampling is gated on a
+// fully uncontended box (soloSampleEligible) and a model that is BUSY at the
+// new concurrency never produces another sample. See
+// TestQualityCapReachesProviderReportedConcurrency for the pinned relationship.
 
 // defaultQualityCapOvercommit is the effective overcommit when the operator has
 // not set EIGENINFERENCE_QUALITY_CONCURRENCY_OVERCOMMIT. The legacy 2.0 diluted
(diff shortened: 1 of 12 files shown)


Gajesh2007:

⚠️ **Both P1s land verbatim in #590** — `servability.go` and `UnifiedMemoryCap.swift` are byte-identical between this head and integration. Tracking them there:
- `ActivationReserveShape` has **zero production call sites** (5 mentions, all inside `UnifiedMemoryCap.swift`), so all four gates take the flat 3 GiB. Meanwhile `servability.go` *did* adopt the piecewise model — so this PR **creates** the provider/coordinator desync it claims to remove. The 9.200→5.906 GiB figure is not reproducible from this tree.
- `EIGENINFERENCE_MODEL_SOLO_TPS_SEED` exists only in `prod.env`, which is not the authoritative runtime file; it is absent from `release-env-defaults` and `required-env-keys.txt`.

Also note the description's claim that the provider 'does now' send `decode_tps` is contradicted by the tree — the field is declared and encoded but never assigned, and `routing.md:153` now documents that as deliberate.

PR #585, opened 2026-07-25
feat(provider): retire kv-quantization from the product

Gajesh2007:
Retires KV-quantization from the provider and the operator surface. **-751 net lines**, 632 of them orphaned scaffolding.

Paged pages are fp16-only, so a kv-quant slot can never be paged. Keeping the feature would mean carrying a permanently paged-ineligible configuration through the migration.

```mermaid
flowchart LR
  subgraph Before
    A1["provider.toml: kv_quant = true"] --> B1[BetaFeatures.kvQuant]
    B1 --> C1[EngineV2SlotFactory installs quantized rows]
    C1 --> D1["slot can never be paged"]
    E1["Benchmarks/KVQuant/ + scripts/kvquant"] --> F1["no target in Package.swift — pure scaffolding"]
  end
  subgraph After
    A2["provider.toml: kv_quant = true"] --> G2["RetiredCodingKeys — key accepted, ignored, WARN once"]
    G2 --> H2["every other setting still loads"]
    E2["deleted"]
  end
```

`kv_quant` goes into the existing `RetiredCodingKeys` mechanism rather than simply disappearing, so a provider upgrading with the key still in its `provider.toml` neither errors nor silently loses the rest of its configuration.

Proven end-to-end with the real CLI binary, not just unit tests: a temp `provider.toml` carrying `kv_quant = true` loads clean, keeps every other setting, and emits the retired-key warning.

One thing deliberately **kept**: the `lossy_snapshot` wire case. It is dead as a *producer* — the guard was unconditionally true once the quantized row died — but it is still decoded by the coordinator and 0.8.0 ships alongside pre-0.8.0 providers for days. Removing a wire case with a live decoder on the other side of a version boundary is a compatibility break, not cleanup. That distinction is now recorded in three places with tests pinning it.

**Verified:** `swift build` clean; CLI smoke test against a temp `provider.toml`; `Package.swift` confirmed to have no kv-quant-gated target.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/585"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787606615&installation_model_id=21733&pr_number=585&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F585&signature=ac5bd4169d25d820aca31614bdb9b635fa01343db1bfdb85d5a3a19132adf766"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Gajesh2007:

Retirement is complete in HEAD; `kv_quant` is handled via `RetiredCodingKeys`. All four Copilot build-break comments are **INVALID** — `kvQuantConfigured` returns zero hits repo-wide; those hunks were reviewed without the signature change.

The threat-model question is answered **no**: `snapshotIsLossless` had exactly one producer of `false` (the quantized row), so the removed guard was provably dead. The wire case is correctly retained.

**Carrying forward:** `darkbloom start --local` never surfaces the retired-key warning; `beta-features.md:84-87` promises no behaviour change, but a config with both `kv_quant = true` and `engine_v2_kv_backend = "paged"` flips contiguous→paged.

PR #586, opened 2026-07-25
feat(engine-v2): an explicit paged request refuses instead of degrading, and a refused sweep exits non-zero

Gajesh2007:
Stacks on #585 (kv-quant retirement, shares `docs/provider/quickstart.md`) and merges `paged-kv/bench-kv-backend` (#583, shares `BenchmarkCommand.swift`).

Two halves of one problem: **a paged rollout with no canary fleet cannot afford silent degradation.**

## An explicit paged request now refuses instead of degrading

`.auto` keeps falling back to contiguous — that is what `auto` means. But an operator who wrote `paged` and got contiguous has an unmeasured, unlabelled fleet, and every gate in this migration would report "paged" while measuring contiguous.

## A refused sweep now exits non-zero

The benchmark still exited **0** having produced zero decode cells. That is indistinguishable from success to `set -e`, to CI, and to any wrapper that checks status before parsing. With no canary fleet the benchmark *is* the safety net, and a safety net that returns 0 on total failure is not one.

```mermaid
flowchart LR
  subgraph Before
    A1["--kv-backend paged"] --> B1{paged constructible?}
    B1 -->|no| C1["silently degrade to contiguous"]
    C1 --> D1["sweep runs, exit 0"]
    D1 --> E1["CI green, dashboard says paged, fleet is contiguous"]
  end
  subgraph After
    A2["--kv-backend paged"] --> B2{paged constructible?}
    B2 -->|no| F2["refuse: named error, no slot"]
    F2 --> G2["report written WITH decodeConstructionFailure"]
    G2 --> H2["exit non-zero — CI fails loudly"]
    A3["--kv-backend auto"] --> I3["degrade to contiguous, exit 0 — unchanged"]
  end
```

Non-zero **only** when an explicit `--kv-backend` was requested and zero decode cells ran. The report is still written, because an operator needs the reason more than the status — which required two additive fields on `ThroughputSweepReport`, since the CLI could not otherwise *see* the reason: `ThroughputSweep` caught the construction failure, logged it to stderr and dropped it. An operator opening an empty report now learns why it is empty. Schema 1 → 2; successful reports byte-identical apart from that field.

Also corrects the two operator docs this falsified — the `--kv-backend` help text and the quickstart fallback paragraph both still promised that an explicit paged selection degrades.

**Verified:** `swift build` clean; 21/21 backend-policy and gate tests; `scripts/gemma_contbatch/validation.py` already failed a refused sweep via `require_positive`, so the missing piece really was only the process exit status.

---
[...]

PR #587, opened 2026-07-25
feat(telemetry): give the paged rollout an observable posture, and fix the backend key collision

Gajesh2007:
Stacks on #586 (paged refusal, shares `EngineV2SlotFactory.swift`) and merges #580 (three-way allowlist parity guard) and #584 (memory/routing, shares `servability_test.go`).

**With no canary fleet, telemetry is how we judge a paged rollout.** Nine v0.8.0 fields were allowlisted and nothing emitted them — which is worse than having no field, because the dashboard renders a clean baseline that is really just absence.

## The `backend` key collided on three semantic axes, not two

Engine identity, KV storage, and prefix-reuse provenance were all being written to one key from four sites, so a paged slot and a contiguous slot could group identically.

```mermaid
flowchart LR
  subgraph Before
    A1[EngineV2SlotFactory] --> K1["backend"]
    A2[PrefixCache wiring] --> K1
    A3[EngineV2Config] --> K1
    A4[slot posture] --> K1
    K1 --> D1["one key, three meanings — paged and contiguous group identically"]
    M1["mtp_active = status.active && engineActive"] --> M2["inert slot reports ACTIVE"]
    M2 --> M3["coordinator applies the TPS-inflation correction to a slot producing zero drafts"]
  end
  subgraph After
    B1[EngineV2SlotFactory] --> KA["kv_backend"]
    B2[PrefixCache wiring] --> KB["prefix_reuse_backend"]
    B3[EngineV2Config] --> KC["backend = engine identity"]
    B4[slot posture, 60s per slot] --> KA & KB & KC
    N1["mtp_active = drafts actually emitted"] --> N2["inert reports FALSE + mtp_inactive_reason"]
  end
```

The discriminator is a **string**, not a bool, because an absent value must stay distinguishable from an explicit empty one — otherwise every pre-0.8.0 provider books as a contiguous sample and the rollout dashboard shows a clean baseline made entirely of old providers.

## MTP has a third state and it was unnamed

Enabled, loaded, and emitting zero drafts. `mtp_active` is now **false** there. The field exists to flag the routing hazard — MTP inflates `observed_decode_tps` — and an inert slot inflates nothing, so reporting it active *inverts* the correction rather than being merely imprecise. All three production readers traced.

## Emission point

A per-bridge sampler on the existing 60s cadence, ungated so every loaded slot reports. Rejected: once-per-construction (the documented `engine_v2_kv_backend` mistake — a drop-on-full sink behind a 100/min limit is a notification, not an inventory); per-request (an idle slot vanishes, and *inert* is precisely a slot that looks busy while doing nothing); edge-triggered (a healthy slot never reports).

## Two fields removed again, deliberately

`pages_pinned` and `cow_events` were allowlisted with no producer **and no possible producer** — PagedKVPool has no pin concept and copy-on-write splitting is unimplemented. The producer site already refused to emit a hardcoded `0` because it would be indistinguishable from a measured zero; that argument applies one level up and had not been followed through. A key that survives the filter but is never written reads as a legitimate zero to anyone building a panel on it.

Paged-pool bytes therefore ship as **raw quantities** — `pool_bytes`, `pool_deferred_growth_bytes`, `pool_stranded_bytes` — not a second ratio beside `pool_utilization`. A ratio clamps at 1.0 exactly when the interesting thing is how far past capacity demand went, and raw quantities compose across slots where ratios do not.

**Verified:** `swift build` clean; `go test -run TestTelemetryAllowlist` 3/3 (the parity guard is what proves the mirrors stayed in sync); gofmt clean; 168 tests / 20 suites. The central claim is mutation-tested — reverting `active` to `status.active && engineActive` fails the inert test on exactly its two assertions and passes again on revert.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/587"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787606641&installation_model_id=21733&pr_number=587&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F587&signature=6d32182cc39f917b7f7046d562400ab90c9b62aa8b0b4a706bbf93dccfe92323"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/console-ui/__tests__/provider-backend-slot-telemetry.test.ts
+++ b/console-ui/__tests__/provider-backend-slot-telemetry.test.ts
@@ -29,3 +29,52 @@ describe("MyBackendSlot measured telemetry mirror", () => {
     expect(slot.model_load_time_ms).toBeUndefined();
   });
 });
+
+// The v0.8.0 paged-KV rollout discriminator. Go models it as `*string` with
+// `omitempty`, so a non-nil pointer to "" is still emitted as `"kv_backend":""`
+// while a pre-0.8.0 provider omits the key entirely. Those two must stay
+// distinguishable here, or the rollout dashboard cannot tell "provider says
+// nothing" from "provider says empty" — and, worse, would be free to fold both
+// into "contiguous".
+describe("MyBackendSlot kv_backend discriminator", () => {
+  const base = `"model":"gemma-4-26b-qat-4bit","state":"running","num_running":1,"num_waiting":0,"active_tokens":100,"max_tokens_potential":400`;
+
+  it("carries an explicit backend kind", () => {
+    const slot = JSON.parse(`{${base},"kv_backend":"paged"}`) as MyBackendSlot;
+    expect(slot.kv_backend).toBe("paged");
+    expect("kv_backend" in slot).toBe(true);
+  });
+
+  it("is undefined when a pre-0.8.0 provider omits it (absent ⇒ unknown)", () => {
+    const slot = JSON.parse(`{${base}}`) as MyBackendSlot;
+    expect(slot.kv_backend).toBeUndefined();
+    expect("kv_backend" in slot).toBe(false);
+    // Absent is UNKNOWN. It must never read as an observation of either kind.
+    expect(slot.kv_backend).not.toBe("contiguous");
+    expect(slot.kv_backend).not.toBe("paged");
+  });
+
+  it("keeps an explicit empty value distinguishable from omission", () => {
+    const explicit = JSON.parse(`{${base},"kv_backend":""}`) as MyBackendSlot;
+    const omitted = JSON.parse(`{${base}}`) as MyBackendSlot;
+    expect(explicit.kv_backend).toBe("");
+    expect("kv_backend" in explicit).toBe(true);
+    expect("kv_backend" in omitted).toBe(false);
+    // The reason the Go side is a pointer and this side is optional: these two
+    // slots are NOT the same observation.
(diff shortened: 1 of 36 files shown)


Gajesh2007:

Three-axis split and chip-class seed are present. Its repo-root comment is **INVALID** — there is exactly one `go.mod`, at the repo root; all three tests PASS rather than SKIP.

**Carrying forward:** the coordinator charges the 64 KiB composed-attention surcharge to *every* cold model, including fused-SDPA ones like gpt-oss-20b (head_dim 64) that the provider does not charge — early 429s; and `telemetry-schema.md:105-106` lists `pages_pinned`/`cow_events` as three-way when all three mirrors deliberately exclude them (the parity guard cannot catch this — the doc is an unguarded fourth mirror).

PR #588, opened 2026-07-25
feat(ssd): windowed sidecar blocks, so a donated window need not be replayed

Gajesh2007:
WS-4.2. **Default OFF — no shipped behaviour changes.**

gemma-4's SSD prefix-cache donation floor is 27,136 tokens against a p50 prompt of 979, so **2.3%** of its traffic can ever donate. The 25,600 term is `windowCount × maxWindow`: the replay needed because a cache hit restores the 5 full-attention layers exactly while the 25 sliding rows start **empty**. Persist the window and the replay becomes unnecessary.

```mermaid
flowchart LR
  subgraph Before
    A1[cache hit] --> B1["5 full-attention layers restored exactly"]
    A1 --> C1["25 sliding rows start EMPTY"]
    C1 --> D1["replay 25,600 tokens to refill windows"]
    D1 --> E1["donation floor 27,136 → 2.3% of traffic eligible"]
  end
  subgraph After
    A2[cache hit] --> B2["5 full-attention layers restored"]
    A2 --> F2["window sidecars tile the 25 sliding rows"]
    F2 --> G2{all blocks present?}
    G2 -->|yes| H2["no replay"]
    G2 -->|partial| I2["REFUSE the window, adopter replays — block run untouched"]
  end
```

## Per-block sidecars, and this is forced rather than stylistic

A donor's ring holds exactly W positions ending at its own `absoluteOffset`, which is **mid-block** — so one donation can only ever cover `W/blockSize - 1` = 3 of gemma's 4 window blocks. Per-block files let successive donations at different offsets **tile** the missing block and dedupe by content address across turns. A window-granular file could never be assembled from two donors.

## A partial window is refused, never restored

Restoring 3 of 4 blocks and replaying the missing 256 tokens is *not* exact — the replayed sliding rows have no window of their own, which is the original frozen-full problem at smaller scale.

## No layout-epoch bump is owed

Sidecars reuse `SSDBlockStore` verbatim — same magic, version, IV, DEK-wrap, AAD and path grammar — differing only in HMAC tag domain and three **optional** metadata fields. Because they are optional the synthesized encoder omits them, so every pre-existing block's canonical-JSON AAD stays byte-identical. Pinned by a test. `windowBase` rides in the GCM AAD, so a sidecar cannot be replayed at another boundary even if a truncated tag collided.

## Two corrections to the migration plan, measured rather than estimated

1. **27,137 is the adoption floor only.** The donate gate is a strict `>` against whole blocks, so the *donation* floor is **27,392**. After WS-4.1 restores residency, adoption drops to **1,025**.
2. **Stage economics were off by an order of magnitude.** The plan claimed +9.4 MB / +6 ms; measured at gemma-4's real per-token geometry over a 2,048-token prefix it is **+209.7 MB / +77.3 ms** — 2.71 GB/s of read+decrypt+rebuild and 7.7% of the 1,000 ms stage budget. Still comfortably inside, but 22x the bytes and 13x the time assumed.

**Verified:** `swift build` clean; 98 tests / 14 suites, including `stage → restart → rehydrate` with `max|delta| == 0` against the donor arrays after close, fresh instance and `scanOnDisk`.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/588"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787606659&installation_model_id=21733&pr_number=588&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F588&signature=0a92f562849f0c2a031e22b1bdc417c96ebfd962f370a86f5f79b463c80255c6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/provider-swift/Sources/ProviderCore/Inference/PrefixCachePolicy.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/PrefixCachePolicy.swift
@@ -129,29 +129,90 @@ enum PrefixCachePolicy {
             backend: backend)
     }
 
-    /// Conservative finite-window replay length. Kept as a policy convenience
-    /// for SSD donation/stage benefit gates; the engine plan remains
-    /// authoritative per matched boundary.
-    static func adoptionBoundTokens(layerKinds: [CBv2LayerKind]) -> Int {
-        CBv2PrefixReuseCapability.derive(
+    // MARK: - Window residency (WS-4.2)
+
+    /// Whether an adopter's sliding rows are RESTORED from windowed sidecars
+    /// or replayed. Default `.replayed` everywhere: WS-4.2 lands the format
+    /// and this plumbing, and the residency flip is a later, separate step.
+    ///
+    /// `.restoredFromSidecar` requires all three of:
+    ///   * the operator knob (`SSDPrefixCachePolicy.windowSidecarEnabled`);
+    ///   * a layout that tiles into whole-block sidecars
+    ///     (`SSDWindowSidecarGeometry.derive` — gpt-oss-20b's 128-token
+    ///     window does not, and correctly keeps its 1,536-token bound);
+    ///   * a row that can accept a restored window. Paged gets that from
+    ///     WS-4.1's `restoreWindow(keys:values:base:)`; a KILLED paged slot
+    ///     has degraded to a contiguous row, so it is resolved as contiguous.
+    static func windowResidency(
+        layerKinds: [CBv2LayerKind],
+        backendSelection: EngineV2KVBackendSelection,
+        pagedKilled: Bool = false,
+        environment: [String: String] = ProcessInfo.processInfo.environment
+    ) -> SSDWindowResidency {
+        guard SSDPrefixCachePolicy.windowSidecarEnabled(environment: environment),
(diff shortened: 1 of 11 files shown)


Gajesh2007:

**All nine review points verified fixed** in the integration branch — the cleanest of the stack. Request-aware donation now has a production overload the engine actually calls; residency claims are fail-closed behind `pagedWindowRestoreLanded`; the raw absolute token position is gone from DBK3 metadata in favour of an HMAC commitment; sidecar ordering, refresh, and the endurance pre-check all landed.

PR #594, opened 2026-07-27
feat(billing): let providers change their payout bank account themselves

Gajesh2007:
## What prompted this

A provider asked support:

> can you please send me the Stripe Express payout-bank update link for my existing Darkbloom provider connected account? I need to replace the current bank account with my business bank account.

We had no way to do that. This adds one.

## Why the existing endpoints couldn't do it

Our connected accounts are Stripe **Express** (`form.Set("type", "express")`, `coordinator/billing/stripe_connect.go`). For an account that has finished onboarding, every path we already had is a dead end:

| Path | Why it fails |
|---|---|
| `POST /v1/billing/stripe/onboard` | Mints an `account_links` link of type `account_onboarding`, which only collects **outstanding requirements**. A `ready` account has none, so Stripe bounces the user straight to `return_url`. |
| Switching that link to `account_update` | Stripe [forbids it](https://docs.stripe.com/api/account_links/create) for any account with a Stripe-hosted dashboard — which every Express account has. |
| `POST /v1/accounts/{id}/external_accounts` | Requires `controller.requirement_collection = application`; Express is `stripe`. |
| `DELETE /v1/billing/stripe/account` then re-onboard | Creates a **brand-new** connected account and discards completed KYC. |

The supported mechanism is the [Express Dashboard login link](https://docs.stripe.com/connect/integrate-express-dashboard), `POST /v1/accounts/{id}/login_links`. That dashboard is where Stripe lets the account holder change their payout bank account or debit card. We never see or store bank details, which is why we don't rebuild this ourselves.

## Before / After — behavior

```mermaid
flowchart TB
  subgraph Before["BEFORE - no way to change the bank"]
    direction TB
    B1["Provider: 'change my bank account'"] --> B2{"Payouts card, status = ready"}
    B2 --> B3["Only action: Withdraw"]
    B2 -. "no affordance" .-> B4["Unlink is hidden<br/>renders only when NOT ready"]
    B3 --> B5["Support ticket"]
    B6["If they force onboarding anyway"] --> B7["account_onboarding link"]
    B7 --> B8["No requirements due<br/>redirect to return_url"]
    B8 --> B5
  end
```

```mermaid
flowchart TB
  subgraph After["AFTER - self-serve"]
    direction TB
    A1["Provider clicks 'Change in Stripe'"] --> A2["POST /v1/billing/stripe/dashboard"]
    A2 --> A3{"Privy session?"}
    A3 -- "API key / device token" --> A4["403"]
    A3 -- yes --> A5{"Account submitted details?<br/>ready / restricted / rejected"}
    A5 -- no --> A6["409 not_onboarded<br/>'Finish your payout setup first'"]
    A5 -- yes --> A7["Stripe: POST /v1/accounts/id/login_links"]
    A7 -- "account_invalid" --> A8["Unlink + 409 stripe_account_gone<br/>500 if the unlink did not persist"]
    A7 -- "5xx / transport" --> A9["502 stripe_error<br/>account left intact"]
    A7 -- ok --> A10["200 with url"]
    A10 --> A11["New tab to Express Dashboard"]
    A11 --> A12["Provider changes their bank account"]
    A12 --> A13["account.updated webhook<br/>refreshes destination_last4"]
  end
```

## Before / After — code

```mermaid
flowchart LR
  subgraph BeforeCode["BEFORE"]
    direction TB
    C1["StripePayoutsCard<br/>ready branch"] --> C2["destination chip + Withdraw<br/>inline, complexity 30"]
    C3["useStripePayouts"] --> C4["onboard / withdraw / unlink"]
    C5["server.go routes"] --> C6["onboard, status, withdraw,<br/>withdrawals, account, webhook"]
    C7["StripeConnect"] --> C8["CreateExpressAccount<br/>CreateAccountLink<br/>GetAccount / Transfer / Payout"]
  end
```

```mermaid
flowchart LR
  subgraph AfterCode["AFTER - new nodes marked NEW"]
    direction TB
    D1["StripePayoutsCard<br/>complexity 24"] --> D2["NEW PayoutDestinationRow<br/>chip + 'Change in Stripe'"]
    D1 --> D3["Withdraw"]
    D2 --> D4["NEW useStripePayouts.openDashboard"]
    D4 --> D5["NEW createStripeDashboardLink"]
    D5 --> D6["NEW /api/payments/stripe/dashboard<br/>Next proxy"]
    D6 --> D7["NEW POST /v1/billing/stripe/dashboard<br/>requirePrivyAuth"]
    D7 --> D8["NEW handleStripeDashboardLink"]
    D8 --> D9["NEW StripeConnect.CreateLoginLink"]
    D9 --> D10["Stripe /v1/accounts/id/login_links"]
    D4 -. on error .-> D11["NEW classifyDashboardError"]
  end
```

## Security

The returned URL is a credential for a session that can **redirect where a provider's money is paid**, so:

- Wired behind `requirePrivyAuth`, not `requireAuth`. This matters more than it looks: `requireAuth` *also* populates the user context for account-bound API keys and provider device-login tokens, so the in-handler `requirePrivyUser` is **not** a second line of defense — the route-level middleware is the only gate. `TestStripeDashboardLinkRouteRejectsAPIKey` drives the real mux so flipping that one line can't pass CI.
- The account ID comes solely from the verified Privy JWT. No body, query, or path parameter — one user cannot mint a link for another's account. `validAccountID` additionally regex-gates it before interpolation into the Stripe path.
- The link is never logged, persisted, or cached. Only issuance is logged.
- The frontend disowns the pre-opened tab (`opener = null`) before it points at Stripe, and every navigation uses `replace()` so the credential never enters session history.

## Edge cases handled

- **Not onboarded** (`""` / `pending`): 409 `not_onboarded`. Stripe has no dashboard for these, and the raw error told the user to retry something that can never work. `restricted` and `rejected` accounts **do** have a dashboard — that's where Stripe explains what's wrong — so they still get through.
- **Account closed on Stripe**: unlinked locally + 409, mirroring the `refresh=1` self-heal in `handleStripeStatus`. Returns 500 instead if the unlink didn't persist, rather than telling the user we unlinked something we didn't.
- **Transient Stripe 5xx**: 502, account explicitly left intact.
- **Popup blocked, or user closes the pre-opened tab mid-flight**: falls back to the current tab instead of burning a single-use link.
- **Response without a `url`**: rejected before navigating, so no tab goes to `/undefined`.

## Modularity

Extracted `PayoutDestinationRow`, which takes `StripePayoutsCard`'s cognitive complexity to **24 — below its pre-existing baseline of 30**, despite adding a feature.

## Test plan

- [x] `go build ./...`, `go vet ./api/... ./billing/...`, `gofmt -l` clean
- [x] `go test ./api/... ./billing/...` — pass
- [x] 6 new client tests (`TestCreateLoginLink*`): path, method, malformed-ID rejection, empty-url rejection, error propagation, mock mode
- [x] 7 new handler tests (`TestStripeDashboardLink*`): auth, no account, **status gate across all five states**, **route-level API-key rejection**, happy path, account-gone unlink, transient error leaves account intact
- [x] `npx eslint src/ __tests__/` — 0 errors
- [x] `npm test` — 488 passed (was 481)
- [x] 12 new frontend tests: `classifyDashboardError` codes, and `openDashboard` navigation paths (tab reuse, popup blocked, closed tab, error cleanup, malformed body) plus card rendering
- [x] `npm run build` — `/api/payments/stripe/dashboard` registered
- [ ] Manual: click **Change in Stripe** on `/billing` and `/providers/earnings` against a real Express account, confirm the dashboard opens and a bank change round-trips via `account.updated`

## Note for the provider who asked

Their account is created with `business_type=individual`. If the business bank account is held by a legal entity rather than by them personally, the receiving bank can reject the deposit on a name mismatch — worth checking with them before they switch.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/594"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787782951&installation_model_id=21733&pr_number=594&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F594&signature=1ccbdce73b62bb7de28fa7b942b58a06000c49a157e2ef50f64811949a057c61"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Gajesh2007:
@codex — went through all four findings. Two were real and are fixed in 06f880e6; two I'm declining, with reasoning below.

## Fixed

**P2 #3 — rate-limit dashboard login-link creation.** Correct, and `POST /v1/billing/stripe/dashboard` was the only Stripe-POST billing route missing the limiter (`create-session` has it). Now `requirePrivyAuth(rateLimitFinancial(handleStripeDashboardLink))`.

One correction to the suggestion: you said *"apply the financial limiter outside the authentication middleware."* That would break it. `rateLimitWithTier` keys on `consumerKeyFromContext(r.Context())`, which only the auth middleware populates — outside `requirePrivyAuth` every caller collapses into one empty-string bucket, so the first anonymous request would consume the shared token and throttle everyone else. `rateLimitFinancial`'s own doc comment says *"Chain inside requireAuth,"* and all ten sibling financial routes do exactly that. Chained inside.

Pinned by `TestStripeDashboardLinkRouteIsRateLimited`, which drives the real mux with a signed Privy JWT (new `privySession` harness — the existing `withPrivyUser` only stuffs context and can't traverse route middleware). Negative-controlled: with the limiter removed the second call returns 409 instead of 429 and the test fails.

**P2 #1 — refresh the payout destination after returning from Stripe.** Correct, and worse than the cosmetic staleness you described. `openWithdraw` seeds `withdrawMethod` from the stale `status.instant_eligible`, and `stripe_withdraw.go:143` admits instant when *either* the freshly fetched account **or** the stored flag says eligible:

```go
if method == "instant" && !acct.InstantEligible && !user.StripeInstantEligible {
```

So after a card → bank switch in the other tab, the stale stored flag lets an instant payout through, Stripe rejects it, and the instant fee is charged then refunded before falling back to the daily sweep.

Implemented as a one-shot `visibilitychange` handler that fires `reload(true)` — `refresh=1`, not the cached read, since `account.updated` may not have landed yet. It lives in a `useEffect` rather than inside `openDashboard` so it's torn down if the page unmounts before the user ever comes back, and the flag is only armed on the new-tab branch (the popup-blocked fallback unloads this page anyway). Two tests: the refresh fires once on return, and tab focus without a dashboard link never triggers it. Negative-controlled.

## Declining

**P2 #2 — distinguish pending verification from unfinished onboarding.** The analysis is right: `stripeStatusForAccount` maps `DetailsSubmitted=true, PayoutsEnabled=false, CurrentlyDue=[]` to `pending` via the default arm, and Stripe *does* issue login links for that state. But `PayoutDestinationRow` renders only in the `ready` branch of `StripePayoutsCard` (line 114), so no UI path can reach the 409 — only a direct API caller. Adding a Stripe round-trip on every `pending` request to unblock a path nothing walks isn't worth it. If we later surface the button for `restricted`/`rejected` (which the backend gate already allows) this becomes worth revisiting.

**P2 #4 — reconcile payouts initiated from the dashboard.** Real gap, but this PR doesn't create it. Live-mode Express account holders can already sign in at `connect.stripe.com/express_login` with their email and an OTP, no platform-issued link required — per [Stripe's Express Dashboard docs](https://docs.stripe.com/connect/express-dashboard). So `reconcileUnmatchedPayout`'s `!pe.Automatic` early return already strands dashboard-initiated payouts in `transferred` today; this PR only makes the door more discoverable. Attributing non-automatic payouts to withdrawal rows means changing the webhook state machine's matching rules, which deserves its own PR and its own reasoning about over-claiming — the exact hazard the current early return exists to avoid. Filing a follow-up rather than bolting it on here.

## Unrelated: this PR cannot go green yet

`Coordinator Lint` fails with:

```
coordinator/api/toolschema.go:722:6: func typelessAssertionFamiliesAmbiguous is unused (unused)
```

That's a **master** breakage inherited via the merge commit — `master` at `fde284903` (#603) is red on the same job. Please don't "fix" it by deleting the function: its doc comment says *"Callers reject these before normalization"* and nothing calls it, so `inferredType` silently falls through to `"string"` for `{"minimum":5,"minLength":2}` and `{"not":{"type":"string"}}`. That's a live behavior gap in #603, not dead code. Needs a fix on master.

`Threat Model Review` is also failing, on infra rather than code: `anthropic.AuthenticationError: 401 - API key is invalid`.

## For the threat-model reviewer's two open questions

- **Is the URL logged or persisted?** No. `stripe_payouts.go:393-396` logs a truncated account ID and the Stripe account ID at issuance; the link itself is never logged, stored, or cached. On the JSON-vs-302 point: a redirect can't work here — the frontend opens a tab synchronously inside the click gesture to survive the popup blocker, then navigates it once the link arrives, so the URL has to come back as data.
- **Is mock mode reachable in production?** No. `BillingConfig.Validate()` (`billing/config.go:74-76`) refuses to boot when `MockMode` and a real `StripeSecretKey` are both set.

Verification for this commit: `gofmt -l` clean · `go vet ./api/... ./billing/...` clean · `go test ./api/... ./billing/...` pass · `eslint src/ __tests__/` 0 errors · `vitest` 499 passed.

PR #596, opened 2026-07-29
fix(coordinator): learn a per-pair token-budget ceiling for the affine paged pool; split capacity rejections from unservable ones

Gajesh2007:
## The insight a reviewer needs first: the paged pool's charge is affine, not linear

A paged slot advertises its token budget as `poolBytes / kvBytesPerToken`, where `kvBytesPerToken` is the **marginal** long-context rate over full-attention layers only (20,480 B/token for gemma-4). But `PagedKVPool.pageDemand` charges every **sliding-window** layer a fixed `ringPageCount` — 97 pages for a 1,024-token window — *per in-flight sequence, regardless of how long that sequence is*. For gemma-4 that is a hidden **303 MiB intercept per concurrent request**, on top of the ~20,480 B per token.

So the real pool charge for `n` concurrent sequences of `t` tokens is

```
bytes(n, t) = n · 303 MiB + n · t · 20,480
```

while the wire budget models it as

```
bytes(n, t) = n · t · 20,480
```

The intercept is worth `303 MiB / 20,480 = 15,518 tokens per in-flight request` that the coordinator never sees. At batch 8 the advertised budget over-states concurrent capacity by roughly **1.6×**: the coordinator computes that 9 requests fit a 6 GiB pool that physically holds 6, and requests 7–9 come back `capacityExhausted`.

**Everything downstream follows from that one fact.** The heartbeat is not stale — it is honestly reporting a number whose *model of the pool* is wrong. Which means:

- Any release condition phrased in terms of that number is always satisfiable, so the budget clamp is a treadmill (Fix 1).
- Any request bounced by it is a *fleet-capacity* event, not an *unservable-request* event, and the ledger says the opposite (Fix 2).
- Any cold estimate that reproduces the contiguous subtraction is describing a different quantity than a paged slot will report (Fix 3).

Production today: ~10% of requests rejected as `oversized_request`, **35% of dispatches are retries**, 439 outright provider `token_budget_exhausted` failures.

### What this PR does not do

It does **not** raise the advertised budget. The pool is physically real and `canEverFit` (`AdmissionV2.swift:354-380`) is a hard gate; over-admitting converts uptime-neutral 429s into provider rejections and, past first content, in-band SSE errors. **Every change here is strictly tighter than current behaviour and cannot over-admit** — see the invariant section at the bottom. No Swift changes, no wire changes, no new modelling constant.

---

## Before / after — behaviour

```mermaid
flowchart TB
  subgraph Before["BEFORE — dispatch, reject, retry x3, 429"]
    direction TB
    B0["request arrives<br/>advertised budget says it fits"] --> B1["scheduler SELECTS the paged pair"]
    B1 --> B2["dispatch"]
    B2 --> B3["provider: token_budget_exhausted<br/>(affine pool is physically full)"]
    B3 --> B4["clamp arms: pair headroom = 0"]
    B4 --> B5["next heartbeat restates the INFLATED budget<br/>raw max - used - queued >= 1024 -> release (a)<br/>an in-flight accept -> release (b)"]
    B5 --> B6{"capacityRetries<br/>&lt; 3 ?"}
    B6 -- yes --> B1
    B6 -- no --> B7["429 reason=oversized_request<br/>ledger: candidate_count=0<br/>could_have_served=FALSE"]
    B7 --> B8["3 wasted dispatches, 3 provider rejections,<br/>a rejection row that blames the prompt"]
  end
```

```mermaid
flowchart TB
  subgraph After["AFTER — deselect before dispatch, queue, serve"]
    direction TB
    A0["request arrives"] --> A1{"used + queued + request<br/>&lt;= effective ceiling ?"}
    A1 -- "no (learned from the reject)" --> A2["pair DESELECTED at selection time<br/>counted as a capacity rejection"]
    A2 --> A3["queue-before-shed spill<br/>(inference_admission.go)"]
    A3 --> A4["slot frees -> used+queued drops<br/>-> pair fits again -> SERVED"]
    A3 -. "queue full / 120s timeout" .-> A5["429 reason=capacity_retries_exhausted<br/>ledger: real candidate + capacity counts<br/>could_have_served=TRUE"]
    A1 -- yes --> A6["dispatch"]
    A6 --> A7["served -> accepts widen the ceiling<br/>back toward the advertised budget"]
    A6 -. "still rejected" .-> A8["ceiling ratchets tighter<br/>(learns the real intercept)"]
  end
```

## Before / after — code

```mermaid
flowchart TB
  subgraph CB["BEFORE"]
    direction TB
    CB1["RecordCapacityReject(provider, model)"] --> CB2["recordBudgetClampLocked<br/>budgetClampEntry{clampedAt, acceptedSince,<br/>budgetReported} — a BOOLEAN"]
    CB2 --> CB3["snap.budgetClamped"]
    CB3 --> CB4["liveRemainingBudget -> 0<br/>freeMemoryAdmits -> false"]
    CB5["heartbeat + accept"] --> CB6["clamp released -> raw advertised budget believed again"]
    CB7["snapshotStructuralBudget (cold)"] --> CB8["coldTokenBudgetEstimate<br/>0.9*mem - weights - 5.5 GiB"]
    CB9["shouldStopFailover: capacityRetries == 3"] --> CB10["unservableReason = oversized_request"]
    CB10 --> CB11["exhausted tail hard-writes<br/>servabilityComputed=true; candidateCount=0"]
  end
```

```mermaid
flowchart TB
  subgraph CA["AFTER"]
    direction TB
    CA1["RecordCapacityRejectSized(provider, model, requestTokens)"] --> CA2["providerBudgetCommitmentLocked (heartbeat)<br/>+ providerPendingTokensLocked (in-gap)<br/>-> (advertised, base = max of the two)"]
    CA2 --> CA3["recordBudgetClampLocked (unchanged)"]
    CA2 --> CA4["recordBudgetCeilingLocked<br/>tokens = min(advertised, base+request-1024)<br/>ratchets DOWN only, TTL anchor held"]
    CA4 --> CA5["snap.learnedTokenBudgetMax"]
    CA5 --> CA6["effectiveTokenBudgetMax(snap)<br/>= min(advertised, learned)"]
    CA6 --> CA7["liveRemainingBudget / freeMemoryAdmits<br/>(LIVE readers only)"]
    CA8["RecordCapacityAcceptOutcome"] --> CA9["noteBudgetCeilingAcceptLocked<br/>+25% per 4 accepts, delete once healed"]
    CA10["snapshotStructuralBudget (cold)"] --> CA11["coldTokenBudgetEstimate (unchanged)"]
    CA11 --> CA12{"provider observed<br/>running paged<br/>AND slot reports a KV rate?"}
    CA12 -- yes --> CA13["min(estimate, pagedColdTokenBudgetCeiling)<br/>= min(8 GiB, RAM/16) / kvBytesPerToken"]
    CA12 -- no --> CA14["estimate unchanged (fail open)"]
    CA15["shouldStopFailover: capacityRetries == 3"] --> CA16["unservableReason =<br/>capacity_retries_exhausted"]
    CA16 --> CA17["exhaustedRejectionInfo<br/>real decision counters;<br/>candidateCount floored by capacityRetries"]
  end
```

---

## Fix 1 — the clamp learns a ceiling instead of flipping a boolean

`coordinator/registry/budget_ceiling.go` (new), read by `servability.go` and `scheduler.go`.

Today one capacity-503 zeroes a pair's live headroom, and release condition (a) at `budget_clamp.go:79-86` — `raw max − used − queued ≥ 1024` — is satisfied by the very next heartbeat *because the advertised budget is structurally inflated*. Bounce, release, bounce. Forever.

The intercept is unknown to the coordinator and differs per box, per model and per batch shape, so this does not model it — it **measures** it. On a capacity reject the coordinator knows the exact commitment that failed, in the same units its own admission gate uses:

```
base = max(used + queued,          // the provider's last heartbeat
           otherPendingForModel)   // the coordinator's own in-gap accounting
S            = base + requestTokens
effectiveMax = min(advertised, S − 1024)
```

That is the weakest statement consistent with the evidence: *this pair could not hold S*. The next request at the same commitment is deselected **before dispatch**, so it takes the existing queue-before-shed spill (`inference_admission.go:457-458`) and is served when a slot frees.

The `max` is load-bearing and was **wrong in the first revision of this PR** (caught in review). Heartbeats arrive every 5 s, so a burst the coordinator itself dispatched inside one gap is invisible in `used + queued`: the heartbeat honestly says ~0 while the box is physically full. Learning off that alone latches a ceiling an order of magnitude tighter than the evidence supports — a single 20 k request against a 314 k advertised budget would derate the pair ~16×, and on a model whose traffic is uniformly long that pair then goes dark for the whole TTL with no request small enough to earn a widen. The coordinator's pending total is exactly the `coordinatorExtra` term `freeMemoryAdmits` already adds to its LHS for this reason, so learning without it was measuring in different units than the gate charges in.

Both halves of that `max` mean *everything the pair holds except the request that failed*, whose demand is then charged once on top. The second revision of this PR got that by subtracting `requestTokens` from the pending sum, and **review caught that too**: `handleInferenceError` calls `RemovePending` as its first statement, before the error is ever pushed onto `ErrorCh`, so the rejected request is never in the pending set when the classifier runs and the sum already meant "the others". The subtraction made every learned ceiling exactly one request-slot too tight on the only path that executes — bounded and self-healing, but the same class of error the `max` exists to prevent. It now excludes the request **by identity**: `providerPendingTokensLocked` skips a request id and `RecordCapacityRejectSized` threads it through, which is exact under either call ordering. A measurement should not depend on a cross-file invariant about who deletes what first.

Design points:

- **Composes with the clamp, does not replace it.** While a clamp holds the pair is FULL exactly as before; the boolean is the ceiling at its tightest. The ceiling outlives it. Every state is at least as tight as today at the same instant.
- **Recovery has three paths, and they are not equally strong.** Draining re-selects a box that finishes work with no state change at all. Widening adds +25% per 4 accepts and deletes the entry once it reaches the advertised budget — but it needs *some* request small enough to fit under the ceiling, so it is fast for a pair whose traffic straddles the ceiling (~40 short requests to climb 1 k → 90 k) and **closed** for a pair whose traffic is uniformly larger than it. For that pair the 10-minute TTL is the only escape, and it is dark until then. `TestBudgetCeilingStarvationRecoversOnlyOnTTL` states that explicitly rather than leaving it to be discovered in production, and both the latch and the heal now emit a log line so the state is visible during an incident.
- **Ratchets down only, TTL anchor included, floored at the observed commitment.** A later reject at a *higher* commitment cannot undo what a tighter one taught (only accepts widen), and it does not refresh `latchedAt` either — a non-tightening reject extending the derating window would turn a bounded fail-open into an unbounded one. But the ratchet stops at the commitment the reject *observed the pair holding*: a ceiling below that is contradicted by direct observation rather than by an accept, and holding it would strand the pair at a level nothing can widen because nothing fits under it. Re-latching there is not widening on a reject; it is refusing to keep a falsified number.
- **No no-op latches.** If the floor would put the ceiling at or above the commitment that just failed, no entry is created: it would re-admit the identical request while still burning a TTL and the accept bookkeeping.
- **Learning requires a sized reject, and a narrower vocabulary than the clamp's.** `RecordCapacityReject` (the size-less form) still arms the clamp and the strike but teaches nothing — without the size the failing commitment is unknown. Request-deterministic / lifecycle / admission-timeout rejects arm neither tracker, via the existing `armClamp` gate. On top of that, only the **token-budget / KV vocabulary** (`isTokenBudgetCapacityReject`) reaches the sized entry point: a `draining` shed for a hot-swap update, or an OOM refusing a model load, is a real capacity statement and still arms the clamp, the strike and the rate window exactly as today — but its commitment number teaches the ceiling nothing, and because fault identity is stable across the reconnect a ceiling latched off a drain would follow a box that comes back healthy. The clamp is a boolean one heartbeat releases; the ceiling is a TTL-bounded derate that survives reconnects. Identical evidence, very different blast radius, so the ceiling gets the narrower gate.
- **The derate is observable as a level, not just as events.** The latch and the heal log, but a log line cannot answer the only question that matters mid-incident: *how many pairs are deselected right now, and is that number decaying?* `routing.budget_ceilings_latched` (registry `BudgetCeilingsLatched`, polled by the existing api metrics loop) is that gauge. A working mitigation shows a handful of pairs that decay; a starved fleet shows the count climbing and staying. This incident was misdiagnosed for an hour because the telemetry conflated two populations, so shipping the remedy without a gauge on its own blast radius would repeat the mistake in a new place.
- **Live readers only.** The structural ceiling (`snapshotStructuralBudget` → `PredictServable`) stays raw, exactly as the clamp already does: a transient per-box measurement must never produce a terminal `prompt_too_long` 429.
- Kill switch `EIGENINFERENCE_BUDGET_CEILING=false`, TTL `EIGENINFERENCE_BUDGET_CEILING_TTL_SECONDS`. Keyed by stable fault identity, bounded by the same >1024 sweep as its siblings. On identity rebind the **tighter live** entry wins: expiry is checked on both sides, because comparing token values blind lets an expired-but-tighter entry destroy a live one and the survivor then carries the stale anchor, so it gates nothing.

## Fix 2 — retry exhaustion is a fleet statement, not a request statement

`coordinator/api/dispatch.go`.

`shouldStopFailover` ran out of `maxCapacityClassRetries` (3) and latched `oversized_request` — the code reserved for genuinely-unservable shapes. Worse, the exhausted tail hard-wrote `servabilityComputed = true; candidateCount = 0`, and `rejection_telemetry.go:141-155` derives `CouldHaveServed = candidateCount > 0` from it. So every one of these rejections reported *zero candidates, could not have served*, when in fact three providers had been dispatched to and were merely full.

**This misreporting sent our incident diagnosis down the wrong path for an hour**, and is worth fixing on its own merits.

- New reason code `capacity_retries_exhausted`, distinct from `oversized_request`.
- New metric `routing.capacity_retries_exhausted` so the two are separately alertable; `routing.oversized_request_rejected` keeps its original meaning.
- New `exhaustedRejectionInfo` helper (also thins the exhausted tail) reports the real routing-decision counters, with `candidateCount` floored by `d.capacityRetries` — each of those rejections came from a provider that was by definition a candidate when it was selected, so a final scan that happened to carry no counters cannot under-report below what the loop itself observed.
- `capacityRejections` is deliberately **not** floored the same way. It counts candidates the admission gate shed *during selection*, never dispatched; `capacityRetries` counts providers that *were* dispatched and 503'd. Those are disjoint halves of the funnel, and maxing them into one field would rebuild exactly the counter conflation this fix exists to remove. (The first revision of this PR did that; caught in review.)
- `d.lastDecision` latches only decisions that carry signal, so a signal-free tail attempt cannot erase what earlier attempts saw.

## Fix 3 — the cold estimate is describing the wrong quantity on a paged box

`coordinator/registry/servability.go`, `coordinator/registry/kv_backend.go`.

`coldTokenBudgetEstimate`'s own doc at `:148-157` argues the cold estimate converges to the warm report "because both are the same subtraction". True on contiguous. False on paged: `PagedKVPhysicalCapacityPolicy` plans a *separate physical pool* bounded by `min(absoluteHardCapBytes, physicalMemoryBytes / physicalMemoryDivisor)` = `min(8 GiB, RAM/16)`. On a 96 GiB box the contiguous subtraction says ~3.4 M tokens where the paged slot will report ~314 k — roughly 10×. That doc is corrected in this PR rather than left to mislead the next reader.

`coldTokenBudgetEstimate` is left byte-identical (it still faithfully mirrors the contiguous reserve arithmetic, which is all it ever claimed to do). The paged case is a **second ceiling** applied on top in `snapshotStructuralBudget`, taking the tighter of the two.

Two gates, both fail-open:

1. **The machine must have been observed running paged.** `kv_backend.go` gains its first routing reader, `runsPagedKVLocked`. The file's `MEASUREMENT ONLY` banner is replaced with a deliberate, scoped justification: provider granularity (the cold model has no slot here, so the question is about the box's pool arithmetic), any paged slot makes the box paged (the backend is a box-level capability, and paged is the conservative answer). A false answer covers both "contiguous" and "never observed", which is correct for the sole reader — it clamps on true and leaves the estimate alone on false.
2. **The slot must report a real per-token KV rate.** There is deliberately **no** `kvCacheBytesPerToken` fallback: that constant is 400 kB/token measured on a *contiguous* 7B cache, ~20× a composed-attention model's marginal paged rate. Dividing a real paged byte bound by that placeholder would manufacture a token ceiling an order of magnitude below the truth and shed traffic every provider could serve.

A swept test asserts the paged estimate is never looser than the contiguous one across box sizes and model weights.

**Reach, stated honestly** (review corrected an overstatement here). Gate 2 is both the correctness condition *and* a severe narrowing, because providers only report a slot for a model they are holding. The 1 h idle unload leaves `slots: []` (`KVBackendPosture.swift`), so the ordinary cold box arrives with no KV rate and is never clamped — the first revision of this PR claimed otherwise, and its end-to-end fixture used a `state: "idle_shutdown"` that no current provider emits. What actually remains is the slot that is present but non-resident to `snapshotStructuralBudget`: state `"reloading"` (wedge-recovery rebuild) or `"crashed"` (wedge suspected), reporting a live `kvBytesPerToken` while `reportedKVBytesCapacity` has collapsed to zero — which is one of the same wedge triggers, so the states co-occur (`EngineV2Bridge+Capacity.swift:120`: `budgetMax = reportedKVBytesCapacity / kvBytesPerToken`). That is real, and the tests now use it, but it is rare.

Narrower still, on review: on that exact shape `knownZeroTokenBudget` (`scheduler.go:1380`) already refuses the box inside `freeMemoryAdmits`, so live admission never consults a cold estimate for it. The only reader this clamp can move is `snapshotStructuralBudget` → `PredictServable` — its entire observable effect is putting the fleet's terminal-429-versus-queue boundary on the truth. **Fix 3 is consistency hardening for the paged fleet, not a material share of the incident.** Fix 1 is where the incident lives.

---

## Why every change is strictly tighter, and cannot over-admit

| Change | Mechanism | Bound |
|---|---|---|
| Learned ceiling | Applied as `min(advertised, learned)` at every read (`effectiveTokenBudgetMax`) | Admission never believes a budget larger than the heartbeat's |
| Learned ceiling vs. the clamp | Clamp behaviour is untouched; the ceiling only applies *after* the clamp releases | No instant is looser than pre-change |
| Learned ceiling latch | `learned >= advertised` does not latch; rejects ratchet down only | A reject can never widen anything |
| Structural shed | `snapshotStructuralBudget` reads the raw advertised budget | No new terminal `prompt_too_long` |
| Reason-code split | Pure relabel + telemetry; identical stop behaviour and identical HTTP 429 | No routing change |
| Paged cold clamp | `min(existing estimate, pool ceiling)` | Can only lower an estimate, never raise one |
| Paged cold clamp gates | Requires an observed paged backend **and** a real KV rate | Ambiguous input leaves today's estimate byte-identical |

"Strictly tighter" is the safety property, but on a capacity-bound fleet it is also the risk: a correlated latch across many paged boxes for the same model darkens them simultaneously and pushes traffic onto the queue-before-shed path. That is why the ceiling is bounded three ways (drain, widen, TTL), why the kill switch exists, and why latch/heal are logged.

## Tests

New: `coordinator/registry/budget_ceiling_test.go`, `coordinator/registry/servability_paged_cold_test.go`, `coordinator/api/dispatch_capacity_retries_test.go`.

- **The treadmill, reproduced** (`TestBudgetCeilingBreaksTheClampTreadmill`): a paged-shaped pair advertising 314,572 tokens at 120,000 committed accepts a 20,256-token request, is rejected by the provider, releases the clamp on the next heartbeat restating the *same inflated budget*, and must **not** be re-selected at the same commitment — then must be selectable again once sustained accepts widen the ceiling. `TestBudgetCeilingKillSwitchRestoresTreadmill` pins the pre-change behaviour behind the kill switch so the two cannot silently converge.
- **The in-gap burst** (`TestBudgetCeilingLearnsFromInGapPendingNotJustTheHeartbeat`): a heartbeat reporting an empty box, six ~20 k requests placed by the coordinator inside the heartbeat gap, and a seventh rejected — the ceiling must be learned off the whole ~141 k commitment, not the ~19 k a stale heartbeat implies. The fixture models the production shape (the rejected request already removed from pending), and a second arm runs the same reject with it *still* resident to pin that the exclusion is by id: both must learn the identical ceiling.
- **The falsified ceiling** (`a ceiling the live commitment falsifies is not ratcheted`): a floor-level ceiling latched by a tiny reject, then a reject observing the pair holding 80 k — the ratchet must yield to the observation rather than strand the pair at 1,024 forever.
- **Narrowed learning scope** (`TestOnlyTokenBudgetRejectsTeachTheCeiling`): a table over the real classifier showing which capacity vocabularies teach the ceiling and which only arm the clamp, with a cross-check that no shape claimed by an earlier branch also reads as a token-budget reject.
- **The starvation path** (`TestBudgetCeilingStarvationRecoversOnlyOnTTL`): a ceiling below one request size, on a model whose traffic is uniformly that size, with the one-shot clamp explicitly released so the ceiling is unambiguously what gates. The pair is dark while idle, widening is unreachable, and only the TTL recovers it. The design's worst case, pinned.
- Latch boundaries: no-latch when not tighter than advertised, floor, no-op floor latch rejected, ratchet-down **and its TTL anchor**, unlearnable inputs, TTL fail-open, sized-reject requirement, non-provider rejects arm nothing, healed-entry deletion, `min(advertised, learned)` never widening, structural shed untouched, `freeMemoryAdmits` honouring the ceiling.
- Identity-rebind merge as a six-case table: tighter-live-source, tighter-live-destination, **expired source must not displace a live destination**, expired source alone dropped, live source onto an empty key, live source replacing an expired destination — plus an assertion that whatever survives can still actually gate.
- Fix 2: transient-vs-deterministic latching; `exhaustedRejectionInfo` counters on all four exits, including that `capacityRejections` keeps the decision's own value; and the headline consequence through the real ledger write — `could_have_served = true` for capacity exhaustion, `false` for an unservable shape.
- Fix 3: mixed-fleet through the real registry (paged and contiguous 96 GiB boxes, same model, same hardware, differing only in the backend their heartbeats name, using the `reloading`-slot shape a provider actually emits); a swept never-looser invariant; three fail-open cases; both branches of `min(8 GiB, RAM/16)` and the 128 GiB crossover; the `runsPagedKVLocked` table.

## Verification

```
cd coordinator && gofmt -l . && go vet ./... && go test ./...   # clean, all green
cd e2e && go build ./... && go vet ./...                        # clean
```

No Swift changes in this PR.

## Honest scope: this is mitigation, not recovered capacity

None of this creates a single token of KV. The pool is physically what it is; these three changes stop the coordinator from *lying to itself about it* and stop us from *burning dispatches discovering it*. My estimate of the ~10% rejection rate each removes is in the PR discussion — the short version is that Fix 1 does essentially all of the work by converting bounced dispatches into queue waits, Fix 2 removes none of it (it makes the remainder legible), and Fix 3 removes approximately none of it (see its reach note above).

**The real capacity is only recoverable provider-side**: either size the pool for the affine charge, or shrink the 1.5156× ring over-provision so the intercept stops costing 303 MiB per in-flight sequence. That is a separate change to `PagedKVPool` / `PagedKVPhysicalCapacityPolicy`.

**Do not merge** without the provider-side follow-up planned.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/596"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787957106&installation_model_id=21733&pr_number=596&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F596&signature=597113abe055399d6240e8d82b88164a6236677b78629d8ad784bdd11a734fbc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -403,13 +403,19 @@ func (s *Server) noteInferenceError(providerID string, pr *registry.PendingReque
 			tripped = s.registry.RecordCapacityRejectLifecycle(providerID, pr.Model)
 		case s.isRequestShapeBatchBudgetReject(providerID, pr.Model, errStr, errReason):
 			tripped = s.registry.RecordCapacityRejectRequestShape(providerID, pr.Model)
-		default:
-			// Sized: the request's token demand plus the pair's reported
-			// used+queued is the exact commitment the provider refused, which
-			// is what the learned budget ceiling latches (registry/
+		case isTokenBudgetCapacityReject(errStr):
+			// Sized: the request's token demand on top of the commitment the
+			// pair was already holding is the exact level the provider refused,
+			// which is what the learned budget ceiling latches (registry/
 			// budget_ceiling.go) so the pair stops being re-selected at that
-			// level once the one-shot clamp releases.
+			// level once the one-shot clamp releases. Only the token-budget /
+			// KV vocabulary qualifies — a drain or an OOM is a real capacity
+			// shed whose commitment number would teach the ceiling nothing.
 			tripped = s.registry.RecordCapacityRejectSized(providerID, pr.Model, pr.RequestTokens())
+		default:
+			// Every other provider-indicting capacity shed: unchanged
+			// behaviour — clamp plus strike plus rate window, no ceiling.
+			tripped = s.registry.RecordCapacityReject(providerID, pr.Model)
 		}
 		if tripped {
 			s.ddIncr(metricCapacityCooldownTripped, []string{"provider_id:" + providerID, "model:" + pr.Model})
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -411,7 +411,8 @@ func (s *Server) noteInferenceError(providerID string, pr *registry.PendingReque
 			// level once the one-shot clamp releases. Only the token-budget /
 			// KV vocabulary qualifies — a drain or an OOM is a real capacity
 			// shed whose commitment number would teach the ceiling nothing.
-			tripped = s.registry.RecordCapacityRejectSized(providerID, pr.Model, pr.RequestTokens())
+			tripped = s.registry.RecordCapacityRejectSized(
+				providerID, pr.Model, pr.RequestID, pr.RequestTokens())
 		default:
 			// Every other provider-indicting capacity shed: unchanged
 			// behaviour — clamp plus strike plus rate window, no ceiling.
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -404,7 +404,12 @@ func (s *Server) noteInferenceError(providerID string, pr *registry.PendingReque
 		case s.isRequestShapeBatchBudgetReject(providerID, pr.Model, errStr, errReason):
 			tripped = s.registry.RecordCapacityRejectRequestShape(providerID, pr.Model)
 		default:
-			tripped = s.registry.RecordCapacityReject(providerID, pr.Model)
+			// Sized: the request's token demand plus the pair's reported
+			// used+queued is the exact commitment the provider refused, which
+			// is what the learned budget ceiling latches (registry/
+			// budget_ceiling.go) so the pair stops being re-selected at that
+			// level once the one-shot clamp releases.
+			tripped = s.registry.RecordCapacityRejectSized(providerID, pr.Model, pr.RequestTokens())
 		}
 		if tripped {
 			s.ddIncr(metricCapacityCooldownTripped, []string{"provider_id:" + providerID, "model:" + pr.Model})
--- a/coordinator/api/dispatch.go
+++ b/coordinator/api/dispatch.go
@@ -165,6 +165,14 @@ type dispatchState struct {
 	// uptime-neutral 429 with unservableReason instead of retrying/5xx'ing.
 	unservable       bool
 	unservableReason string
+	// lastDecision is the most recent scheduler decision that carried signal
+	// (a selection, or a scan that saw candidates / capacity rejections). The
+	// exhausted ladder reports its counters in the rejection ledger, so a
+	// "could_have_served" verdict is computed from what the scheduler actually
+	// saw instead of being hard-written to zero. Attempts whose scan found
+	// nothing at all do not overwrite it: an empty tail decision would erase
+	// the evidence of the providers earlier attempts really did try.
+	lastDecision registry.RoutingDecision
 	// terminalClientError is set when a dispatched provider returned a DETERMINISTIC
 	// client-shape 4xx (400/413/422/415 — invalid tool payload / role / response_format
 	// / unsupported media). That rejection is identical on every provider (the bad
@@ -781,6 +789,18 @@ func (d *dispatchState) rejectionInfo(stage, reason string, status, retryAfterMs
 	return info
 }
 
+// noteDecision latches a scheduler decision for the exhausted ladder's
+// rejection telemetry. Only decisions that carry signal are kept — a later
+// attempt whose scan found nothing (every provider already excluded, say) must
+// not erase what earlier attempts saw.
+func (d *dispatchState) noteDecision(decision registry.RoutingDecision) {
+	if decision.ProviderID != "" || decision.CandidateCount > 0 ||
+		decision.CapacityRejections > 0 || decision.ModelTooLargeRejections > 0 ||
+		decision.VisionRejections > 0 {
+		d.lastDecision = decision
+	}
+}
+
 func (d *dispatchState) rejectionInfoWithDecision(stage, reason string, status, retryAfterMs int, decision registry.RoutingDecision) rejectionInfo {
 	info := d.rejectionInfo(stage, reason, status, retryAfterMs)
 	info.servabilityComputed = true
@@ -789,6 +809,45 @@ func (d *dispatchState) rejectionInfoWithDecision(stage, reason string, status, 
 	info.modelTooLargeRejections = decision.ModelTooLargeRejections
 	info.visionRejections = decision.VisionRejections
 	info.bestTTFTMs = decision.BestTTFTMs
+	return info
+}
+
+// exhaustedRejectionInfo builds the rejection-ledger record for the exhausted
+// dispatch ladder's 429/503 exit. The counterfactual it carries is what
+// recordRejection turns into could_have_served, so the three exits must not
+// share one shape:
+//
+//   - CAPACITY RETRIES EXHAUSTED: the fleet could have served this request and
+//     was merely full. Report the scheduler's real counters, floored by the
+//     number of providers that actually capacity-rejected a dispatched attempt
+//     (each was by definition a candidate), so a decision that happened to
+//     carry no counters still cannot under-report below what the loop itself
+//     observed. This path previously hard-wrote candidateCount = 0, which
+//     booked every busy-fleet rejection as "could not have served" — the
+//     misreporting that sent the 2026-07 incident diagnosis down the wrong
+//     path for an hour.
+//   - DETERMINISTICALLY UNSERVABLE: candidates existed and every one of them
+//     would reject the same request identically, so zero is the honest answer
+//     and it is stated authoritatively (mirroring the preflight gate).
+//   - EVERYTHING ELSE: leave servabilityComputed false and let recordRejection
+//     compute the counterfactual itself off the request path.
+func (d *dispatchState) exhaustedRejectionInfo(reason string, statusCode, retryAfterMs int) rejectionInfo {
+	if !d.unservable {
+		return d.rejectionInfo("dispatch", reason, statusCode, retryAfterMs)
+	}
+	if reason != rejectionReasonCapacityRetriesExhausted {
+		info := d.rejectionInfo("dispatch", reason, statusCode, retryAfterMs)
+		info.servabilityComputed = true
+		info.candidateCount = 0
+		return info
+	}
+	info := d.rejectionInfoWithDecision("dispatch", reason, statusCode, retryAfterMs, d.lastDecision)
+	if info.candidateCount < d.capacityRetries {
+		info.candidateCount = d.capacityRetries
+	}
+	if info.capacityRejections < d.capacityRetries {
+		info.capacityRejections = d.capacityRetries
+	}
 	return info
 }
 
@@ -925,6 +984,7 @@ func (d *dispatchState) dispatchPrimary() dispatchOutcome {
 	)
 	d.dispatchErr = dispatchErr
 	d.dispatchErrCode = dispatchErrCode
+	d.noteDecision(decision)
 	if !routeRecorded {
 		d.recordRoutingDecision(decision, dispatchErr, "")
 	}
@@ -1340,12 +1400,31 @@ func (d *dispatchState) noteProviderError(provider *registry.Provider, pr *regis
 	return d.s.noteDispatchProviderError(provider, pr, statusCode, errStr, errReason, terminalCause, held)
 }
 
-// rejectionReasonOversized is the rejection-ledger reason_code for a request the
-// dispatch loop stopped because no provider can serve it (deterministic context
-// overflow, or a transient-capacity shortage that exhausted
-// maxCapacityClassRetries). Distinct from the preflight "context_exceeded" /
-// "prompt_too_long" and the legacy dispatch-exhausted "unservable_token_budget".
+// rejectionReasonOversized is the rejection-ledger reason_code for a request
+// the dispatch loop stopped because it is DETERMINISTICALLY unservable — a
+// context overflow that every provider in the fleet would reject identically.
+// Distinct from the preflight "context_exceeded" / "prompt_too_long" and the
+// legacy dispatch-exhausted "unservable_token_budget".
+//
+// Reserved for genuinely-unservable SHAPES. It used to double as the code for
+// a transient-capacity shortage that exhausted maxCapacityClassRetries, which
+// is a statement about the fleet being busy, not about the request — see
+// rejectionReasonCapacityRetriesExhausted.
 const rejectionReasonOversized = "oversized_request"
+
+// rejectionReasonCapacityRetriesExhausted is the rejection-ledger reason_code
+// for a request the dispatch loop stopped because it burned
+// maxCapacityClassRetries (3) TRANSIENT-capacity failovers. The three
+// providers it tried were candidates that were merely full; a fourth might
+// have served it, and the same request an hour later on an idle fleet
+// certainly would.
+//
+// Split out of rejectionReasonOversized deliberately. Conflating the two makes
+// a busy fleet indistinguishable from an unservable request in the rejection
+// ledger, which is precisely how the 2026-07 paged-rollout incident was
+// misdiagnosed: ~10% of traffic booked as "oversized_request" when the prompts
+// were ordinary and the boxes were simply out of KV pool.
+const rejectionReasonCapacityRetriesExhausted = "capacity_retries_exhausted"
 
 // rejectionReasonTemplateRenderFailed is the rejection-ledger reason_code for
 // a request the dispatch loop stopped because the model's chat template
@@ -1428,8 +1507,14 @@ func (d *dispatchState) shouldStopFailover() bool {
 		d.s.ddIncr("routing.dispatch_to_capacity_503", []string{"model:" + d.model, "reason:transient"})
 		d.capacityRetries++
 		if d.capacityRetries >= maxCapacityClassRetries {
+			// Out of retries, NOT out of servability: every one of these
+			// rejections came from a provider that was a real candidate and
+			// simply had no KV room right now. The ladder still stops (a
+			// fleet-wide transient must not storm 64 providers) and still
+			// emits the uptime-neutral 429, but under its own reason code so
+			// the ledger can tell a busy fleet from an unservable request.
 			d.unservable = true
-			d.unservableReason = rejectionReasonOversized
+			d.unservableReason = rejectionReasonCapacityRetriesExhausted
 			return true
 		}
 		return false
@@ -2656,16 +2741,26 @@ exhausted:
 			}
 			s.ddIncr("routing.client_error_passthrough", []string{"model:" + d.model, "code:" + strconv.Itoa(statusCode)})
 		} else if d.unservable {
-			// The loop stopped early because no provider can serve this request
-			// (deterministic context overflow, or a capacity transient that
-			// exhausted maxCapacityClassRetries). We already know the verdict, so
-			// skip the quick-capacity probe and the 5xx→429 reclassification below:
-			// emit a single uptime-neutral 429. This is the proactive complement to
-			// the always-on backstop — it converts the request BEFORE storming the
-			// fleet, not after 64 attempts.
+			// The loop stopped early: either no provider can serve this request
+			// (deterministic context overflow) or a capacity transient exhausted
+			// maxCapacityClassRetries. We already know the verdict, so skip the
+			// quick-capacity probe and the 5xx→429 reclassification below: emit a
+			// single uptime-neutral 429. This is the proactive complement to the
+			// always-on backstop — it converts the request BEFORE storming the
+			// fleet, not after 64 attempts. The two verdicts carry DIFFERENT
+			// reason codes and different metrics (an unservable shape is a
+			// property of the request; exhausted capacity retries are a property
+			// of the fleet), so the ledger and the dashboards can tell them apart.
 			statusCode = http.StatusTooManyRequests
-			reason = rejectionReasonOversized
-			s.ddIncr("routing.oversized_request_rejected", []string{"model:" + d.model, "stage:dispatch"})
+			reason = d.unservableReason
+			if reason == "" {
+				reason = rejectionReasonOversized
+			}
+			if reason == rejectionReasonCapacityRetriesExhausted {
+				s.ddIncr("routing.capacity_retries_exhausted", []string{"model:" + d.model, "stage:dispatch"})
+			} else {
+				s.ddIncr("routing.oversized_request_rejected", []string{"model:" + d.model, "stage:dispatch"})
+			}
 		} else if statusCode == 0 {
 			// Distinguish capacity exhaustion (429) from genuine unavailability (503).
 			// A quick capacity check tells us if providers exist but are full.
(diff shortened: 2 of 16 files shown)


Gajesh2007:
## Review round 2 — three blockers fixed, one claim corrected

An independent review of `64ae7b7a7` returned FAIL. All three blockers are addressed in `7b885e31d`; the PR description is updated where it overstated what a fix does.

**1. The ceiling was learning from a stale heartbeat.** This was the real bug. `providerBudgetCommitmentLocked` reads `used + queued` from the last heartbeat, cadence 5 s — so a burst the coordinator *itself* dispatched inside that gap is invisible. A reject mid-burst latched `requestTokens − 1024` against a 314 k advertised budget: a ~16× derate learned from a measurement we knew was stale, and on a long-prompt-only model that pair then goes dark for the full TTL with no request small enough to earn a widen. The commitment is now `max(heartbeat committed, pendingForModel − requestTokens) + requestTokens`, which is the same quantity `freeMemoryAdmits` already charges as `coordinatorExtra`. A design that says "measure, don't model" cannot measure the wrong side of a 5-second window. `TestBudgetCeilingLearnsFromInGapPendingNotJustTheHeartbeat` fails without the fix.

**2. The rebind merge ignored expiry.** `migrateFaultStateLocked` compared ceiling token values blind, so an expired-but-tighter source could displace a live destination — and the survivor carried the source's stale anchor, so it read as expired and gated nothing, silently discarding the live ceiling. Expiry is now checked on both sides. The sibling maps never had this hole because they order by timestamp; the ceiling orders by tokens, so expiry has to be explicit. Six-case table test, including the case that was broken.

**3. Fix 3's reach was overstated, and its test fixture was fiction.** The claim that a box which idle-unloaded a model still reports its KV rate is false: `KVBackendPosture` leaves `slots: []`, and `idle_shutdown` appears nowhere in the Swift sources. The reachable window is the slot that is present but non-resident — state `reloading` (wedge-recovery) or `crashed` — reporting a live `kvBytesPerToken` while its byte capacity has collapsed to zero, which is one of the same wedge triggers, so the states genuinely co-occur. The code is right and stays; the fixtures now use the shape a provider actually emits, and the PR says plainly that **Fix 3 contributes approximately nothing to the incident**. Fix 1 is where it lives.

Also fixed from the same review:

- Non-tightening rejects no longer refresh `latchedAt`. The doc comment already forbade it; the code did it anyway. Bounded today only because a dark pair receives no dispatches — one caller away from an unbounded derating window.
- No ceiling entry is created when the floor would put it at or above the commitment that failed. That entry re-admits the identical request while burning a TTL and the accept bookkeeping.
- `capacityRejections` is no longer floored by `capacityRetries`. Selection-stage shedding and post-dispatch 503s are disjoint funnel halves; merging them rebuilt the counter conflation this PR exists to remove, one field over. `candidateCount` keeps its floor — a dispatched provider *was* a candidate, so the floor cannot over-report.
- Latch and heal now log. Previously you could not answer "is it latching, how tight, how many pairs are dark" from anything but routing outcomes.
- The starvation path is now a test, not a footnote. A ceiling below one request size on a uniformly-long-prompt model is dark until the TTL, full stop — `TestBudgetCeilingStarvationRecoversOnlyOnTTL` says so with the one-shot clamp explicitly released so it is unambiguously the ceiling under test.
- `runsPagedKVLocked` returns one bool; both callers discarded the tri-state, and a test was pinning a return value nothing consumed.
- Dropped a tautological test that only asserted two `const` strings differ.

**Consciously not changed:**

- **The 1024 floor stays.** It is the provider's own minimum serving budget (`BatchScheduler+Telemetry` floors `tokenBudgetMax` there), and a pair whose honest ceiling is below that cannot serve anything. The review's objection — that 1024 admits only ~768-token prompts — is true but is a statement about the pair, not about the constant. The no-op case it enabled is fixed above.
- **Logs, not DogStatsD.** `registry` has no metric sink; every sibling breaker (clamp, cooldown, rate window) is equally unmetered. Wiring one in is a worthwhile change and a separate one — doing it here would mean either threading an emitter through `Registry` or duplicating the classification in `api`.

`gofmt -l . && go vet ./... && go test ./...` clean in `coordinator`; `go build ./... && go vet ./...` clean in `e2e`.

Gajesh2007:
## Review round 3 — the second independent review found four things the first missed

A second reviewer ran against `04e8e86b8` and returned FAIL. Its three blockers were the same three already fixed in `7b885e31d` (stale-heartbeat commitment, ratchet TTL refresh, migration liveness) — but its version of the ratchet defect is **sharper than the fix I shipped**, and it raised four items the first review did not. All addressed in `69df04ad8`.

### The ratchet could hold a value the pair had already disproved

I fixed the TTL half; this reviewer found the token half. Concretely:

1. A small reject latches the ceiling at the 1024 floor.
2. A later reject observes the pair **holding 80 k** and refusing at 80.1 k.
3. The ratchet's rule was "a reject can never widen" — so it pulled 80,000 back down to 1,024.

The pair then sits at a 1,024-token ceiling while its own accounting says it is holding 80 k. Nothing fits under 1,024, so no accept can land, so widening is unreachable. The TTL anchor fix stops that from being *indefinite*, but a full 10 minutes of a provider being dark on evidence it has itself contradicted is still wrong.

The premise was too broad. "Only accepts may widen" is right for the *learned* number, but `alreadyCommitted` is not an inference — it is the commitment the pair is observed holding, from the same reject. A ceiling below it is falsified by direct observation, not by an accept. The ratchet floor is now that observed commitment, and re-latching there is not widening on a reject; it is refusing to keep a number that has been disproved. Still strictly below `advertised`, so the tighter-than-today property is untouched.

### The ceiling was learning from rejects that say nothing about token budgets

The `default:` arm sent *every* provider-indicting capacity shed to the sized entry point — including `draining` (hot-swap update), `out of memory`, `insufficient memory`, and generic overload strings. The reviewer's framing is the right one: identical evidence to the clamp, but ~600× the consequence. The clamp is a boolean one fresher heartbeat releases; the ceiling is a TTL-bounded derate of live headroom that **survives reconnects on stable fault identity**.

The drain case is the sharpest: a provider draining for an update comes back healthy, and the ceiling follows it across the reconnect. Learning is now gated on the token-budget/KV vocabulary specifically (`isTokenBudgetCapacityReject`); everything else keeps today's behaviour byte-for-byte — clamp, strike, rate window, no ceiling. Strictly less learning than the previous revision, so strictly closer to current production.

### Two smaller ones, both real

- **I broke a read atomicity invariant in the refactor.** `providerBudgetSnapshotLocked` was taking `p.mu` twice — once for `LastHeartbeat`, again inside `providerBudgetCommitmentLocked`. `Heartbeat` stamps both in one critical section, and the clamp's release proof is exactly "a heartbeat strictly newer than the clamp showed headroom", so pairing one heartbeat's freshness with another's budget breaks the thing the reader depends on. Back to one section, with `providerBudgetCommitmentLocked` as the thin wrapper.
- **`noteBudgetCeilingAcceptLocked` grew the token value forever with no advertised budget.** Widening needs something to widen *toward*; with no budget report the entry cannot gate either way, and the grown number would be meaningless against the next real heartbeat. It now holds the value and the accumulated progress.

Plus `exhaustedRejectionInfo` branches on `d.unservableReason` instead of the `reason` parameter — they agree at the only call site today, but a future caller passing them independently would silently revert to `candidateCount = 0`, which is the misreporting this split exists to remove.

### On its observability finding

Both reviewers landed on this independently, which is a signal. `7b885e31d` added latch/heal log lines. I still think DogStatsD belongs in a separate change — `registry` has no metric sink and every sibling breaker is equally unmetered — but the reviewer's point stands that the clamp is a boolean lasting one heartbeat while this is a 10-minute derate, so the precedent is weaker than I claimed.

`gofmt`, `go vet`, full `go test ./...`, and `-race` on `registry` + `api` all clean; `e2e` builds and vets.

PR #654, opened 2026-08-21
Auto-size activation reserve per-model using measured B=8 peaks

### LINKED ISSUE #653: Activation reserve floor prevents single-model providers from fitting models that should fit
## Summary

The 5.5 GiB activation reserve floor in `UnifiedMemoryCap.resolvedActivationReserveBytes` is sized for the worst-case model (gemma-4-26b at B=8: 5.05 GiB peak). Operators serving only gpt-oss-20b (B=8 peak: 2.56 GiB) cannot lower it via `DARKBLOOM_ACTIVATION_RESERVE_GB` because the env override is clamped to `max(env, default)`.

## Reproduction

On a 32 GB Mac with gpt-oss-20b as the only model:

```
darkbloom doctor
# [FAIL] model fits in RAM — gpt-oss-20b needs ~20.0 GB but only 14.5 GB is usable
```

The 20.0 GB breaks down as:
- Model weights (on-disk × 1.2 overhead): ~13.5 GB
- Activation reserve: 5.5 GiB (sized for gemma-4 B=8)
- Minimum serveable KV: 1.0 GiB

But gpt-oss-20b's measured B=8 activation peak is only **2.56 GiB** (v0.8.0 benchmark data). A 2.75 GiB reserve (peak + slack) is sufficient, which would drop the model's need from 20.0 → 17.2 GB.

## Root cause

`resolvedActivationReserveBytes` enforces a floor:

```swift
return max(bytes, defaultActivationReserveBytes)  // line 307
```

The env var can only raise the reserve, never lower it. The comment says this prevents a "legacy 3" from recreating the B=8 OOM — but for operators serving only gpt-oss-20b, the floor is unnecessarily conservative and blocks models that would fit with a model-appropriate reserve.

## Impact

On 32 GB machines (the most common provider hardware), gpt-oss-20b fails the doctor check and refuses to load, even when the system has enough free memory. The operator sees "this box's RAM is too small" when the real issue is a one-size-fits-all reserve sized for a different model.

## Proposed fix

Allow `DARKBLOOM_ACTIVATION_RESERVE_GB` to set the reserve directly (remove the `max` floor). Operators serving a single model can size the reserve against their model's measured peak. The default remains 5.5 GiB for operators who don't set the env var.

The coordinator must be updated in lockstep when changing this on a fleet provider (`servabilityActivationFloorGB` in `servability.go`).

## Additional findings

`darkbloom status` reports "Inference memory: 28 GB available" (static `physical - 4`), while `darkbloom doctor` reports \~14.5 GB usable (live OS free+inactive minus load reserve). These are inconsistent and confusing to operators. The status number should also reflect actual system memory.

zulufoxtrot:
## Summary

Adds a per-model activation peak lookup so the provider automatically sizes the activation reserve for the model being served, instead of using a flat 5.5 GiB worst-case reserve for all models.

## Motivation

The 5.5 GiB activation reserve is sized for gemma-4-26b at B=8 (5.05 GiB measured peak). Operators serving only gpt-oss-20b (B=8 measured peak: 2.56 GiB) were forced to use the same reserve, wasting ~2.8 GiB of KV budget. On a 32 GB Mac, this made the difference between fitting and not fitting.

## How it works

Resolution order in `resolvedActivationReserveBytes`:
1. Explicit programmatic value (tests)
2. `DARKBLOOM_ACTIVATION_RESERVE_GB` env var (operator override)
3. **Per-model lookup** via `modelActivationPeaks` dictionary (NEW)
4. Default 5.5 GiB (worst-case: gemma-4-26b at B=8)

The lookup table matches model ID substrings case-insensitively:
- `gpt-oss` → 2.75 GiB (measured 2.56 peak + 0.19 slack)
- Unknown → 5.5 GiB (worst-case default)

## Changes

- **`UnifiedMemoryCap.swift`**: Add `modelActivationPeaks` dictionary, `activationReserveBytes(forModelID:)`, and `modelID` parameter to `resolvedActivationReserveBytes`, `loadHeadroomBytes`, `kvBudgetBytes`, `liveKVHeadroomBytes`, `canAdmit`
- **`ModelFitDiagnostic.swift`**: Add `modelID` parameter to `requiredGb`, thread through `diagnose` and alternatives check
- **`ProviderLoop+ModelLoading.swift`**: Convert `loadHeadroomGb` from `static let` to `static func` accepting `modelID`
- **`ProviderLoop+StartupPreload.swift`**: Pass model ID to `loadHeadroomGb`
- **`StandaloneServer.swift`**: Pass model ID to `loadHeadroomBytes`
- **`UnifiedMemoryCapTests.swift`**: Add 7 new tests for per-model lookup, env override priority, and model-specific headroom

## Impact

gpt-oss-20b on a 32 GB Mac:
- Before: needs ~20.0 GB (5.5 GiB reserve)
- After: needs ~17.2 GB (2.75 GiB reserve, auto-sized)

The env var still takes priority, so operators can override for fine-tuning. Unknown models fall back to the safe 5.5 GiB default.

## Coordinator note

The coordinator's `servabilityActivationFloorGB = 5.5` in `servability.go` is unaffected — it uses the worst-case floor for cold-load prediction, which is correct. The per-model sizing is LOCAL to the provider and makes the provider MORE conservative than the coordinator expects (smaller reserve = more KV budget = more headroom).

Closes #653

## End-to-end verification

Built from source on a 32 GB M5 Mac (macOS 26) and served `gpt-oss-20b` via the standalone local endpoint (`darkbloom start --local --model gpt-oss-20b --port 8000`), then issued a live OpenAI-compatible request:

```bash
curl http://localhost:8000/v1/chat/completions \
  -H "Authorization: Bearer $(cat ~/.darkbloom/local_token)" \
  -H "Content-Type: application/json" \
  -d '{"model":"gpt-oss-20b","messages":[{"role":"user","content":"Say hello in one word."}],"max_tokens":10}'
```

Result (22.8s wall clock, including cold load of 13.5 GB weights from disk):

```json
{"choices":[{"finish_reason":"length","index":0,
  "message":{"content":"<|channel|>analysis<|message|>The user says: \"Say hello","role":"assistant"}}],
 "created":1787313930,"id":"chatcmpl_237D395ED9F34F6785430A03","model":"gpt-oss-20b",
 "object":"chat.completion",
 "usage":{"completion_tokens":10,"prompt_tokens":73,"total_tokens":83}}
```

(The `<|channel|>analysis` prefix is gpt-oss's harmony reasoning-channel format; `max_tokens=10` truncated mid-stream, which is expected.)

**This is a pass that the old code could not do.** At request time the OS reported 18.8 GB available:

- Old flat reserve: gate requires ~20.0 GB → **rejects**, "Provider capacity is temporarily unavailable"
- Per-model reserve (this PR): gate requires ~17.2 GB → **admits**, loads, serves

The admission path exercised is exactly what this PR changes: `fastAdmissionReject` / `ensureModelLoaded` → `loadHeadroomGb(modelID:)` → `activationReserveBytes(forModelID: "gpt-oss-20b-…")` = 2.75 GiB.

Build-from-source notes (not needed for release installs):

- New Xcode no longer bundles the Metal compiler backend: run `xcodebuild -downloadComponent MetalToolchain` once (after `-runFirstLaunch`).
- The shader library is not an SPM artifact: `make provider-build` runs `scripts/fetch-metallib.sh <bin-path>` to compile and place `mlx.metallib` next to the binary. Skipping it yields `MLX error: Failed to load the default metallib` and failed model loads.

### DIFF
--- a/provider-swift/Sources/ProviderCore/Diagnostics/ModelFitDiagnostic.swift
+++ b/provider-swift/Sources/ProviderCore/Diagnostics/ModelFitDiagnostic.swift
@@ -24,12 +24,16 @@ public enum ModelFitDiagnostic {
     /// footprint plus one-request headroom — exactly `ensureModelLoaded`'s
     /// requirement. `estimatedMemoryGb` is the scanner's overhead-included size
     /// (the same value the runtime passes), not the raw on-disk bytes.
-    public static func requiredGb(estimatedMemoryGb: Double) -> Double {
+    ///
+    /// When `modelID` is provided, the headroom uses the model's measured
+    /// activation peak instead of the worst-case default — e.g. 3.75 GiB for
+    /// gpt-oss-20b (2.56 peak + 1 GiB min KV) vs 6.5 GiB for gemma-4.
+    public static func requiredGb(estimatedMemoryGb: Double, modelID: String? = nil) -> Double {
         // Cap-aware headroom (activation reserve + min serveable KV) so the
         // doctor's "needs ~X GB" matches what the runtime load gate requires.
         ModelLoadAdmission.requiredToLoadGb(
             weightsGb: estimatedMemoryGb,
-            headroomGb: Double(UnifiedMemoryCap.loadHeadroomBytes()) / (1024.0 * 1024.0 * 1024.0))
+            headroomGb: Double(UnifiedMemoryCap.loadHeadroomBytes(modelID: modelID)) / (1024.0 * 1024.0 * 1024.0))
     }
 
     /// The memory (GB) the provider would actually have free to load a model,
@@ -88,7 +92,7 @@ public enum ModelFitDiagnostic {
                 message: "couldn't determine the model size or available memory; skipping the fit check.",
                 fix: nil)
         }
-        let needed = requiredGb(estimatedMemoryGb: weightGb)
+        let needed = requiredGb(estimatedMemoryGb: weightGb, modelID: modelID)
         if needed <= usableGb {
             return Diagnostic(
                 section: .traffic, name: "model fits in RAM", level: .pass,
@@ -96,13 +100,13 @@ public enum ModelFitDiagnostic {
                 fix: nil)
         }
         let fits = alternatives
-            .filter { requiredGb(estimatedMemoryGb: $0.weightGb) <= usableGb }
+            .filter { requiredGb(estimatedMemoryGb: $0.weightGb, modelID: $0.id) <= usableGb }
             .sorted { $0.weightGb > $1.weightGb }
         let suggestion: String
         if fits.isEmpty {
             suggestion = "this box's RAM is too small for the models on this network; consider a machine with more unified memory."
         } else {
-            let list = fits.prefix(3).map { "\($0.id) (~\(fmt(requiredGb(estimatedMemoryGb: $0.weightGb))) GB)" }.joined(separator: ", ")
+            let list = fits.prefix(3).map { "\($0.id) (~\(fmt(requiredGb(estimatedMemoryGb: $0.weightGb, modelID: $0.id))) GB)" }.joined(separator: ", ")
             suggestion = "set `enabled_models` in provider.toml to a model that fits: \(list)."
         }
         return Diagnostic(
--- a/provider-swift/Sources/ProviderCore/Inference/UnifiedMemoryCap.swift
+++ b/provider-swift/Sources/ProviderCore/Inference/UnifiedMemoryCap.swift
@@ -122,10 +122,15 @@ public enum UnifiedMemoryCap {
     /// headroom was LESS than the 3 GiB activation reserve, so a near-cap model
     /// loaded with zero serveable KV). Returns
     /// `activationReserve + minimumLoadKV`.
+    ///
+    /// When `modelID` is provided, the activation reserve is looked up from
+    /// ``modelActivationPeaks`` so the headroom matches the model's measured
+    /// B=8 peak instead of the worst-case default.
     public static func loadHeadroomBytes(
-        activationReserveBytes: UInt64? = nil
+        activationReserveBytes: UInt64? = nil,
+        modelID: String? = nil
     ) -> UInt64 {
-        let activations = activationReserveBytes ?? resolvedActivationReserveBytes()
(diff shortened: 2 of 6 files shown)


powtac:
### Qwen coverage gap on the catalog 32 GB tier

This per-model lookup currently covers `gpt-oss`, while unknown models retain the 5.5 GiB fallback. A current catalog model appears to expose the remaining gap:

- Hardware: Mac mini M4, 32 GB unified memory, macOS 26
- Provider: Darkbloom 0.8.10
- Model: `qwen3.6-35b-a3b-vl-mtp-mxfp8`
- Catalog: `min_ram_gb: 32`, 21.3 GB decimal download size
- `darkbloom models list`: 19.8 GB weights, 23.8 GB estimated memory
- `darkbloom doctor`: needs ~30.3 GB
- Live local-endpoint request: immediate `Provider capacity is temporarily unavailable`, with no model load
- After `sudo purge`: usable headroom rose from 17.3 GB to 20.4 GB, but the required value remained 30.3 GB

The reported 30.3 GB is exactly consistent with 23.8 GB estimated model memory + 5.5 GiB activation reserve + 1.0 GiB minimum KV. Because Qwen is not in `modelActivationPeaks`, this PR appears to leave the catalog 32 GB tier unable to admit it.

Would it make sense to add a measured Qwen activation peak and explicit lookup entry before merging, or otherwise raise its catalog `min_ram_gb` if the safe 5.5 GiB fallback must remain? I am deliberately not proposing a lower Qwen reserve without a measured B=8 peak. I can run a targeted benchmark or validation build on the M4 32 GB machine if useful.

Gajesh2007:
@codex can you review this and also tell me how we can have this activation sizing dynamically as we add models and not hard code it all the time?

Gajesh2007:
Superseded by #791, which lands the per-model floor machinery from #683 (threaded through the post-load probes, runtime KV budget, and engine grants — the pieces this PR left on the flat reserve) plus measured floors for the qwen family and measured resident weights. The raise-only env clamp is preserved there; the underlying #653 problem is solved via measured floors instead. Also: the Qwen coverage gap flagged in this thread is addressed — qwen3.6/3.5 measured at 4.0 GiB (docs/reports/2026-08-30-activation-floor-measurements.md in #791).

PR #434, opened 2026-06-21
feat(coordinator): route Gemma 4 only to dedicated providers; 429 shed

Gajesh2007:
## What & why

Gemma 4 destabilizes when it shares a Mac with other models/work (memory contention, OOM/jetsam, ragged-batch NaN, Metal resource-count leaks). This restricts **a dedicated model family (default `gemma-4`) to providers whose entire advertised catalog is that family** — i.e. machines dedicated to Gemma 4. Those dedicated boxes are the only routing candidates *and* the only providers counted when deciding whether to accept or shed a request to OpenRouter.

Coordinator-only — no provider/Swift or protocol changes.

## Routing behavior: before → after

**Before** — a Gemma 4 request could land on any box serving it (including mixed boxes where it contends), and "no provider" shed as `503`:

```mermaid
flowchart TD
    Req["Gemma 4 request"] --> Gate{"Provider serves gemma-4?"}
    Gate -->|"Dedicated box ✓<br/>(gemma-4 only)"| Cand["Candidate pool"]
    Gate -->|"Mixed box ✓<br/>(gemma-4 + qwen/gpt-oss)"| Cand
    Cand --> Cap{"Capacity?"}
    Cap -->|"Yes"| Route["Route to best score —<br/>⚠️ often a MIXED box →<br/>Gemma 4 contends → OOM / NaN / leak"]
    Cap -->|"All busy"| R429["429 rate_limit"]
    Cap -->|"None eligible"| R503["❌ 503 model_unavailable<br/>(OpenRouter may derank us)"]
```

**After** — mixed boxes are filtered out at the single shared gate (routing *and* capacity preflight), so only dedicated boxes are eligible; "no dedicated box" sheds as `429`:

```mermaid
flowchart TD
    Req["Gemma 4 request"] --> Gate{"Provider serves gemma-4?"}
    Gate -->|"Dedicated box ✓"| Ded{"Dedicated to gemma-4?<br/>(advertised catalog == gemma-4 only)"}
    Gate -->|"Mixed box ✓"| Ded
    Ded -->|"No — mixed"| Drop["🚫 Excluded from routing<br/>AND capacity count<br/>(self-route to own box exempt)"]
    Ded -->|"Yes — dedicated"| Cand["Candidate pool =<br/>dedicated boxes only"]
    Cand --> Cap{"Capacity?"}
    Cap -->|"Yes"| Route["✅ Route to a DEDICATED box →<br/>Gemma 4 isolated, no contention"]
[...]

PR #664, opened 2026-08-22
fix(fleet): make "connected but not advertising" explainable

Gajesh2007:
## The incident

A provider reported a node that served **zero requests for hours** while every surface said it was healthy:

- `trust=hardware`, online, proofs `mdm+mda`, attestation challenges answered every 5 minutes
- `darkbloom doctor`: **30 PASS / 0 FAIL**
- `qwen3.6-35b-a3b-vl-mtp-mxfp8` warm, `preload_models` / `enabled_models` set
- clean 0.8.9 reinstall done, hash `02bab0f4…c0e2cab5`, the runtime-mismatch banner cleared

…and yet: the node's `/v1/models` was empty and self-test requests returned `model_not_loaded`, 20+ minutes after the trust grant. The network page said **88 connected / 85 advertising** with no way to learn which three or why. They also saw four provider identities minted in one day by the update churn.

Three asks, answered below: **(a)** why is the catalog record empty when everything is green, **(b)** how do I force a re-sync without another reinstall, **(c)** the updater failure mode will strand any auto-updating provider.

## (a) Why the record was empty — and why nothing said so

Whether a connected provider advertises is the conjunction of **~15 independent gates** spread across four call sites:

| Layer | Gates |
|---|---|
| Liveness core (`providerLivenessGateLocked`) | status, private-only, trust floor, `runtime_verified`, challenge freshness ≤ 16 min |
| Private-text chokepoint (`providerSupportsPrivateTextLocked`) | encryption key, backend support, encrypted chunks, `runtime_manifest_checked`, coordinator-verified SIP, code attestation, five privacy capabilities |
| Public listing (`ListModels`) | its own subset of the above |
| Per model | catalog membership, weight-hash match, `template_render_ok`, dedicated-box isolation |

**Exactly one of them reached the operator** — the runtime hash, as a UI banner. The rest failed silently. `doctor` is provider-side and cannot see any of them, which is why it reported 30 PASS / 0 FAIL on a fenced machine.

Worse, the gates are asymmetric in ways that hide the problem: **`ListModels` does not apply the `runtime_verified` or challenge-freshness checks that dispatch applies**, and it does not apply the per-model template-render or dedicated-box gates either. A machine can therefore be *counted as advertising* on the network page and still receive nothing.

Two things were also worth answering directly:

- **Four identities in one day is expected, not a symptom.** `provider_id` is a fresh UUID per WebSocket connection (`coordinator/api/provider.go:125`); stable identity is serial / Secure Enclave key / account (`stableProviderIdentityLocked`). Four IDs means four reconnects from the update churn, not four registrations competing. Live duplicates are already evicted by serial (`DisconnectDuplicatesBySerial`).
- **Warm ≠ advertised.** Advertising comes from the register-time `models[]` list; heartbeats can only *filter* that set (`canonicalHeartbeatModelState`), never add to it. A locally-warm model that never made it into registration is invisible to the coordinator.

## The fix — the coordinator reports its own verdict

`GET /v1/me/providers` now carries a `routing` block per machine:

```json
"routing": {
  "advertising": true,
  "routable": false,
  "owner_routable": false,
  "blockers": ["attestation_challenge_stale"],
  "models": [
    {"id": "qwen3.6-35b-a3b-vl-mtp-mxfp8", "publicly_listed": false,
     "routable": false, "owner_routable": false,
     "blockers": ["model_weight_hash_mismatch"]}
  ],
  "challenge_max_age_seconds": 960
}
```

`advertising` and `routable` are separate on purpose — that is the asymmetry above, made visible rather than papered over. `routable` applies the gates dispatch applies, including the per-model template-render and dedicated-box fences.

**It cannot drift from the router, because it is not a second implementation.** Each gate now has one body that returns the reason it refused; the booleans the router calls are one-line delegates asking whether that reason is empty:

```go
func (r *Registry) providerSupportsPrivateTextLocked(p *Provider) bool {
	return r.providerPrivateTextBlockerLocked(p) == ""
}
```

`ListModels`' per-provider filter moves into the same shared `publicListingBlockerLocked`, and `OwnedModels` drops its hand-rolled copy of the liveness gate.

## (b) Forcing a re-sync

With the verdict exposed, "force a re-sync" stops being a guess. Every blocker maps to a specific remedy in the console (`darkbloom restart` for a stale challenge, `darkbloom models pull` for a weight-hash mismatch, `csrutil enable` for SIP, and so on) — no identity-burning reinstall. Note that the machine-level state was never sticky: a runtime mismatch does not blacklist a serial or SE key; registration re-evaluates `RuntimeVerified` from scratch.

## (c) The updater failure mode

Confirmed and fixed. Two pieces of the provider derive their world from the running executable's own path, and both go wrong at once when that path is a staging tree:

- `SelfUpdater.resolvedInstallRoot()` walks up from the executable, so the **staging directory becomes the install root** — every subsequent stage, commit and rollback happens inside it and the real install is never repaired. Self-perpetuating.
- `BinaryHasher.locateMetallib()` resolves `mlx.metallib` next to the executable, so the coordinator receives the staged tree's hashes, they match no release, and the node is derouted while still connected.

`InstallLocation` classifies the running executable against the six transient directories the updater creates, referencing `SelfUpdater.stagingDirPrefix` and `UpdateRecoveryStore.staleAppAsidePrefix` directly rather than copying the strings. `doctor` now **FAILS** on a transient verdict, and both `darkbloom start` entry points refuse — including `runForeground`, the process launchd actually runs and therefore the only one a failed update can strand.

## Before / after

```mermaid
flowchart TB
  subgraph Before["BEFORE"]
    direction TB
    B1["operator: node serves nothing"] --> B2["darkbloom doctor"]
    B2 --> B3["30 PASS / 0 FAIL<br/>(no coordinator gate is visible here)"]
    B1 --> B4["console dashboard"]
    B4 --> B5["derives 4 of ~15 gates from<br/>exposed booleans"]
    B5 --> B6["verdict: ROUTABLE<br/>2 informational notes"]
    B1 --> B7["network page"]
    B7 --> B8["88 connected / 85 advertising<br/>no drill-down"]
    B3 --> B9["only remaining move:<br/>full reinstall (new identity)"]
    B6 --> B9
    B8 --> B9
  end
```

```mermaid
flowchart TB
  subgraph After["AFTER"]
    direction TB
    A1["operator: node serves nothing"] --> A2["GET /v1/me/providers"]
    A2 --> A3["registry.RoutingDiagnostics<br/>(same predicates the router uses)"]
    A3 --> A4["advertising: true<br/>routable: false<br/>owner_routable: false"]
    A3 --> A5["blockers:<br/>attestation_challenge_stale"]
    A3 --> A6["per model:<br/>qwen → weight_hash_mismatch<br/>gemma → template_render_broken<br/>local → not_in_catalog"]
    A4 --> A7["console verdict: BLOCKED<br/>each row with a copyable fix"]
    A5 --> A7
    A6 --> A7
    A1 --> A8["darkbloom doctor"]
    A8 --> A9["[FAIL] install location:<br/>running from .update-staging-…"]
  end
```

Code-level delta:

```mermaid
flowchart LR
  subgraph CB["Before"]
    C1["providerSupportsPrivateTextLocked<br/>(7 gates, bool)"]
    C2["providerLivenessGateLocked<br/>(6 gates, bool)"]
    C3["ListModels<br/>(inline copy of 4 gates)"]
    C4["OwnedModels<br/>(inline copy of 6 gates)"]
    C5["warnings.ts<br/>(TS re-derivation of 4 gates)"]
  end
  subgraph CA["After"]
    D1["providerPrivateTextBlockerLocked<br/>→ RoutingBlocker"]
    D2["providerLivenessBlockerLocked<br/>→ RoutingBlocker"]
    D3["publicListingBlockerLocked<br/>→ RoutingBlocker"]
    D5["modelBlockersLocked<br/>→ catalog / hash / template / dedicated"]
    D1 --> D2 --> D4["RoutingDiagnostics"]
    D3 --> D4
    D5 --> D4
    D1 -.delegate.-> E1["providerSupportsPrivateTextLocked (bool)"]
    D2 -.delegate.-> E2["providerLivenessGateLocked (bool)"]
    D3 -.-> E3[ListModels]
    E2 -.-> E4[OwnedModels]
    D4 --> F1["/v1/me/providers .routing"]
    F1 --> F2["routing-blockers.ts + fixes.ts"]
  end
```

## Verification

**Coordinator** (`go test ./...`, full suite green, `-race` clean). The drift tests mutate one gate at a time across a 17-case matrix — including code attestation and an unrecognised trust level — and assert the diagnostic agrees with `ListModels`, with `OwnedModels`, and with the production `ReserveProviderEx` path **in both directions**, so neither an over- nor under-optimistic verdict can ship. HTTP-level tests decode the wire JSON so the field names the UI keys off are pinned.

**Console UI** (533 vitest tests, 0 eslint errors, `next build` clean). Includes the fix-table coverage test `fixes.ts` has always claimed to have: every blocker must resolve to an explicit remedy, not the generic fallback.

Driving the console from the **real bytes** a live coordinator emits for the reported machine shape — online, hardware trust, runtime verified, zero failed challenges, model warm:

[GET /v1/me/providers routing block from a live coordinator](https://cursor.com/agents/bc-08c3df13-d956-5fba-9a68-59ffb1ccd7e1/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Frouting_verdict_api_response_v2.png)

[Dashboard output for the same machine, before and after](https://cursor.com/agents/bc-08c3df13-d956-5fba-9a68-59ffb1ccd7e1/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Frouting_verdict_dashboard_before_after_v2.png)

The dashboard went from **ROUTABLE** with two informational notes to **BLOCKED** with the stale challenge, the Qwen build's weight-hash mismatch, and the Gemma build's failed template render — each with the command that fixes it.

**Provider.** MLX needs Metal, so `swift test` runs on the macOS CI runner; the pure classifier was additionally compiled and executed here against a copy with only the two constant references substituted for their literals:

[install-location classifier verification](https://cursor.com/agents/bc-08c3df13-d956-5fba-9a68-59ffb1ccd7e1/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Finstall_location_check_v2.png)

## Follow-ups (not in this PR)

- `revalidateConnectedProvidersAgainstRuntimePolicy` says it deroutes "until the next successful challenge re-verifies it", but when the manifest is withdrawn (`knownRuntimeManifest == nil`) the challenge handler's re-verification block is skipped entirely, so nothing can restore `RuntimeVerified` short of a reconnect. Unreachable in prod today (a manifest is always configured) but the comment promises a recovery path that does not exist.
- `Register` does not reject an empty model id, so `ListModels` can emit an `{"id": ""}` row that the diagnostic (correctly) skips. Best fixed at registration.
- The server-side `needsAttention` count on `/v1/me/summary` still uses the old approximation and never consults `mp.Routing`, so the header count and the per-card verdict can disagree — the same class of inconsistency this PR removes on the client.
- The provider reports advertising state only at register / `models_update`. A `darkbloom models refresh` verb that re-sends `models_update` on demand would give operators a re-sync that costs neither a reconnect nor a reinstall.

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#team-pull-requests">enable</a> in settings.</sub>

[Slack Thread](https://darkbloom.slack.com/archives/C0B0CAQC8P5/p1787343622971919?thread_ts=1787343622.971919&cid=C0B0CAQC8P5)

<div><a href="https://cursor.com/agents/bc-08c3df13-d956-5fba-9a68-59ffb1ccd7e1?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-08c3df13-d956-5fba-9a68-59ffb1ccd7e1&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/664"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789973642&installation_model_id=21733&pr_number=664&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F664&signature=1653e8512cbc24c15dc382e23a9ead39f8686299accf4e4fc7c094ae6e16aac6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/console-ui/src/app/providers/dashboard/fixes-coverage.test.ts
+++ b/console-ui/src/app/providers/dashboard/fixes-coverage.test.ts
@@ -35,6 +35,7 @@ const MODEL_BLOCKERS: RoutingBlocker[] = [
   "model_not_in_catalog",
   "model_weight_hash_mismatch",
   "model_template_render_broken",
+  "model_requires_dedicated_box",
 ];
 
 describe("every routing blocker is actionable", () => {
@@ -58,6 +59,7 @@ describe("every routing blocker is actionable", () => {
           {
             id: "some/model-with:colons",
             publicly_listed: false,
+            routable: false,
             owner_routable: false,
             blockers: [blocker],
           },
--- a/console-ui/src/app/providers/dashboard/fixes-coverage.test.ts
+++ b/console-ui/src/app/providers/dashboard/fixes-coverage.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, it } from "vitest";
+
+import { routingWarnings } from "../routing-blockers";
+import type { RoutingBlocker } from "../types";
+import { hasFix, resolveFix } from "./fixes";
+import { makeProvider, makeRouting } from "./testFixtures";
+
+// fixes.ts states the invariant "every warning id produced by computeWarnings()
+// must have an entry here" but the test enforcing it never existed. This is it,
+// for the routing verdict — the one warning family whose ids are generated
+// rather than hand-written, and therefore the one that can silently grow past
+// the fix table.
+
+const MACHINE_BLOCKERS: RoutingBlocker[] = [
+  "offline",
+  "untrusted",
+  "private_only",
+  "trust_below_minimum",
+  "runtime_hash_mismatch",
+  "attestation_challenge_never_passed",
+  "attestation_challenge_stale",
+  "no_encryption_key",
+  "unsupported_backend",
+  "unencrypted_response_chunks",
+  "runtime_manifest_unchecked",
+  "sip_unverified",
+  "code_attestation_missing",
+  "privacy_capabilities_missing",
+  "privacy_capabilities_incomplete",
+  "no_models_registered",
+  "no_routable_models",
+];
+
+const MODEL_BLOCKERS: RoutingBlocker[] = [
+  "model_not_in_catalog",
+  "model_weight_hash_mismatch",
+  "model_template_render_broken",
+];
+
+describe("every routing blocker is actionable", () => {
+  it.each(MACHINE_BLOCKERS)("machine blocker %s has an explicit fix", (blocker) => {
+    const p = makeProvider({
+      status: "serving",
+      online: true,
+      routing: makeRouting({ routable: false, blockers: [blocker], models: [] }),
+    });
+    const warnings = routingWarnings(p)!;
+    expect(warnings).toHaveLength(1);
+    expect(hasFix(warnings[0].id)).toBe(true);
+  });
+
+  it.each(MODEL_BLOCKERS)("model blocker %s has an explicit fix", (blocker) => {
+    const p = makeProvider({
+      status: "serving",
+      online: true,
+      routing: makeRouting({
+        models: [
+          {
+            id: "some/model-with:colons",
+            publicly_listed: false,
+            owner_routable: false,
+            blockers: [blocker],
+          },
+        ],
+      }),
+    });
+    const warnings = routingWarnings(p)!;
+    expect(warnings).toHaveLength(1);
+    // Model ids can contain colons; the fix lookup must still resolve.
+    expect(hasFix(warnings[0].id)).toBe(true);
+    expect(resolveFix(warnings[0].id).label).not.toBe("View setup guide");
+  });
+
+  it("falls back to the generic fix for an id it has never seen", () => {
+    expect(resolveFix("routing:some_future_blocker").label).toBe("View setup guide");
+  });
+});
--- a/console-ui/src/app/providers/dashboard/fixes.ts
+++ b/console-ui/src/app/providers/dashboard/fixes.ts
@@ -248,6 +248,11 @@ const FIX_TABLE: Record<string, FixAction> = {
     command: "darkbloom models pull",
     note: "The chat template failed its render self-check, so every request shape is fenced away.",
   },
+  "routing:model:model_requires_dedicated_box": {
+    kind: "guidance",
+    label: "Serve only this model",
+    note: "Remove the other model families from `enabled_models` and restart, or accept that this build stays unrouted on a mixed box.",
+  },
 };
 
 /**
--- a/console-ui/src/app/providers/dashboard/fixes.ts
+++ b/console-ui/src/app/providers/dashboard/fixes.ts
@@ -132,7 +132,136 @@ const FIX_TABLE: Record<string, FixAction> = {
     label: "No action needed",
     note: "Routing starts within ~5 minutes of the first attestation challenge.",
   },
+
+  // ── Coordinator routing verdict ────────────────────────────────────────
+  // One entry per RoutingBlocker the coordinator can report (see
+  // coordinator/registry/routing_diagnostics.go). Per-model blockers are keyed
+  // without the model id; resolveFix strips it.
+  "routing:offline": {
+    kind: "command",
+    label: "Start the provider",
+    command: "darkbloom start",
+  },
+  "routing:untrusted": {
+    kind: "command",
+    label: "Restart & re-link",
+    command: "darkbloom restart && darkbloom login",
+    note: "Clears the untrust, then re-attests this device.",
+  },
+  "routing:private_only": {
+    kind: "guidance",
+    label: "Intentional — private mode",
+    note: "Disable private-only mode if you want this machine to earn from public network traffic.",
+  },
+  "routing:trust_below_minimum": {
+    kind: "link",
+    label: "Complete hardware attestation",
+    href: "/providers/setup",
+    note: "The network requires MDM enrollment + Apple Device Attestation.",
+  },
+  "routing:runtime_hash_mismatch": {
+    kind: "command",
+    label: "Reinstall",
+    command: INSTALL_COMMAND,
+    note: "Replaces a self-updated or partial bundle with a manifest-matching one.",
+  },
+  "routing:attestation_challenge_never_passed": {
+    kind: "guidance",
+    label: "No action needed",
+    note: "Routing starts within ~5 minutes of the first attestation challenge.",
+  },
+  "routing:attestation_challenge_stale": {
+    kind: "command",
+    label: "Force a fresh handshake",
+    command: "darkbloom restart",
+    note: "Re-runs the attestation challenge so routing can resume.",
+  },
+  "routing:no_encryption_key": {
+    kind: "command",
+    label: "Restart the provider",
+    command: "darkbloom restart",
+    note: "Republishes this machine's end-to-end encryption key.",
+  },
+  "routing:unsupported_backend": {
+    kind: "command",
+    label: "Reinstall",
+    command: INSTALL_COMMAND,
+  },
+  "routing:unencrypted_response_chunks": {
+    kind: "command",
+    label: "Reinstall",
+    command: INSTALL_COMMAND,
+  },
+  "routing:runtime_manifest_unchecked": {
+    kind: "command",
+    label: "Reinstall",
+    command: INSTALL_COMMAND,
+    note: "Installs a build the coordinator can match to a registered release.",
+  },
+  "routing:sip_unverified": {
+    kind: "guidance",
+    label: "Re-enable SIP",
+    note: "Boot into Recovery, run `csrutil enable`, reboot, then `darkbloom restart`.",
+  },
+  "routing:code_attestation_missing": {
+    kind: "guidance",
+    label: "Keep the Mac awake",
+    note: "Code-identity attestation completes over an APNs push; it needs the Mac awake and reachable.",
+  },
+  "routing:privacy_capabilities_missing": {
+    kind: "command",
+    label: "Reinstall",
+    command: INSTALL_COMMAND,
+  },
+  "routing:privacy_capabilities_incomplete": {
+    kind: "command",
+    label: "Restart the provider",
+    command: "darkbloom restart",
+    note: "If it persists after a restart, reinstall with the official installer.",
+  },
+  "routing:no_models_registered": {
+    kind: "command",
+    label: "Check local models",
+    command: "darkbloom models",
+    note: "Confirm the weights finished downloading and that `enabled_models` is not filtering them out, then `darkbloom restart`.",
+  },
+  "routing:no_routable_models": {
+    kind: "link",
+    label: "Add an approved model",
+    href: "/models",
+    note: "Download a catalog model, then run `darkbloom restart`.",
+  },
+  "routing:model:model_not_in_catalog": {
+    kind: "guidance",
+    label: "Local-only model",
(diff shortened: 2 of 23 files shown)


hnshah:
Operator note from multi-host providers (sanitized):

This PR’s framing matches a recurring failure mode we hit: **daemon/trust/doctor all green while paid pull is ~0** for hours — e.g. network underlay dead under a live process, or post-repair under-route while warm models still show local. Provider-side `doctor` cannot see coordinator advertising/dispatch gates; making those gates **explainable** is exactly the product gap.

We won’t open a competing PR. Happy to validate any new operator-facing fields on a lab host after merge.

(Related: our fleet feedback also called out serving ≠ earning as P0.)

brandon-eigenlabs:
A concrete instance of "connected but not advertising / not selected, with no
way to learn why", plus a note that most of the data needed to answer it is
already computed and discarded.

## The instance: dedicated-family exclusion is invisible to the operator

`EIGENINFERENCE_DEDICATED_MODELS` (default `gemma-4`,
`cmd/coordinator/main.go:167`, set in `deploy/environments/prod.env:26`)
restricts a model family to providers whose ENTIRE advertised catalog matches
the pattern. The predicate is `providerDedicatedToPatternLocked`
(`registry/dedicated_models.go:74-86`): every catalog-allowed advertised build
must contain the pattern, and at least one must exist.

The operator-visible consequence is sharp and, as far as I can tell,
undiscoverable from the provider side:

    advertises gemma-4 ONLY           -> dedicated: receives gemma traffic
    advertises gemma-4 + gpt-oss      -> receives ZERO gemma traffic
    advertises gpt-oss only           -> unaffected (gpt-oss is not dedicated)

So an operator who is serving gemma-4 and then downloads a second model to
"help more" silently removes their box from the gemma pool. Nothing tells them:

- `models_update` / `desired_models` carry no notion of it —
  `DesiredModelsForProvider` (`registry/registry.go:3607-3656`) emits alias /
  desired / previous builds only, and is not dedication-aware.
- The word `dedicated` does not appear anywhere in `coordinator/protocol/`, so
  it never crosses the wire in any form.
- `provider-swift` has no concept of it (the only `dedicated` hits in
  `Sources/` are unrelated SSD-cache paths).
- `docs/provider/` does not mention model-family isolation at all, so there is
  no doc an operator could have read.

To be clear this is intended behaviour, not a defect: #434 introduced it
because "Gemma 4 destabilizes when it shares a Mac with other models/work
(memory contention, OOM/jetsam, ragged-batch NaN, Metal resource-count leaks)",
and #439 fixed the pool-collapse follow-on. The gap is purely
observability — the rule is correct and the operator cannot see it.

## The data already exists server-side

The warm-pool controller already computes exactly the per-reason breakdown this
issue is asking for. `WarmPoolSnapshot` carries `ColdIneligible` plus
`ColdDisqualifiers map[string]int`
(`registry/warm_pool_controller.go:75-76`), populated by
`warmPoolCandidateReasonLocked` (`:687-746`) with a dedicated bucket per cause:

    warmColdOfflineUntrust   warmColdTooLarge
    warmColdPendingLoad      warmColdNoFreeForLoad
    warmColdNotIdle          warmColdNotServing
    warmColdThermal          warmColdDedicated
    warmColdTrust            warmColdStaleChallenge

`warmPoolCandidateReasonLocked`'s own docstring says the granular reason exists
precisely because "a single boolean helper cannot preserve" it, and that the
gate ordering determines which reason wins.

Today that map is (a) logged only when the model is dedicated
(`:375-384` — gated on `c.registry.IsDedicatedModel(model)`) and (b) surfaced
in the admin utilization snapshot. It is counts-only, no provider identities,
which is the right privacy posture for a fleet-wide view — but it means the one
operator who needs the answer cannot get it.

## Three tiers, cheapest first

**1. Local-only `doctor` warning — no protocol change.** `doctor` can already
see the shape that triggers the rule (its own advertised set), so it can warn
without knowing the server's policy:

> WARN: this box advertises `gemma-4-26b-qat-4bit` alongside 2 other models.
> Some model families are isolated to machines that serve ONLY that family; if
> `gemma-4` is one of them on this network, this box will receive no gemma
> traffic. Serve it alone to join that pool.

Deliberately hedged — it must NOT hardcode "gemma-4 is dedicated", because the
pattern is a coordinator env var and a client-side copy would go stale the
moment it changed. Suggest, do not assert, and never block: a mixed box is a
legitimate configuration that serves the non-dedicated families fine.

**2. Put the reason on the wire (the real fix).** Send the provider the
exclusion reason the coordinator already computed — as a field on the heartbeat
ack, or a push in the shape of the existing `trust_status` message. Then
`darkbloom status` can print

    qwen3.6-35b-a3b-vl-mtp-mxfp8   eligible
    gemma-4-26b-qat-4bit           EXCLUDED: not dedicated to this family

...and no client-side policy copy exists. This also generalises: the same
channel answers `warmColdStaleChallenge`, `warmColdTooLarge`,
`warmColdNoFreeForLoad` and the rest, which are all currently invisible in the
same way. That directly addresses this issue's "88 connected / 85 advertising
with no way to learn which three or why" — the three ARE in a
`ColdDisqualifiers` bucket right now.

**3. Document it.** One paragraph in `docs/provider/` on model-family
isolation, and a line in the `enabled_models` / `preload_models` reference
noting that adding a model can remove a box from an isolated family's pool.

Tier 1 alone would have prevented the self-inflicted version of this. Tier 2 is
what closes the class.

One caveat on tier 2's scope: `warmPoolCandidateReasonLocked` returns the FIRST
failing gate in a fixed order, so a box failing several gates sees only one
reason at a time. That is fine for an operator (fix it, re-run, see the next
one) but worth stating in the UI so it does not read as "this is the only
problem".

Issue #675, opened 2026-08-23
[bug] Three different answers to "how much memory can this box use", two for "how much does this model need" (0.8.10)


redstar-systems:
### Component

provider-swift (Swift CLI)

### What happened?

One binary, one machine, same minute, disagreeing with itself.

### Same model, two memory requirements

`mlx-community/Qwen3-TTS-12Hz-1.7B-Base-bf16`:

```
$ darkbloom models list
ID                                           TYPE       QUANT  SIZE    EST MEM
mlx-community/Qwen3-TTS-12Hz-1.7B-Base-bf16  qwen3_tts  bf16   4.2 GB  5.1 GB

$ darkbloom doctor
[FAIL] model fits in RAM — mlx-community/Qwen3-TTS-12Hz-1.7B-Base-bf16 needs ~11.6 GB
       but only 6.6 GB is usable — it will show online but every request fails to load.
```

**5.1 GB** vs **~11.6 GB** for the same model.

### Available memory, three figures

```
doctor:  only 6.6 GB is usable
doctor:  metal gpu: Apple M2 Max, 21 GB working set
status:  Inference memory: 28 GB available
```

The `6.6 GB usable` figure did not change between two runs several minutes apart, while system-wide
free memory went from roughly 200 MB to 67%. So it does not appear to track actual availability.

### Consequence in the catalog picker

`darkbloom start` offers **Qwen 3.6 35B A3B at 21.3 GB** on this box, whose reported Metal working
set is **21 GB**, under a header reading:

```
Select models (RAM: 32 GB)
0 selected · 0.0 GB total · all models can be served simultaneously
```

The picker appears to gate on total RAM (`min_ram_gb` ≤ 32) rather than on the GPU working set that
`doctor` reports. Weights fitting RAM is not the same as the model fitting Metal residency, and on a
32 GB machine that difference is the whole margin.

I did not test whether selecting Qwen is actually permitted or what happens if you do, so I am
describing what the picker offers and on what basis, not asserting it would fail.

Related to #653 from the opposite direction: that one blocks models that should fit, this one
advertises one that likely cannot.

### Expected behavior

One memory model, shared by `models list`, `doctor`, `status`, and the catalog picker.

Whichever figure is authoritative, the other surfaces should quote it rather than compute their own.
In particular the picker should gate on the same number `doctor` fails on, so a model cannot be
offered and then rejected by the next command you run.

### Steps to reproduce

[...]

PR #683, opened 2026-08-23
Resolve the activation reserve per serving set (measured per-model floors) — fixes the unserveable 24 GB tier

### LINKED ISSUE #653: Activation reserve floor prevents single-model providers from fitting models that should fit

jkjoplin:
## Summary

Fixes #653: the flat 5.5 GiB activation reserve — sized against gemma-4-26b's measured B=8 peak — is charged to every model, including gpt-oss-20b whose worst measured B=8 peak is **3.20 GiB** (compiled; 2.56 eager — same sweep, `*-paged-gate-2026-07-09.md`). On the catalog's own `min_ram_gb: 24` tier that difference is the whole margin: `needs = 13.5 (padded weights) + 5.5 (reserve) + 1.0 (min KV) = 20.0 GB` requires ~22.7 GiB reclaimable out of 24 — below the macOS floor — so every 24 GB box **shows online and 429s every request, permanently** (verified live on an M4 Pro 24 GB: HTTP 429 in ~13 ms with `trust=hardware/online`; measurements in [the #653 corroboration comment](https://github.com/Layr-Labs/d-inference/issues/653#issuecomment-5384189907)).

The reserve stays **measurement-driven and raise-only** — both documented rationales survive — but resolves per serving set instead of one-size-fits-worst-model:

- `UnifiedMemoryCap.measuredActivationFloorsBytes`: measured per-model floors, catalog-id keyed, exact match only (`gpt-oss-20b: 3.5 GiB` = worst measured B=8 peak 3.20 compiled + 0.30 slack; the default's own basis covers gemma's 5.05 eager / 5.34 compiled). A serving set takes the **max** over members; any unmeasured member pins the flat 5.5 default.
- `DARKBLOOM_ACTIVATION_RESERVE_GB` stays **raise-only against the resolved floor** — the stale-legacy-`3` OOM scenario from the doc comment still cannot happen.
- The resolved reserve flows to **every** consumer in lockstep, so neither direction of the admit-then-reject trap can open:
  - load gate + startup preload + doctor (`ProviderLoop.loadHeadroomGb`, computed over the live set),
  - the **measured post-load / post-bridge serveability guards** (`KVHeadroomProbe` gains `activationReserveBytes`; without this the probe would unload the freshly-loaded model against the flat reserve — churn instead of serving),
  - `free_for_load_gb` reporting (`ProviderLoop+Capacity`),
  - runtime KV admission (`GlobalKVCacheBudget` gains an actor setter; re-pushed on every serving-set mutation),
  - **engine KV grants and the heartbeat budget clamp** (`fleetKVBudgetBytes`, `EngineV2KVSizing`, `FleetKVContext`) so `active_token_budget_max` warm reports converge with the coordinator's cold estimate.
- The floor resolves over **advertised ∪ resident** models: a hard-swapped build that leaves `advertisedModels` but is still draining decodes keeps pinning its floor until it actually unloads (`unloadModel` refreshes before survivors regrow); prefetch pushes the grown set's floor **before** the new build becomes advertised/loadable.
- Doctor's per-model verdict uses the box's **actual serving-set floor** (scan ∩ `enabled_models`), so it cannot print a PASS the daemon would still gate; the "alternatives that fit" suggestions keep each candidate's solo floor (they model `enabled_models = [candidate]`).
- Coordinator mirror (`servability.go`): `servabilityModelActivationFloorsGB` + a third version regime in `servabilityActivationFloor(version, modelID)`, following the staged-rollout shape the 3.0→5.5 move established. Per the file's "Mirroring, not modelling" note this is still a mirror; for a multi-model provider the per-model figure is a lower bound (optimistic — the sanctioned fail-open direction; warm reports take over as slots load).

**Effect on the 24 GB tier**: `needs` drops 20.0 → 18.0 GB. Headless steady-state usable measures ~16–18 GB on the repro box — tight but genuinely reachable, instead of arithmetically impossible. 32 GB boxes (#653's report) clear it comfortably.

## Scope and known limitations (deliberate, flagged for maintainers)

- **Build-id expiry**: floors are exact-match by catalog id. A hot-swapped build id (`gpt-oss-20b--r2`-style) misses both tables **in lockstep** (no cold/warm desync; the tier reverts to the flat default until the new build is measured). Documented in both tables; the provider logs the resolved floor + serving set at startup so operators can see it. A family mapping was deliberately not added — unmeasured means unmeasured.
- **Release coupling**: `servabilityPerModelFloorMinVersion = "0.8.11"` is a placeholder; the release shipping the provider half must be numbered exactly that (or the constant updated in the release commit). Note `CompareVersions` parses non-numeric segments as 0, so the shipped version constant must not carry a `-swift.N`-style suffix or the coordinator will charge the flat floor against a per-model binary (the unsanctioned tighter direction).
- **StandaloneServer (`--local` mode) keeps the flat default throughout** (its served set is operator-mutable at runtime via `setModels`, and it is not the fleet path); `EngineV2Factory+Production`'s paged pool sizing also keeps the flat probe default — conservative direction (smaller pool or contiguous fallback, never an OOM).

## Before / After

```mermaid
flowchart TB
  subgraph Before
    A1[request for gpt-oss-20b] --> B1["load gate: needs = weights 13.5 + FLAT 5.5 + 1.0 = 20.0 GB"]
    B1 --> C1{"usable ≥ 20.0?<br/>(24 GB box: max ~18)"}
    C1 -- never --> D1[429 capacity unavailable<br/>box online, serves nothing]
  end
  subgraph After
    A2[request for gpt-oss-20b] --> B2["load gate: needs = weights 13.5 + measured floor 3.5 + 1.0 = 18.0 GB"]
    B2 --> C2{"usable ≥ 18.0?"}
    C2 -- yes --> E2["model loads; post-load probe, KV grants,<br/>and runtime gate all carve the SAME 3.5"]
    C2 -- no --> D2[429 — genuinely out of memory]
  end
```

```mermaid
flowchart TB
  subgraph Code_Before["Code (before)"]
    F1["UnifiedMemoryCap.resolvedActivationReserveBytes()<br/>= max(env, flat 5.5)"] --> G1["loadHeadroomBytes()"]
    F1 --> H1["GlobalKVCacheBudget (nil → flat)"]
    F1 --> P1["KVHeadroomProbe (flat)"]
    F1 --> Q1["fleetKVBudgetBytes / EngineV2KVSizing (flat)"]
    G1 --> I1["ProviderLoop static loadHeadroomGb<br/>(baked at first use)"]
    J1["servabilityActivationFloorForVersion(version)<br/>3.0 | 5.5"] --> K1[coldTokenBudgetEstimate]
  end
  subgraph Code_After["Code (after)"]
    F2["measuredActivationFloorsBytes table<br/>+ activationFloorBytes(forModelIDs:)"] --> G2["resolvedActivationReserveBytes(modelIDs:)<br/>= max(env, set floor)"]
    G2 --> H2["ProviderLoop computed loadHeadroomGb /<br/>resolvedActivationReserveBytes<br/>(live advertised ∪ resident)"]
    H2 --> I2["GlobalKVCacheBudget.setActivationReserveBytes<br/>(pre-insert on prefetch; pre-regrow on unload)"]
    H2 --> P2["KVHeadroomProbe(activationReserveBytes:)<br/>post-load + post-bridge guards"]
    H2 --> Q2["fleetKVBudgetBytes / EngineV2KVSizing /<br/>FleetKVContext → grants + heartbeat clamp"]
    G2 --> L2["ModelFitDiagnostic.diagnose(servingSetIDs:)<br/>doctor verdict = daemon's floor;<br/>alternatives keep solo floors"]
    J2["servabilityActivationFloor(version, modelID)<br/>3.0 | 5.5 | table[model] ?? 5.5"] --> K2[coldTokenBudgetEstimate + snap.model]
  end
```

## Test plan

- `provider-swift`: `swift test` — **2162 tests / 223 suites green.** New tests pin: the floor table and max-over-set semantics, raise-only env against the set floor, `loadHeadroomBytes(modelIDs:)`, the doctor's 20.0→18.0 arithmetic and the 24 GB repro verdict flipping FAIL→PASS, the multi-model false-PASS regression (serving-set floor), per-alternative floors in the fix suggestion, and `GlobalKVCacheBudget.setActivationReserveBytes` flipping admission. One pre-existing re-slice wiring test's regrow expectation was updated to model the new relax-on-unload behavior (survivor regrows against the relaxed floor).
- `coordinator`: `go test ./...` — all packages green, `gofmt` clean. New tests pin all three version regimes of `servabilityActivationFloor` and the per-model cold-estimate goldens (109223 vs 103854 on the canonical roomy node); existing goldens unchanged (pre-per-model versions resolve identically for any model).
- Live repro data (24 GB M4 Pro, 0.8.10): doctor `needs ~20.0 / usable 14.5–16.1`; local-endpoint completion → HTTP 429 in 13 ms with trust online; `models list` EST MEM 13.5 for the same model. Logs uploaded via `darkbloom report` (2026-08-23).

Closes #653. Related: #675 (the surfaces-disagree side — `status` "Inference memory" is untouched here).

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/683"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790055755&installation_model_id=21733&pr_number=683&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F683&signature=1bb65cabb8581ba7b51a8c9565fdaad5f701921c53c04b8450af205cb6587b1a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/coordinator/registry/scheduler.go
+++ b/coordinator/registry/scheduler.go
@@ -104,7 +104,7 @@ type routingSnapshot struct {
 	// binaryVersion is the provider's reported binary version (p.Version, read
 	// under p.mu at snapshot time; empty = unreported/legacy). Feeds the
 	// version-gated activation-reserve selection in the cold servability
-	// estimate (servabilityActivationFloorForVersion) so a mixed-version fleet
+	// estimate (servabilityActivationFloor) so a mixed-version fleet
 	// is charged the reserve each binary actually holds.
 	binaryVersion    string
 	slotState        string
--- a/coordinator/registry/servability.go
+++ b/coordinator/registry/servability.go
@@ -44,11 +44,12 @@ const (
 	// servabilityCapFraction mirrors the provider's UnifiedMemoryCap default
 	// (90% of physical memory usable for MLX).
 	servabilityCapFraction = 0.90
-	// servabilityActivationFloorGB mirrors the provider's activation reserve
-	// (UnifiedMemoryCap.defaultActivationReserveBytes, 5.5 GiB): the working
-	// set held back on top of weights before any KV cache. It is FLAT on the
-	// provider — every model, every attention posture, every batch — so it is
-	// flat here. It does not scale with anything.
+	// servabilityActivationFloorGB mirrors the provider's DEFAULT activation
+	// reserve (UnifiedMemoryCap.defaultActivationReserveBytes, 5.5 GiB): the
+	// working set held back on top of weights before any KV cache. On
+	// pre-per-model binaries it is FLAT — every model, every attention
+	// posture, every batch; on per-model binaries it is the fallback for
+	// models without a measured floor (servabilityModelActivationFloorsGB).
 	//
 	// 5.5 as of v0.8.0, moved in the SAME commit as the provider constant:
 	// the release ships decode batch 8 and the measured gemma-4 B=8
@@ -63,8 +64,11 @@ const (
 	// removed: the provider half it claimed to mirror was never wired, so the
 	// coordinator was charging for a reserve no provider ever held. See
 	// coldTokenBudgetEstimate's "Mirroring, not modelling" note before adding
-	// any term here. Moving the FLAT floor — what this change does — is the
-	// sanctioned shape; re-adding a model-dependent term is not.
+	// any term here. That note still governs: the per-model floors below are
+	// NOT a second opinion about prefill memory — they mirror the measured
+	// floors the provider's own UnifiedMemoryCap now holds, exactly as this
+	// flat constant mirrors its default. A term the provider does not hold
+	// remains unsanctioned.
 	servabilityActivationFloorGB = 5.5
 	// servabilityLegacyActivationFloorGB is the reserve a pre-0.8.0 provider
 	// actually holds (the old defaultActivationReserveBytes). During the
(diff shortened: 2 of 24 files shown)


Gajesh2007:
hey @jkjoplin ! can you sign your commits?

Gajesh2007:
@codex review this code

Gajesh2007:
This landed (with your authorship preserved on the cherry-picked commit — thank you for the excellent groundwork!) as part of #791, together with the 0.8.16 version-gate fix, measured qwen3.5/3.6 floors, measured resident weights, and several rounds of review hardening on the serving-set reserve machinery. Closing in favor of #791.

PR #728, opened 2026-08-25
Ground earnings calculator in bandwidth-limited capacity

0xkydo:
## Summary

Reworks the console and homepage earnings calculators around one transparent capacity estimate, without changing the coordinator or adding an endpoint.

- requires Mac model, chip family, and unified memory before showing an estimate;
- pins the three currently supported models, their active parameters, model size, memory floor, and output price;
- finds the highest-earning supported model that fits memory;
- estimates single-stream decode from 65% of chip memory bandwidth and active model weights;
- applies an adjustable duty cycle, defaulting to 5%, with no batching;
- keeps the estimate caveat prominent and places provider setup directly below the result;
- limits current enrollment to 48 GB+ Macs while inviting owners of smaller Macs to register interest;
- adds M5 Ultra and M6 hardware profiles and removes MacBook Air, MacBook Neo, and iMac options.

| Model | Active parameters | Weights | Output price / 1M tokens |
| --- | ---: | ---: | ---: |
| Qwen3.6 35B A3B | 3.0B | 22 GB | $0.700 |
| Gemma 4 26B A4B | 4.0B | 17 GB | $0.220 |
| GPT-OSS 20B | 3.6B | 12 GB | $0.069 |

## Problem

The previous calculator mixed several theoretical and payout assumptions into the prominent number. Hardware selection was ambiguous, low-memory Macs could appear production-ready, and the disclaimer was easy to crop out of a screenshot.

During local verification, the revised console still showed **Estimate unavailable** for a valid M5 Max / 48 GB selection. The calculator never reached its formula because model discovery failed first.

## Root cause

The calculator treated `/api/models` as a required runtime dependency, then tried to infer total parameters, active parameters, quantization, memory requirements, and pricing from catalog strings. A failed request—or a catalog response without the calculator-specific fields—emptied the model list and forced a terminal unavailable state.

The invariant for this iteration is simpler: the estimate must be available whenever the user selects supported hardware, because the supported calculator model set and its pricing are known and intentionally pinned.

## Before

```mermaid
flowchart LR
  subgraph Behavior
    A1[Select valid Mac hardware] --> B1[Fetch model catalog]
    B1 --> C1{Request and inferred fields valid?}
    C1 -- No --> D1[Estimate unavailable]
    C1 -- Yes --> E1[Calculate earning]
  end
  subgraph Code
    F1[/api/models] --> G1[Client fetch effect]
    G1 --> H1[Parse model names and pricing]
    H1 --> I1[Stored catalog state]
    I1 --> J1[Console calculator]
    F1 --> K1[Separate homepage fallback]
  end
```

## After

```mermaid
flowchart LR
  subgraph Behavior
    A2[Select Mac model + chip + memory] --> B2{48 GB or more?}
    B2 -- No --> C2[Welcoming interest-registration state]
    B2 -- Yes --> D2[Find fitting pinned model]
    D2 --> E2[65% bandwidth / active weights]
    E2 --> F2[Apply duty cycle, default 5%]
    F2 --> G2[Apply pinned output price]
    G2 --> H2[Monthly earning + screenshot-visible caveat]
  end
  subgraph Code
    I2[Three-model constant] --> J2[Focused capacity formula]
    J2 --> K2[Derived console model rows]
    J2 --> L2[Homepage calculator core]
    K2 --> M2[Result + readiness + calculation UI]
    L2 --> M2
  end
```

## Implementation

- Replaces the old reward/share calculation with the requested six-step formula: fitting model, chip bandwidth, 65% bandwidth efficiency, duty cycle, no batching, and output-token pricing.
- Defines Qwen, Gemma, and GPT-OSS inputs directly in both calculator cores, with parity tests pinning every value.
- Removes the console fetch effect, loading/error catalog state, model-name parsing, and homepage catalog fetch.
- Derives ranked model rows directly from selected hardware and the static model list, avoiding redundant client state and an extra render cycle.
- Requires explicit sequential hardware selection and shows no result until all three inputs are set.
- Adds a collapsible calculation explanation and a prominent black-text earning caveat.
- Shows a welcoming interest state below 48 GB instead of a production enrollment path.
- Makes no coordinator, store, registry, schema, or route changes.

## Before-merge follow-up: interest destination

- [ ] Wire **Register your interest** to the canonical, durable registration destination on both surfaces.
  - The homepage currently opens `https://console.darkbloom.dev/earn`.
  - The console currently records analytics/local storage and invokes sign-in.
  - Preserve the selected Mac details and only show the email-notification confirmation after registration succeeds.

This PR intentionally does not add an endpoint for that follow-up.

## Alternatives considered

- **Keep fetching the existing model catalog and add more parser fallbacks.** This would still make a fixed three-model calculator dependent on a runtime request and on display-name conventions that are not calculator contracts.
- **Add a calculator-specific coordinator endpoint.** It would provide canonical data but expands backend scope for three stable records; the user explicitly preferred hardcoding for now.
- **Use the full bandwidth limit.** Too optimistic for the prominent estimate; the disclosed 65% factor is more conservative.

## Verification

- `cd console-ui && npm test -- --run` — 57 files, 500 tests passed.
- `cd console-ui && npm run build` — production build passed.
- Repository pre-push TypeScript lint and Next.js build checks — passed.
- `node --test landing/earn-calculator-core.test.js` — 5 tests passed.
- `node --check landing/earn-calculator.js` and `node --check landing/earn-calculator-core.js` — passed.
- `git diff --check` — passed.
- Local console walkthrough: M5 Max (40-core GPU), 48 GB, 5% duty cycle shows Qwen $19.20/mo, Gemma $4.35/mo, and GPT-OSS $1.65/mo.
- Local homepage walkthrough with the same inputs produces the same $19.20/mo primary estimate.

## Rollout

Frontend-only. The calculators no longer depend on model-catalog availability, so supported selections calculate immediately. Updating a supported model's price or active-weight specification requires changing the two mirrored constants and their parity test. No provider protocol or production coordinator behavior changes.


0xkydo:
Before merge, the **Register your interest** action needs to lead to the canonical registration destination and persist the submission on both surfaces. The homepage currently opens `/earn`; the console currently records analytics/local storage and invokes sign-in. Please verify the selected Mac details carry through and only show the email-notification confirmation after durable registration succeeds.

0xkydo:
Simplified this PR to a single frontend-only commit on current master: 17 console/landing files, no coordinator changes, and no new endpoint. The calculators use the existing model catalog plus a temporary hardcoded Qwen output price. The earlier **Register your interest** follow-up comment still applies to both surfaces before merge.

Gajesh2007 (dismissed):

Gajesh2007 (approved):

PR #803, opened 2026-09-01
feat(receipts): bind response attestation to model weights and request bytes

### LINKED ISSUE #802: Response attestation binds neither model weights nor request bytes
### Problem

The response attestation signs `sha256(requestId:completionTokens:responseBody)`. That hash binds neither the model weights, nor the request bytes, nor any sampling parameter. A provider can serve one model at another model's price, or quietly alter parameters, and every attestation still verifies. The privacy chain (NaCl Box, Secure Enclave attestation, MDM, Hardened Runtime) is complete; correctness of what actually ran is the one gap left open.

### Proposal

Upgrade the attestation payload without changing the signing channel. The provider seals each completed inference into a canonical receipt:

```json
{"completion_tokens":42,"model_id":"…","model_weight_hash":"<aggregate SHA-256 of served weights>",
 "prompt_tokens":17,"request_id":"…","request_sha256":"<SHA-256 of exact decrypted request bytes>",
 "response_sha256":"<SHA-256 of full response text>","v":2}
```

`response_hash` becomes the receipt's SHA-256 hash and `se_signature` keeps its existing meaning, a Secure Enclave signature over `response_hash`. One new optional `receipt` field on `inference_complete` carries the record. Older providers and consumers are unaffected.

Three properties fall out:

1. The request digest covers the exact decrypted body bytes, so it binds every sampling parameter with no canonicalization machinery.
2. `model_weight_hash` reuses the aggregate hash the provider already computes and registration already cross checks against the model catalog.
3. Receipts hold digests only, never prompt or response content, so the coordinator can serve receipt and verdict at a public `GET /v1/receipts/{hash}`.

A dishonest provider can still sign a well formed receipt over a lie. The coordinator refutes it by cross checking the digest of the body it sealed and the weight hash the provider registered. That scenario is integration tested.

### Reference implementation

A working implementation is attached as a companion PR: mirrored Go and Swift encoders pinned by shared golden vectors, full coordinator suite green on Linux, receipt and protocol tests green on Apple Silicon. Per CONTRIBUTING.md this issue is the scoping venue; naming, wire format, and scope are all open to maintainer guidance.

humuhumu33:
Inference receipts: seal every completed inference into a canonical record binding the model weight hash, the exact decrypted request bytes, and the response digest, signed through the existing attestation channel. `response_hash` becomes the receipt's SHA-256 hash, `se_signature` keeps its existing meaning, and one new optional `receipt` field carries the record. The coordinator cross checks each receipt and serves it with a verdict at public `GET /v1/receipts/{hash}`.

Closes #802

## Why

- `provider-swift/.../SecurityHardening.swift` signed `sha256(requestId:completionTokens:responseBody)`. That hash binds neither the model weights, nor the request bytes, nor any sampling parameter. A provider serving one model at another model's price produces valid attestations today.
- Everything needed to close the gap already exists: providers compute an aggregate weight hash that registration cross checks against the model catalog (`registry.go`), the Secure Enclave signing channel already rides `inference_complete`, and the coordinator holds the exact plaintext it seals for each provider at both encrypt sites.
- Receipts hold digests only, never prompt or response content, so the verdict endpoint can be public without touching the privacy model.

## Behavior: before and after

```mermaid
flowchart LR
  subgraph Before
    A1[consumer request] --> B1[provider runs a model]
    B1 --> C1["se_signature over sha256(requestId:tokens:text)"]
    C1 --> D1[model substitution and parameter tampering undetectable]
  end
  subgraph After
    A2[consumer request] --> B2[provider runs model]
    B2 --> C2["receipt binds weight hash, request digest, response digest; response_hash = receipt hash, SE signed"]
    C2 --> D2["coordinator cross checks 4 bindings; public GET /v1/receipts/{hash}"]
  end
```

Unchanged by design: the signing contract (`se_signature` over `response_hash`) is byte compatible; providers without receipts keep today's path exactly; consumers see the same fields they always saw plus one optional `receipt`; a failed binding never fails the request (logged, counted, visible on the record); no billing, routing, or trust level behavior changes.

## Code: before and after

```mermaid
flowchart LR
  subgraph Before
    S1["InferenceHandler"] --> A1b["computeResponseAttestation\nsha256(requestId:tokens:text)"]
    A1b --> W1["inference_complete: se_signature, response_hash"]
    W1 --> P1["provider.go: copy to pending request"]
    P1 --> C1b["consumer response passthrough"]
  end
  subgraph After
    S2["InferenceHandler\nsha256(decrypted body) + registered weight hash"] --> A2b["InferenceReceipt.swift\ncanonical JSON, hash, SE sign"]
    A2b --> W2["inference_complete: + receipt"]
    W2 --> P2["provider.go: recordReceipt"]
    P2 --> V2["receipt.Verify: hash, dispatched body digest,\nweight hash, signature (coordinator/receipt)"]
    V2 --> E2["receiptCache: GET /v1/receipts/{hash}"]
    P2 --> C2b["consumer response: receipt + response_hash + se_signature"]
  end
```

## Protocol (Go and Swift mirrored)

- `inference_complete` gains one optional field, `receipt` (string): mirrored in `coordinator/protocol/messages.go` and `provider-swift/.../Protocol/Messages.swift` plus the coordinator client codec. Legacy wire shape is unchanged when absent; older coordinators ignore it.
- Canonical receipt form: single line JSON, alphabetical keys, minimal escaping. `receipt.Parse` requires that re encoding reproduces the wire bytes, so each receipt has exactly one byte representation and re serialization forgery is closed. Go and Swift encoders are pinned byte for byte by `fixtures/receipts/receipt_vectors.json`, generated independently of both implementations.
- New public endpoint `GET /v1/receipts/{hash}`: digests, signature, attested public key, and the coordinator's per binding verdict. Checks are paired with `*_checked` flags so sealed sender mode and unattested providers degrade honestly instead of passing silently.
- No release artifact changes, no version bump, no migration.

## Verification

- Coordinator: `go test ./coordinator/...` green on Linux, 25 of 25 packages including the full `api` suite. `go vet` clean.
- `coordinator/receipt`: golden vectors, canonical form strictness, and a tamper matrix (every field mutation, wrong key, malformed signatures, tri state checks).
- `TestIntegration_ReceiptGenuine`: real httptest plus WebSocket loop with a simulated provider signing with its attested P-256 key. The receipt reaches the consumer stream, `response_hash` equals the receipt hash, a third party verifies the signature from public endpoint data alone, and every coordinator check is green.
- `TestIntegration_ReceiptLyingProviderIsFlagged`: a canonically encoded, correctly signed receipt claiming a different request digest and weight hash. `address_match` and `signature_valid` stay true, which is the attack, and the dispatched body digest and registered weight hash cross checks refute it.
- Provider: `swift build --build-tests` compiles the full package including every test target (Swift 6.3, Apple Silicon). Receipt twin tests green: golden vector byte identity with Go, hash movement on any field change, escaping parity, and the signer contract. Protocol and codec round trips carrying the new field green.
- Not run here: the MLX and Metal backed Swift suites (compile verified only; no metallib on the runner used). This repo's own CI covers them.

## Notes for reviewers

- The v1 `computeResponseAttestation` is removed; its signing contract survives unchanged inside `computeReceiptAttestation`.
- The receipt cache is in memory and bounded (`receiptCacheCap`). The receipt travels inline to the consumer, so the endpoint is a lookup convenience, not a custody layer. Durable storage is an easy follow up if wanted.
- On speculative backup dispatch, `DispatchedBodySHA256` follows the same last writer pattern as `SessionPrivKey`; if the two sealed bodies ever differ per provider, a false digest mismatch flag is possible (advisory only, never request failing).
- Follow ups sketched in `docs/architecture/receipts.md` and deliberately out of scope: console receipt display, replay audits on top of direct mode, and the determinism measurement replay verdicts depend on.
- Scope, naming, and wire format are open to maintainer guidance in #802.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## Unreleased
+
+- **Inference receipts** — Every completed inference seals a canonical
+  receipt binding the model (aggregate weight hash), the exact decrypted
+  request bytes (and therefore every sampling parameter), and the response
+  digest. `response_hash` becomes the receipt's SHA-256 hash; `se_signature`
+  keeps its existing contract (Secure Enclave signature over
+  `response_hash`); one new optional `receipt` field carries the record, so
+  older providers and consumers are unaffected. The coordinator checks the
+  receipt against the body it sealed, the provider's registered weight hash,
+  and the attested key, then serves receipt and verdict at public
+  `GET /v1/receipts/{hash}` (digests only, never prompt or response
+  content). A well formed, correctly signed receipt over a lie is refuted by
+  those cross checks (integration tested). Canonical encoding is byte
+  identical across Go (`coordinator/receipt`) and Swift
+  (`ProviderCore/Security/InferenceReceipt.swift`), pinned by independently
+  generated golden vectors (`fixtures/receipts/receipt_vectors.json`).
+  Spec: `docs/architecture/receipts.md`.
+
 ## Release candidate v0.8.15 (not shipped; 2026-08-28)
 
 - **Exact Qwen3.8 dense VLM artifact** — Providers serve
--- a/coordinator/api/consumer.go
+++ b/coordinator/api/consumer.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/eigeninference/d-inference/coordinator/auth"
 	"github.com/eigeninference/d-inference/coordinator/internal/e2e"
+	"github.com/eigeninference/d-inference/coordinator/receipt"
 	"github.com/eigeninference/d-inference/coordinator/payments"
 	"github.com/eigeninference/d-inference/coordinator/promptcontract"
 	"github.com/eigeninference/d-inference/coordinator/protocol"
@@ -1046,6 +1047,9 @@ func (s *Server) dispatchOneProvider(
 		cleanupPending()
 		return nil, nil, decision, "failed to encrypt request", http.StatusInternalServerError
 	}
+	// Receipt binding: the digest of the exact plaintext bytes this
+	// provider will decrypt; its receipt's request_sha256 must match.
+	pr.DispatchedBodySHA256 = receipt.SHA256Hex(sealedBody)
 	if pr.Timing != nil {
 		pr.Timing.EncryptedAt = time.Now()
 	}
@@ -2225,12 +2229,15 @@ func (s *Server) handleStreamingResponseWithFirstChunkAndError(
 							pendingUsage["se_signature"] = pr.SESignature
 							pendingUsage["response_hash"] = pr.ResponseHash
 						}
+						if pr.Receipt != "" {
+							pendingUsage["receipt"] = pr.Receipt
+						}
 						attachChatCompletionMetadata(pendingUsage, pr)
 						if out := finalizeUsageChunk(pendingUsage, usage, pr); out != "" {
 							fmt.Fprintf(w, "%s\n\n", out)
 							flusher.Flush()
 						}
-					} else if pr.SESignature != "" || hasChatCompletionMetadata(pr) {
+					} else if pr.SESignature != "" || pr.Receipt != "" || hasChatCompletionMetadata(pr) {
 						// No held usage chunk to ride on: emit the signature and/or
 						// opt-in metadata as a fully-shaped chat.completion.chunk
 						// (id/object/created/model/choices) so strict decoders parse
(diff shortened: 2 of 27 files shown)


humuhumu33:
Rebased with a wording pass before review starts: renamed the Go package to coordinator/receipt, retitled the doc to docs/architecture/receipts.md, and rewrote comments in repo terms. No behavior change; full coordinator suite and the provider receipt tests re-verified green.

PR #804, opened 2026-09-01
feat(sandbox): finish the macOS sandbox — guest execution, egress isolation, billing

owizdom:
> **This is the single PR that continues #672 to a working sandbox.** Each commit is one self-contained step, ordered so it reads top to bottom. The checklist below is the live progress map — it gets ticked as commits land, so at any point it shows exactly what is done and what is left.
>

## Where #672 left off

The substrate is genuinely strong: a 17-state lifecycle with compare-and-swap transitions, monotonic fencing tokens, crash-consistent cancellation, scoped idempotency with CBOR request commitments, a dedicated host protocol, and 4 tables with careful partial indices.

What it cannot do is run a single line of tenant code. `LumeGuestCommandPolicy` has only `.disabled` and `.baseImagePreparationAndDevelopment`, the public `LumeRuntimeConfiguration` initializer hardcodes `.disabled`, and `ServeCommand` uses exactly that initializer — so `admitExecute` reaches the guard in `LumeVirtualMachineRuntime+GuestCommands.swift` and throws `unsupported`.

That is the correct call, not an oversight: Lume's unattended guest boots with the shared, static `lume`/`lume` credentials, and that identity can rewrite its own launchd metadata and SSH startup files. Multi-tenancy on that footing would be theatre.

So the work is to replace that identity with a real one, and this PR does it in order.

---

### 1. Guest vsock transport — landed

Only the process holding the `VZVirtioSocketDevice` can open a guest connection, so the daemon cannot do it itself. Patch `0005-guest-vsock-channel` attaches the device in Lume's macOS path, connects to the agent's port on the machine's own dispatch queue, and hands the descriptor to the daemon over `SCM_RIGHTS`. `lume run` then leaves the data path entirely — no tenant bytes are read, written or buffered by the patched third-party binary.

Three decisions worth your review:

- **The device is opt-in.** Attached only when a broker supplies a port, so an ordinary `lume run` keeps exactly the device set it had before.
- **The handoff does not gate boot.** It retries in the background and only logs on failure. Gating startup would either hang a healthy VM or fail one whose agent has simply not bound yet.
- **A second channel, not a reused one.** Child descriptor 4, beside the lifecycle channel at 3. The lifecycle protocol is EOF-only and treats any byte as a fault, so it structurally cannot carry a descriptor.

```mermaid
flowchart TB
  subgraph Before
    direction TB
    B1[daemon spawns lume run] --> B2[VM boots]
    B2 --> B3["only route in: lume ssh<br/>shared lume/lume identity"]
    B3 --> B4["guest commands disabled<br/>execute throws unsupported"]
    C1["ProcessExecution<br/>one channel, child fd 3"] --> C2["createConfiguration<br/>socketDevices never set"]
    C2 --> C3["VM.run<br/>no guest channel"]
  end
  subgraph After
    direction TB
    A1["daemon spawns lume run<br/>+ guest channel fd + port"] --> A2["VM boots with<br/>VZVirtioSocketDevice"]
    A2 --> A3["lume connects to agent port<br/>retried, never a boot gate"]
    A3 --> A4["SCM_RIGHTS handoff"]
    A4 --> A5["daemon owns the fd,<br/>lume leaves the data path"]
    A5 --> A6["guest commands still disabled"]
    D1["ProcessExecution<br/>second channel, child fd 4"] --> D2["createConfiguration<br/>socketDevices when port set"]
    D2 --> D3["connectGuestPort<br/>on the machine queue"]
    D3 --> D4["GuestChannelCapability.send<br/>sendmsg + SCM_RIGHTS"]
    D4 --> D5["GuestChannelReceiver<br/>recvmsg, non-blocking"]
  end
  Before ~~~ After
```

Verified: `0001`–`0005` apply clean to a fresh archive of `737dc2a0` and reproduce the tree exactly; patched Lume builds and its **full suite passes 163 tests, 0 failures** (159 baseline plus 4 new); `build-pinned-lume.sh` with `DARKBLOOM_LUME_RUN_TESTS=1` exits 0 with all 10 required tests gated individually; the **descriptor handoff passes 10/10** through the real `SandboxProcessRunner`, proving env injection, child fd 4, `POSIX_SPAWN_CLOEXEC_DEFAULT` survival, `SCM_RIGHTS` transfer and the receiver.

The suites covering this branch's code are written with **swift-testing**, so **38 of them run locally** under Command Line Tools — 22 for the executor, 16 for the channel client, transport seam and readiness. The older XCTest suites still need Xcode and run in CI.

### 2. Guest wire protocol — landed

The channel is a raw byte stream, so both ends need agreed framing, and the host has to treat everything from a guest as untrusted.

`SandboxGuestProtocol` is a dependency-free module both sides link: one byte of kind, four bytes of big-endian length, a closed frame vocabulary, and a payload cap **enforced from the header alone** — an oversized frame is refused before a single payload byte is buffered. Partial reads decode to `nil` and leave the buffer untouched, so a streaming reader can just append.

The result envelope is deliberately **byte-identical** to the one the launchd bootstrap path already produces. `LumeGuestCommandJournal` persists raw envelope bytes and re-validates them through `LumeGuestCommandResultDecoder` on both write and read, so a different envelope would stop stored results replaying. The transport moves; the format does not. `GuestProtocolWireCompatibilityTests` decodes agent envelopes through the real host decoder and pins magic, schema version and the stream cap to the host's own constants, so the two cannot drift silently.

The agent announces a versioned identity and answers framed requests. **It still does not execute** — a well-formed command gets an `execution_disabled` failure, because tenant work also needs randomized bootstrap credentials and an unprivileged sandbox user, and executing before both exist would be exactly the boundary the host is refusing to cross.

### 3. Host client, and protocol split from the socket family — landed

`SandboxGuestChannelClient` adopts the descriptor handed over by the run process and speaks the framed protocol across it. Everything arriving is untrusted: reads are deadline-bounded through `poll`, the inbound buffer is capped independently of the codec's per-frame cap, agent-to-host-only frame kinds are refused, and a handshake that does not validate is a failure rather than a warning.

Results come back as **raw envelope bytes**, not decoded. The client never becomes a second source of truth for the format — the caller persists those bytes in the journal and decodes them with the existing host decoder.

`SandboxGuestAgentCore` now holds all of the agent's protocol handling and knows nothing about vsock; the executable is only the front door. That split means the **real client and the real agent session are exercised against each other** over a socketpair, rather than either half being tested against a stub. `AF_VSOCK` is the one thing substituted, because it cannot be created on a host.

Verified 7/7: handshake identity, mismatched image id rejected, execution refused while disabled, agent-side re-validation of a request the host would never send, and a client that times out instead of hanging.

### 4. Guest command execution — landed

The agent now runs the command and returns the same envelope the launchd path produces. The environment policy is copied from `LumeGuestLaunchDefinition` deliberately — a command's observable environment must not depend on which transport carried it — and reserved keys are **refused** rather than silently overridden, since the host rejects them at construction.

Two review findings drove the design, both reproduced before fixing:

- **Streams and process must be supervised together.** A command that closes its own stdout and stderr reaches EOF immediately, so a stream-only loop finished at once and an unbounded `waitpid` then ran as long as the command liked — measured **8.01s against a 2s deadline, reported as a clean success**. A session serves one command at a time, so any untrusted command could wedge the whole agent.
- **A fast exit with a background child is not a timeout.** A grandchild holding the pipe kept the loop running to the deadline, so `echo done; sleep 30 & exit 7` came back as a 124 timeout with the real exit code discarded — measured **4.00s / exit 124, now 0.31s / exit 7**.

`SandboxGuestCommandSupervisor` reaps with `WNOHANG` inside the loop and gives the streams a short grace once the command itself has exited.

The guest validator was also **strictly weaker than the host's**, despite existing precisely because the channel cannot be trusted. It now enforces the same reserved-key denylist, key charset, key length and aggregate cap, so a peer cannot reach `posix_spawn` with a request the host would have refused.

Verified: 24 execution checks over the real client against the real agent session, both regressions pinned by tests, and the 7 conversation / 23 protocol checks and descriptor handoff all still passing.

**Execution remains disabled in production** — `executionEnabled: false` until Stage 3 replaces the shared `lume` identity.

### 5. Transport seam, and readiness over the channel — landed

There was no seam: `runGuestSSH` was a private static method and the `lume ssh` argv was written out in two places. The fixtures substitute a fake `lume` **binary on disk** rather than a Swift type, so the seam is **additive** — SSH stays the default and its behaviour is unchanged.

`LumeGuestCommandTransport` is prepare-then-deliver rather than one call, and the split is load-bearing: `execute` claims the command in the durable journal **between** the two steps, so a request that cannot be encoded fails before the claim rather than leaving an unresolvable one behind.

Both transports return the **same raw envelope bytes**, so the journal cannot tell which one ran. A test decodes vsock-produced bytes through the real host decoder to keep that true.

Cancellation deliberately stays on the bootstrap path — it boots out a launchd job by label, which is inherently transport-specific, and the agent protocol has no cancel frame yet.

`LumeGuestChannelReadinessProbe` treats a completed handshake as readiness. That proves more than the bootstrap probe with less work, and it removes a dependency that matters later: the bootstrap probe is gated on Lume's `sshAvailable`, which comes from **scraping DHCP leases and the ARP table**, and the packet gateway replaces exactly that networking. Readiness had to move off guest IP discovery before the gateway can land.

Behaviour preservation checked explicitly and pinned by tests: same argv and order, same guest+5 Lume budget and guest+10 host budget, same envelope cap, and the same `commandFailed(command: "lume ssh", …)` name existing failure tests assert verbatim.

---

### 6. Per-sandbox guest identity — landed

Lume's unattended install produced one administrator with the fixed credential `lume`/`lume`, shared by every guest on every host. That account can rewrite its own launchd metadata and SSH startup files, so a tenant command running as it could tamper with the control path supervising it. That is the whole reason `LumeGuestCommandPolicy` is gated shut.

Lume patch **0006** splits the single account into a bootstrap administrator, named and passworded per sandbox, and a separate unprivileged tenant account that joins no groups, carries no login credential, and is never the autologin user. The password travels in the environment on every path, never in argv, because arguments are visible to every process on the host through `ps`.

The single hardcode was one `GuestUser` literal, but the plumbing around it was single-account by construction: uid availability, group membership, autologin, `kcpassword`, power settings and the finalization script all assumed one user named `lume`. `diskutil apfs updatePreboot /` already publishes offline-created accounts and now covers both.

Verified on a real guest: `admin` and `staff` group records both read `root <bootstrap>`, so the tenant is in neither, and its `20(staff)` in `id` is the implicit primary-gid membership every macOS local user has rather than a group write. The tenant has no `AuthenticationAuthority` at all. The old `lume`/`lume` is rejected.

---

### 7. Sandbox pricing — landed

`coordinator/payments/sandboxpricing` reproduces the published rate card exactly, which is the strongest check available: if the implementation and the rate card ever disagree, one of them is wrong and a customer is quoted a number nobody agreed to.

Base rates match E2B's published per-second CPU and RAM benchmark so a Linux sandbox is directly comparable. macOS carries a premium because the supply is genuinely constrained by Apple's licence. The premium is a separate line item rather than a collapsed effective rate, because chip scarcity and computer-use capacity move independently and a stored total cannot be decomposed later. Money is integer micro-USD end to end, rounded once and always toward the customer, with no hidden compute floor.

---

### 8. Base-image preparation — landed

Everything above was unprovable against a real guest, because preparing a base image died with exit 70 before the offline setup patcher ran. A control run carrying no identity flags failed identically, which puts the defect in patch **0003** rather than in anything this PR added.

`UnattendedInstaller` cancels the run task and calls `vm.stop()` back to back. Both are MainActor, so the stop reaches the virtualization service first and suspends; the cancelled task then wakes into `VM.run`'s emergency path, sees `.stopping`, and issues a second stop. `VZVirtualMachine.stop` is valid only from Running or Paused, so the framework refuses it, the machine is not yet `.stopped`, and the owner process kills itself over a guest that was already on its way down. `VM.stop()` has always gated on Running/Paused; only this path did not.

```mermaid
flowchart TB
  subgraph Before["Before: exit 70, patcher never runs"]
    A1["run task cancelled<br/>CancellationError"] --> B1{"state != .stopped?"}
    B1 -->|"yes, it is .stopping"| C1["service.stop()"]
    C1 --> D1["VZ refuses:<br/>invalid state transition"]
    D1 --> E1{"state == .stopped?"}
    E1 -->|"no, still .stopping"| F1["_exit(70)"]
  end
  subgraph After["After: waits for terminal"]
    A2["run task cancelled<br/>CancellationError"] --> B2{"terminal?<br/>.stopped or .error"}
    B2 -->|yes| G2["cleanup, patcher runs"]
    B2 -->|no| C2{"Running or Paused?"}
    C2 -->|yes| D2["service.stop()"]
    C2 -->|"no, stop already in flight"| E2
    D2 --> E2["bounded wait for terminal"]
    E2 -->|reached| G2
    E2 -->|"expired, still live"| F2["fail-stop<br/>property preserved"]
  end
```

Lume patch **0007** gates the emergency stop the same way `VM.stop()` always did and replaces the racy state read with a bounded wait that becomes the single arbiter of whether the guest is stuck alive. A guest still Running or Paused when that wait expires still terminates the process, so the fail-closed property is unchanged. `.error` counts as terminal because `didStopWithError` has already torn the guest down.

The same pattern existed in two more places. In `stopRetainedRunAfterFailStop` it had a second consequence: a retained run whose guest ended in `.error` could never be released and leaked its run locks for the process lifetime. All three now share the terminal-state helpers.

Two things made this much harder to find than it should have been, and both are fixed. `Logger` is buffered `print()` and `_exit` skips the stdio flush, so the captured log of the control run was exactly 8192 bytes, one stdio block, cut mid-word, and the error explaining the failure was written and destroyed; the broker fail-stop did not log at all, leaving its exit 70 indistinguishable from this one. And no test could catch the bug because the mock went `.running` to `.stopped` atomically with no `.stopping` window, never refused an illegal stop, and left `.stopped` where real VZ leaves `.error`. With the mock corrected the suite reproduces the bug by exiting 70.

Three smaller defects surfaced alongside, each found by running a real create rather than by a test. Every built-in unattended preset ships a health check hardcoding `lume`/`lume`, so patch 0006's fallback never fired and a per-sandbox guest was verified against an account it does not have. `materializeFirstBootState` had no retry for the auxiliary storage lock the installer releases asynchronously, though `verifySetup` has retried exactly that since before this work, so on a loaded machine preparation failed outright. And `launchctl bootout` is asynchronous while both call sites checked `launchctl print` immediately afterwards, so a cancellation that had succeeded reported failure whenever the host was loaded enough for the unload to lag; both now wait, bounded.

The pattern across all of them is the same one, which is worth naming: issue an asynchronous teardown or release, then immediately assert it finished. It appeared in the VM stop path three times, in the launchd teardown twice, and in the first-boot start once.

Two adversarial reviews ran over this. Between them they caught a racy state read I had left inside the fail-stop decision (the same bug, moved inward), a 30s wait that outlived the broker's own 15s deadline so the diagnostic would never have printed, a shell bound documented as 4.5s that measures 5.5s and overran the grace it had to fit inside, a missing CI named-test gate, and a regression assertion of mine that could never fail. All fixed. Known and deliberately deferred, none affecting correctness: `.starting` is not re-evaluated for stoppability during the wait (fail-closed, and not a regression); one pre-existing test now waits out the full 10s budget; the fail-stop logs `VZVirtualMachineState(rawValue: 1)` rather than a name; `DarwinVM` did not get the injectable-timeout seam `LinuxVM` got.

The first-boot retry is the one change not observed firing live: the race it handles is real and was hit, but the machine was quiet on the confirming run. It is a line-for-line mirror of the retry `verifySetup` has carried since before this work.

Verified on real hardware, not by inference: a full `lume create --unattended` reaches exit 0, the offline patcher runs, the health check passes on its first attempt as the per-sandbox administrator, and the resulting image has sshd and both accounts.

---

### 9. Guest vsock proven against a real VM — verification, no code

No commits in this section. It closes the one open item under Transport and identity, which
everything after it was being designed on top of.

**It had never been done.** `Tests/SandboxGuestIntegrationTests/GuestChannelIntegrationTests.swift:17`
says so in a comment: *"A socketpair stands in for the vsock connection, because AF_VSOCK can
only be [created inside a VM]"*. Every vsock test to date is a host-side socketpair stand-in,
`LumeGuestChannelReadinessProbe` has only ever been called from tests, and no production path
attaches a channel. So the transport this PR builds on was entirely unexercised.

Probed against a prepared guest (macOS 26.6.2), driving the real patched `lume` and the real
`darkbloom-guest-agent` release build:

| | Result |
|---|---|
| Guest binds `AF_VSOCK` | **PASS** as root — `SOCKET_OK`, `LOCAL_CID=3`, `BIND_OK`, `LISTEN_OK` |
| Patch 0005's bridge | **PASS** — `connectGuestPort` → `SCM_RIGHTS` → host read guest bytes |
| Unprivileged bind | **PASS** as uid 502 `sandbox`, in neither `admin` nor `staff` |
| Real agent handshake | **PASS** — `{"magic":"darkbloom_guest_agent","protocol_version":1,"agent_version":"0.1.0","image_id":"probe-image-42"}` |
| LaunchDaemon at boot | **PASS**, after fixing the defect below |

This is the first time vsock has worked guest-to-host here. It also demonstrates
`DARKBLOOM_GUEST_IMAGE_ID` propagating end to end, so `expectedImageID` — a protocol slot
that is currently unwired on both sides — has a proven path.

**Boot timing: 9 seconds**, cold `lume run` to completed handshake with the agent started by
a LaunchDaemon, against the host's existing 600s retry window (`0005:327`). The margin is
structural rather than lucky: a system-domain LaunchDaemon starts well before the login
window, so the listener is up long before sshd is. That is also the strongest argument yet
for moving readiness off Lume's `sshAvailable` and onto the channel.

**A false negative worth recording.** The first attempt returned `ENODEV` *as root*, which
reads exactly like "macOS guests cannot do vsock". It is not — `lume run` attaches
`VZVirtioSocketDeviceConfiguration` only when `guestChannelPort != nil` (`0005:482-484`), and
a plain run has no device. Read any future `ENODEV` as "no device attached" first.

**A real defect found, for whoever does the bake.** A LaunchDaemon with `UserName = sandbox`
and `StandardOutPath` under `/var/log` fails with `last exit code = 78: EX_CONFIG`, respawns
indefinitely (`runs = 45` and climbing), and writes **no log at all** — launchd opens the
stdout path as the job's user, and `/var/log` is root-only. It is self-concealing: the log
that would explain the failure is the thing that cannot be created. The agent's log has to
live somewhere the tenant can write, and `spawn scheduled` with `active count = 0` should be
treated as a failure signature, not a transient.

Also confirmed, and relevant to baking the agent in: both `DARKBLOOM_LUME_GUEST_CHANNEL_FD`
(a real inherited descriptor) and `_PORT` are required with no CLI flag (`0005:99-117`);
`lume stop` from another process is refused as "liveness is inconclusive" by patches
0002/0003, so a VM must be stopped through its owner; and `/tmp` in the guest is cleared on
boot. Separately, `SandboxManagedProcess.receiveGuestChannelDescriptor()`
(`SandboxProcessRunner.swift:67-69`) is `public` with **zero callers** — the descriptor
adoption step does not exist yet and is the next thing to build.

---

### 10. The agent is baked in and the host adopts its channel

Stage 3c proved the transport worked by hand. This makes it happen by itself, and stops short of letting anything execute: production still refuses every command.

**Baking happens at the template layer.** `lume clone` is an APFS `clonefile` of the whole VM directory, and this PR already has clones inherit the template's accounts, so an agent installed once into a base image appears in every sandbox at no per-clone cost and with no per-clone step that can fail.

**The install splits along a privilege boundary that already existed**, because neither half can do the other's job. The offline patcher writes the executable and the LaunchDaemon plist into the mounted data volume — `writeData` always took an arbitrary mode, though this is the first file it writes executable. It cannot set an owner: there is no `chown` anywhere in lume's source except one line of guest-side shell. So the first-boot finalization script, which is already root and already does `chown root:wheel` and `launchctl enable` for other files, takes ownership and enablement. launchd refuses a daemon plist that is not `root:wheel`, so both halves are required.

The daemon runs as the unprivileged tenant, never the bootstrap administrator, and `--guest-agent` therefore requires `--tenant-user`. That is a correctness guard: baking an agent into a single-account image would leave it running as the account that can rewrite the control path supervising it, which is what splitting the identity was for.

**The missing host-side link is closed.** `SandboxManagedProcess.receiveGuestChannelDescriptor()` was `public` with **zero callers** anywhere — every VM created a channel and dropped it. Three properties of the handover shaped the adoption: it is one-shot (patch 0005 returns after the first success, so there is no reconnect), optional (only a VM this process spawned can have one, so absence is ordinary and SSH stays the fallback), and long-lived (the agent serves many sequential commands on one connection).

**An unverified channel is worse than no channel**, because it would be trusted for every command afterwards. The handshake is the only thing distinguishing our agent from anything else listening on that port inside the guest, so one that fails is dropped and its descriptor closed. `expectedImageID` needed no new state: the ownership record already stores what the VM was made from, and a clone's recorded source *is* the name its template was baked with.

Three defects were found and fixed on the way, two of them in this PR's own earlier work:

- **`launchctl` opens a job's stdout as the job's user.** A daemon running as the tenant with its log under `/var/log` fails `EX_CONFIG` (78), respawns forever, and writes no log at all — the diagnostic that would explain it is the thing that cannot be created. Measured at 45 respawns and climbing on a real guest, with no other symptom. The log now lives somewhere the tenant can write, and it is a test rather than a comment.
- **Delivery would have blocked the runtime actor.** `LumeGuestVsockTransport.deliver` is `async` but awaits nothing: it calls a synchronous poll that blocks until the agent answers. Straight from the actor that holds it for the whole command budget — up to 910 seconds at the 900-second cap — stalling every other operation including `stop`. The detach lives in the transport so it is correct however it is called.
- **Cancellation would have escalated to a forced VM teardown.** It boots out a launchd job by label, which only the SSH path creates; the agent spawns commands itself with `posix_spawn`, under no label. Ungated it would boot out a job that never existed, fail, and — because a cancellation failure forces `mustStopVM` and is rewrapped as `cleanupFailed` — turn a recoverable command failure into a VM teardown reported as a cleanup fault. It is gated on the transport now. Nothing is lost: the guest-side deadline still kills the process group, and the VM stop already forced ends the command with the machine. A cancel frame is the real fix and is its own change.

**Production is unchanged.** `guestCommandPolicy` is still `.disabled` on the serve path, so `execute` refuses before a transport is chosen, and `guestChannelPort` is nil there, so no channel is attached. Only `prepare-base` reaches any of this. Opening the gate is the next stage, with a threat-model entry as a prerequisite — the sandbox plane is currently absent from `docs/threat-model.yaml` entirely.

**Proven on real hardware, with no manual step inside the guest.** A `lume create` bakes the agent in — `Installed guest agent into the data volume bytes=290392 runAs=sandbox` — the offline setup completes and the health check passes on its first attempt. Booting that image with a channel attached then yields `{"magic":"darkbloom_guest_agent","protocol_version":1,"agent_version":"0.1.0","image_id":"agent-vm"}`, where the image id is the one the bake was given. The identity check has a real end-to-end path for the first time.

One bug surfaced only on hardware and is worth recording, because it wore two faces. `/var`, `/etc` and `/tmp` are symlinks into `private/` on a running macOS system, but a mounted data volume has no such links, so the agent's log directory was created where the guest could not see it. The visible symptom was `chown: /var/db/darkbloom: No such file or directory` on a path that looks perfectly correct — and, separately, `prepare-base` reporting a VM that "couldn't be moved to storage", which was just the promote step finding no VM after `lume create` had already failed on that same chown.

---

### 11. Base-image preparation actually reaches the guest

Section 10 left `prepare-base` building a correct image and then failing at the end. Three separate bugs sat behind that one symptom, and each was only visible once the one in front of it was fixed. All three are operator-path bugs; production still refuses every command.

```mermaid
flowchart TB
  subgraph Before["Before: three ways to fail as one timeout"]
    A1["prepare-base<br/>create + start"] --> B1["readiness probe"]
    B1 --> C1["wrapper targets /Users/lume<br/>the account 0006 deleted"]
    C1 --> D1["guard: [[ -d ... ]] || exit 70<br/>no output"]
    D1 --> E1["timed out: guest readiness"]
    E1 -.->|"and if it had passed"| F1["guest fact routed to vsock<br/>because a channel exists"]
    F1 --> G1["agent: execution_disabled"]
    E1 -.->|"and even with the right home"| H1["1s guest budget<br/>vs load average 60"]
    H1 --> I1["envelope 124 + timed_out<br/>120s window expires"]
  end
  subgraph After["After: PREPARE_EXIT=0"]
    A2["prepare-base<br/>create + start"] --> B2["readiness probe<br/>credential.bootstrapHome"]
    B2 --> C2["10s guest budget,<br/>600s window"]
    C2 --> D2["ready"]
    D2 --> E2{"handshake said<br/>execution_enabled?"}
    E2 -->|no| F2["route over lume ssh"]
    E2 -->|yes| G2["route over the channel"]
    F2 --> H2["sw_vers, uname read back"]
    C2 -.->|"if it still times out"| I2["...: the guest command hit<br/>its 10s deadline"]
  end
```

**The account.** `0006` replaced the shared `lume` account with a per-sandbox administrator and an unprivileged tenant, but two defaults in the guest command wrapper kept naming the account it deleted: `SandboxGuestCommandRequest.workingDirectory` and the encoder's `home`. Both are deleted rather than corrected — a working directory and a `HOME` have to exist inside one specific guest, so no compile-time value is right, and a missing argument should be a compile error rather than a path that silently does not exist. The readiness probe now takes the credential it was already authenticating as; reading a guest fact runs at the root; tenant work runs in the tenant's home.

I had this mechanism wrong at first and the correction matters, because it points at a different line. I wrote that launchd refuses a job whose `WorkingDirectory` is missing. Running the real encoded command by hand shows the wrapper's own guard fires first, in the outer SSH-delivered script, before any launchd job exists: with `/Users/lume` it exits 70 and writes nothing, and with an existing home it exits 0 and returns a well-formed envelope. Measured separately on a throwaway job, `launchctl bootstrap` succeeds and the job then fails at spawn with `last exit code = 78: EX_CONFIG`. The path is wrong twice over, but only the guard is reached.

Three signatures also still defaulted `credential` to `.legacy` — `waitForGuestReady`, the SSH transport's initialiser, and the readiness probe's `run`. Every production call site already passed one explicitly, so none was live, but they are latent traps of the same shape and are gone.

**The routing.** Adopting the channel made the host prefer vsock for every VM that had one, but a channel says nothing about whether the agent will run anything: the agent gates its own executor, hardcoded off until the production gate opens. So the moment adoption started succeeding, every operator command was routed to a peer that refuses commands. The handshake now carries `execution_enabled` and the host routes on it. A refusing agent is still a fully valid peer — the handshake is what proves the image identity, so refusing to execute must not fail readiness, and the channel stays adopted and verified while commands go over SSH. The field decodes leniently, so an agent baked before it existed still handshakes and reads as refusing. This also closes a smaller hole: a channel adopted but not yet handshaked used to be eligible, so an unidentified peer could be sent a command.

**The budget.** Base-image preparation was a coin flip, and the guest was not the problem. Booting a failed VM and running the probe's own command by hand, on the same machine a minute apart:

| load average | probe result |
|---|---|
| 60.46, one minute after first boot | `exit_code 124`, `timed_out true` |
| 28.66, two minutes after first boot | `exit_code 0`, `timed_out false` |

A guest that has just installed macOS is heavily loaded for the first couple of minutes, and the probe gave its `/usr/bin/true` a **one-second** guest-side deadline — so the wrapper spawned the process correctly and then killed it, and the envelope came back `124` and `timed_out`, indistinguishable from an executor that does not work. The probe was measuring load, not readiness. It gets ten seconds, still well inside the 35s Lume budget and the 40s attempt timeout, and the readiness window goes from 120 seconds to 600, because the thing it waits for is a machine that has just installed an operating system.

**And the diagnostic, which is the more valuable half.** A readiness timeout said only that it had happened, so telling "no SSH" from "busy guest" from "broken wrapper" cost a live VM, a credential recovered from the ownership record, and a hand-run command. The probe returned `Bool`, discarding everything it knew. It now returns the reason and the timeout repeats it — `base-agent-04 guest readiness: the guest command hit its 10s deadline, which a heavily loaded first boot can do` — and the three tests that pinned the old bare message now require an observation to be present.

The regression test decodes the base64 the encoder emits before asserting on it, because asserting on the wrapper proves nothing when every path inside it is unreadable and a `contains` check silently passes whatever it is given. It also proves its own absence assertion can fail, by encoding with the legacy credential and requiring that to contain the path the real one must not.

**Verified on real hardware.** `prepare-base` runs to completion and prints its report:

```
Prepared base-agent-06
Runtime: lume 0.5.3
Guest: macOS 26.6.2 arm64
CPU: 4
Memory bytes: 8589934592
Disk bytes: 107374182400
Guest agent: .../darkbloom-guest-agent
```

`macOS 26.6.2` and `arm64` are the two guest facts, read back through the wrapper that was broken. Two of the six runs died inside Apple's installer with `An error occurred during installation`, which is the pre-existing flakiness already noted on this branch and unrelated to any of this.

One thing found and deliberately parked. `VM.writeVNCConfigViaSSH` still authenticates as the deleted shared account. An earlier note here called it hung SSH; it is not — it is a detached task bounded at six attempts with a 10s SSH timeout, so it gives up silently after about 90 seconds. On every guest this branch builds it is dead code that can never succeed, worth fixing, but it cannot stall or fail a create.

---

### 12. The gates open, except the one that should not

Four independent things refused every tenant command. Three are ours to move and move here; the fourth is the packet gateway, and it stays shut on purpose.

```mermaid
flowchart TB
  subgraph Before["Before: four refusals, none of them computed"]
    A1["coordinator: execute"] --> B1{"isolationReadiness<br/>.unavailable, a constant"}
    B1 -->|refuse| X1["isolation_unavailable"]
    A1 -.-> C1{"guestCommandPolicy<br/>.disabled, hardcoded in the public init"}
    C1 -->|refuse| Y1["unsupported"]
    A1 -.-> D1{"agent executionEnabled<br/>false, hardcoded in main.swift"}
    D1 -->|refuse| Z1["execution_disabled"]
    A1 -.-> E1["isLegacyShared<br/>exists, read nowhere"]
  end
  subgraph After["After: three open, one honestly false"]
    A2["coordinator: execute"] --> B2{"isolationReadiness<br/>derived from configuration"}
    B2 -->|"networkPolicy false"| X2["still refuses:<br/>no enforced tenant network policy"]
    B2 -->|"once Stage 5 lands"| C2{"guestCommandPolicy<br/>.tenantExecution"}
    C2 --> D2{"credential legacy shared?"}
    D2 -->|yes| Y2["refuse: image predates<br/>per-sandbox identity"]
    D2 -->|no| E2{"handshake said<br/>execution_enabled?"}
    E2 -->|yes| F2["run it over the channel"]
    E2 -->|no| G2["run it over lume ssh"]
  end
```

**The runtime policy.** `LumeGuestCommandPolicy` gains `.tenantExecution`, and the guard asks the policy whether commands are permitted instead of matching one case, so a fourth case becomes a compile error here rather than a silent refusal there. The public initialiser keeps `.disabled`, which is not the hardcode it looks like: the policy type is package-scoped, so that initialiser exists for callers outside the package and refusing is the only safe answer for one that cannot name a policy. `ServeCommand` uses the package initialiser and asks for tenant execution explicitly, the same way `prepare-base` already asked for its own.

**The agent's executor becomes a property of the image.** It rides the LaunchDaemon plist beside the image id, set only by a new `lume create --guest-agent-allow-execution`, so an image is made executable deliberately rather than by default. The agent enables it on an exact `"1"` and nothing else — a permissive parse would turn a typo in a plist into an executing image, and this is the flag that decides whether a guest runs tenant code at all. The parsing moved into the core so the rule has a test instead of living in the executable's top-level code.

This is where the previous commit's handshake work pays off: an image baked before the flag existed reports `execution_enabled: false`, the host keeps using SSH for it, and there is no version negotiation anywhere.

**A legacy-shared credential is refused before a transport is chosen.** The flag has existed since per-sandbox identities landed and was read nowhere. It matters more than it looks: an ownership marker with no recorded credential resolves to the shared account, which is exactly what an image built before that change has, and that account can rewrite the launchd metadata supervising its own commands. The clearest evidence the guard is real is that the test fixture had to start recording a credential — it had been indistinguishable from a stale image.

**Isolation readiness is derived rather than asserted.** It was the constant `.unavailable`, and nothing in `Sources/` computed a value; the only place the flags were ever true was a test. Two are honestly true now. `signedGuestControl` means the pinned Lume passed provenance verification and a channel port is configured, so every VM must complete a validated handshake before its channel carries anything. `workspaceQuota` is enforced by construction: a guest's disk is a fixed-size image sized at create and validated against the resource policy, and a template whose disk differs is refused.

`networkPolicy` stays false, and that is the whole point. Nothing confines tenant egress yet, so `permitsJobs` is false and this host still refuses every sandbox job — the heartbeat reports `draining`, and prepare and execute are both refused at admission. Opening the other gates without the gateway would put tenant code on the provider's LAN. The daemon now says which flag is missing at startup instead of leaving an operator to work out why a healthy-looking host advertises itself as draining, and a test pins the dependency so setting the flag without a gateway fails.

**The threat model gets the surface before the gate opens.** It had none of it: no `sandbox`, `lume`, `vsock` or `guest` anywhere in 146 KB, no tenant among its adversaries, and a `threat-model-review` job grades every PR against it. Adds `ADV-004 malicious_sandbox_tenant`, the tenant workload and the guest-control channel as assets, `TB-011` for the host-to-guest boundary with how it actually works, and five threats: rewriting the control path that supervises a tenant's own commands, something other than our agent answering on the channel, LAN reachability, resource exhaustion through the channel, and driving an image that is not fit to run tenant code.

**Verified on real hardware.** `prepare-base` reaches `PREPARE_EXIT=0` with an executable agent baked, and the agent's own log inside the resulting image reads:

```
darkbloom-guest-agent: listening on vsock port 8888
darkbloom-guest-agent: execution enabled for this image
darkbloom-guest-agent: accepted a broker connection
```

That is the flag arriving end to end — CLI flag to plist to agent — with the daemon `root:wheel` and its log directory owned by the tenant, so launchd loads it and it can write. Stated precisely: the agent logs only failures per command, so what is directly observed is that the agent bound, advertised execution, and was connected to over the channel, and that the guest facts came back correct. Which transport carried each individual command is covered by tests rather than by a log line.

**One thing seen twice and not explained.** Two images built by one particular pinned-Lume binary came out with the agent owned by the bootstrap account rather than `root:wheel`, so launchd never loaded it, and their first-boot finalization took 4 seconds against the usual 15. Four images built from freshly built Lume — debug, release, and a fresh `build-pinned-lume.sh` output — are all correct, and applying the patch chain to a clean `737dc2a0` archive reproduces the working tree exactly. It reproduced only with that one binary and not with a fresh build of the same inputs, so it is recorded rather than diagnosed.

**Pre-existing and left alone:** `docs/threat-model.yaml` has a duplicate threat id, `T-041`, at lines 1601 and 2163 on master. Renumbering ids in a security document silently is worse than reporting it.

Patch **0008** changes, so it is re-pinned in the lock file and the Swift constants, and its two new tests are added to the named-test gate that verifies each required Lume test executes exactly once. Patches 0001 to 0007 are byte-identical.

---

### 13. The plane goes live: readiness off the network, tenants off the network

`permitsJobs` had one flag left false, so a provider still advertised `draining` and refused every sandbox job. The plan made that flag wait on a packet gateway — the largest piece left — and nothing would run until all of it worked. Confining tenant egress to *nothing* gets there first, and is stronger than what the gateway will offer.

```mermaid
flowchart TB
  subgraph Before["Before: readiness needs an IP, so isolation is impossible"]
    A1["start"] --> B1["inspect() -> guestReady"]
    B1 --> C1["= details.sshAvailable<br/>DHCP leases + ARP scrape"]
    C1 --> D1["SSH probe over the guest IP"]
    D1 --> E1["ready"]
    C1 -.->|"remove the NIC"| F1["sshAvailable false forever<br/>readiness times out"]
    G1["isolationReadiness<br/>networkPolicy = false, a constant"] --> H1["permitsJobs false<br/>every job refused"]
  end
  subgraph After["After: the agent is the readiness signal"]
    A2["start"] --> B2["adopt + validate channel"]
    B2 --> C2{"agent serves commands?"}
    C2 -->|yes| D2["run a no-op over vsock<br/>check a real envelope"]
    D2 --> E2["ready — no IP consulted"]
    C2 -->|no| F2["inspect -> SSH probe<br/>unchanged"]
    G2["tenantNetworkPolicy"] -->|isolated| H2["lume run --network none<br/>no device at all"]
    H2 --> I2["networkPolicy true<br/>permitsJobs TRUE"]
    G2 -->|hostNAT| J2["still refuses"]
  end
```

**Readiness was a networking fact.** `guestReady` is literally `details.sshAvailable`, scraped from DHCP leases and the ARP table, and the probe that followed needed that address — so a guest with no network device could never become ready, and any isolation work was blocked behind it. A VM whose agent serves commands is now proven ready over the channel. The ordering already allowed it: adoption runs immediately before the readiness wait, so the channel and its validated handshake are already in hand. The two other readiness call sites are the branches that never spawn a process and so can never have a channel; they keep the SSH path exactly as it was.

It runs a real no-op and checks a real envelope rather than treating the handshake as sufficient. The handshake says the agent bound its port and is the image this host provisioned; it says nothing about whether the executor works, which is the whole thing the SSH probe existed to establish. An agent that handshakes and then refuses every command is precisely the case that would otherwise report ready, and there is a test for it.

The probe needs the handshake frame already consumed — the agent sends it first, so a command issued before it reads a handshake where a result belongs. That is guaranteed structurally rather than by comment: the guard is `servesCommands`, set only inside `validate`, after a successful handshake.

**Tenant VMs now get no network device.** Lume patch **0009** adds a mode that attaches none: not a device with a dead attachment, but no `VZNetworkDeviceConfiguration` at all, so there is no interface for a guest to bring up and none it can add. `createNetworkDeviceConfiguration` returns an optional and the two attachment sites build an empty list. The switch stays exhaustive, so a fourth mode is still a compile-time break — which is exactly how this change was caught the moment the case was added.

The mode is chosen per run rather than baked into the image, which is what keeps this small: one base image is still built over host NAT, because the unattended install's health check and first-boot finalization both run over SSH, and the same image runs isolated once it carries tenant code.

`networkPolicy` reads that choice instead of being a constant. True for isolated, because egress is confined by a decision the host makes at spawn and a guest cannot undo; false for host NAT, which would put tenant code on the provider's LAN. **A packet gateway granting filtered egress is a later stage, and must add a third case rather than flip this flag for NAT.**

Base-image preparation deliberately keeps host NAT. If a bake ever fails, guest facts still fall back to SSH instead of the operator path dying outright. It was verified live at isolated too, so that is a choice about keeping the fallback rather than a limitation.

**Verified on real hardware, two ways.** `prepare-base` with its verification boot isolated reaches `PREPARE_EXIT=0` — which it can only do if readiness and both guest facts travelled the channel, because no address exists to reach. And on one VM with one MAC:

| run flag | ping to its leased address |
|---|---|
| `--network none` | 2 sent, **0 received**, 100% loss |
| `--network nat` | 2 sent, **2 received**, 0.3 ms |

🛑 Worth knowing for anyone checking this by hand: `lume get` still prints an IP for an isolated VM. That field is a DHCP lease lookup by MAC, not a live property of the running guest, so a stale lease from an earlier NAT boot survives there and looks exactly like working networking. The ping above is what settles it.

**What a sandbox cannot do yet:** reach anything. That is the deliberate trade — the plane is live with the strongest isolation available, and the gateway becomes an additive change that grants filtered egress rather than a prerequisite blocking everything.

---

### 14. Frames reach a gateway (packet gateway, step 1 of 4)

The gateway is the last thing between this branch and sandboxes that can fetch dependencies. It is also the largest piece left, so it lands in four steps with a gate each. This is the first: get the guest's frames into a process that can decide what to do with them. **Nothing filters yet and nothing claims to.**

```mermaid
flowchart LR
  subgraph Before["Before: the kernel decides"]
    G1["guest NIC"] --> N1["VZNATNetworkDeviceAttachment"]
    N1 --> H1["host LAN + internet"]
    N1 -.->|"no inspection point"| X1["nothing can drop a packet"]
  end
  subgraph After["After: the host decides"]
    G2["guest NIC"] --> A2["VZFileHandleNetworkDeviceAttachment<br/>connected datagram socket, fd 5"]
    A2 --> C2["GuestFrameChannel<br/>1 datagram = 1 Ethernet frame"]
    C2 --> E2["EthernetFrame<br/>raw EtherType kept"]
    E2 -.->|"5c-2..4"| F2["ARP / DHCP / UDP / TCP<br/>+ per-packet filter"]
  end
```

**Why this shape.** Apple's header says the attachment exchanges data "at the level of the data link layer" over a handle that "must hold a connected datagram socket". One datagram is exactly one Ethernet frame, which removes every framing concern a stream socket would bring — no length prefix, no partial frame, no reassembly, and two frames never splice.

**Two refusals are deliberate.** A stream socket is rejected before it is ever attached: it would be accepted and then deliver spliced frames that look like guest corruption rather than a wrong socket type. And gateway mode with no inherited descriptor is an error, not a fallback to NAT — a caller that asked for a filtered guest must never silently get an unfiltered one.

**The descriptor reuses the guest channel's machinery.** Fd 3 is the cooperative control channel and fd 4 the guest vsock channel, so the gateway takes fd 5, named through the environment because `POSIX_SPAWN_CLOEXEC_DEFAULT` closes everything the parent does not explicitly `dup2`. The host end is non-blocking: a blocking read would stall the daemon on a guest that never speaks, and a blocking write on one that has stopped reading.

**It runs in the daemon, not a helper.** A userspace NAT needs no privileges, so a separate process would buy parser fault-isolation at the cost of a second signed executable, provenance pinning, descriptor hand-off and explicit fail-closed handling. In-process, fail closed is structural — the execution closes the descriptor on teardown, so frames stop when the VM does. It is also the same class of surface the daemon already has, since it already parses untrusted guest frames from the vsock channel (**T-054**).

**The policy deliberately still reports no confinement.** Frames reach the gateway long before anything drops them, and a policy claiming `networkPolicy` at this point would open the plane to guests with *unfiltered* egress. There is a test pinning that, because it is the one mistake in this step that would be both silent and serious.

**Verified on real hardware.** A base-image preparation run with a gateway attached completes and promotes its image, and the daemon decoded the guest's first eight frames off the socket:

```
frame 2  bytes=343  type=0x800   dst=ff:ff:ff:ff:ff:ff
frame 1  bytes=110  type=0x86dd  dst=33:33:00:00:00:16
frame 7  bytes=207  type=0x86dd  dst=33:33:00:00:00:fb
```

The first is the DHCP DISCOVER a booting guest broadcasts. The rest are IPv6 — MLD and mDNS — which the guest starts emitting **immediately and repeatedly, before it has an IPv4 address at all**. That turns "IPv6 off" from a tidiness item into a real one: under NAT those frames reach the host's LAN, and they are the first thing a guest sends.

Patch **0010** is pinned; 0001–0009 are byte-identical. Steps 2 to 4 — ARP/DHCP/IPv4/UDP with the filter, then TCP, then rate limits and the policy flip — follow.

---

## Checklist to a working sandbox

Ticked as commits land on this branch.

### Transport and identity
- [x] Lume patch: attach `VZVirtioSocketDeviceConfiguration`
- [x] Lume patch: bridge the guest vsock connection to an inherited descriptor
- [x] Daemon-side channel: socketpair at child fd 4, `recvmsg` + `SCM_RIGHTS`
- [x] Guest-side agent skeleton binding `AF_VSOCK`
- [x] Prove a macOS guest binds its vsock listener early enough in boot (9s cold boot to handshake, against a 600s host retry window)
- [x] Randomized per-sandbox bootstrap credentials, replacing `lume`/`lume`
- [x] Unprivileged sandbox user in the base image
- [x] Base-image preparation completes end to end (Lume patch 0007; it used to die with exit 70 before the patcher ran)
- [x] Base-image preparation reaches a working guest and reads facts back through the command wrapper (`PREPARE_EXIT=0`)

### Guest agent
- [x] Framed request/response protocol over the channel
- [x] Authenticated handshake reporting image id and agent version
- [x] Spawn command processes (unprivileged user pending Stage 3)
- [x] Stream stdout, stderr and exit status
- [x] Guest-side deadline enforcement
- [ ] Bounded file upload and download
- [ ] Disk and process health reporting
- [x] Treat every guest message as untrusted host-side
- [x] Bake the agent into the base image (Lume patch 0008; signing is a Stage 4b item)
- [x] Transport seam, and readiness over the channel
- [x] Route commands only to an agent whose handshake says it serves them
- [x] Open the production `LumeGuestCommandPolicy` case (`.tenantExecution`; the adapter still refuses, see below)
- [x] Refuse execution for an image that predates per-sandbox credentials
- [x] Compute isolation readiness instead of asserting it — `networkPolicy` stays false until the packet gateway
- [x] Threat model covers the sandbox plane (ADV-004, A-015/A-016, TB-011, T-051…T-055)

### Network isolation
- [x] Readiness no longer depends on guest IP discovery (proven over the channel)
- [x] Tenant VMs run with no network device at all (Lume patch 0009) — `networkPolicy` true, `permitsJobs` **true**
- [ ] Signed packet-gateway helper behind `VZFileHandleNetworkDeviceAttachment` (grants filtered egress; renumbers to 0010)
- [ ] DHCP, DNS forwarding, stateful TCP/UDP NAT
- [ ] Per-packet destination filtering, so DNS rebinding cannot bypass private-range blocks
- [ ] Default-deny until policy and fencing token are installed
- [ ] Byte and connection rate limits; IPv6 disabled until equivalent filtering exists
- [ ] Fail closed if helper, daemon, policy lease or socket dies

### Billing
- [ ] Quotes as separate expiring objects
- [ ] Durable balance holds
- [ ] Metering from `ready` to durable `metering_ended_at`
- [ ] Settlement and usage records
- [ ] Rate card: CPU, memory, chip family, workspace, retention, exclusivity
- [ ] Failed starts are free; refunds work
- [ ] Retained-storage metering
- [ ] Provider payout split and net floor
- [ ] API-key sandbox policies — `requireSandboxAuth` currently admits any valid key

### Persistence
- [ ] Checkpoint transport to an object store — no upload or download code exists today
- [ ] Encrypted boot, workspace, auxiliary and configuration snapshots
- [ ] Merkle manifest with compare-and-swap against parent
- [ ] Dual recovery modes, platform-managed and tenant-managed
- [ ] Portable restore across hosts
- [ ] Short-lived object credentials scoped to one sandbox generation

### Gates that are not engineering
- [ ] Production code signing under team `SLDQ2GJ6TL`
- [ ] Named legal owner records launch approval

Two things worth flagging from reading the plan against the code. **Four of the six properties Phase 1's own gate demands are billing properties** — no double charge, no leaked hold, no unfunded execution, no third billable admission — so that gate cannot be met by lifecycle work alone. And **the packet gateway is the same size as the agent** and blocks the same phase; it is easy to under-weight because the agent is the more interesting problem.

Phases 3 to 5 — sticky cache and GitHub Actions, macOS Computer, Linux Firecracker — are expansion, and out of scope here.

## Deploy

No deploy impact so far. `guestChannel` and `guestChannelPort` default to `nil` throughout, so no existing caller attaches a device or a channel. Tenant execution is still refused with `execution_disabled`; the production `LumeGuestCommandPolicy` case does not open until the agent is baked into the signed image.

The patch chain is now seven, so the provenance manifest changes and a host running the old pinned Lume fails closed against the new pin, as intended. Patches 0001 to 0004 are byte-identical to what #672 shipped, and 0005 is unchanged since it landed earlier on this branch. 0003 and 0004 stay untouched despite both defects above living in them: the fixes are additive in 0007 so the diff against the original work stays reviewable.

Providers holding an image prepared before 0007 are unaffected, since the bug was in preparing an image rather than in running one. Any image prepared with 0006 but before the health-check fix will not exist, because that path could not complete.


Gajesh2007:
We are working towards our sandbox stuff. We will re-open and re-use the PR as we restart work on it.

owizdom:
okayy, sure sure.

PR #806, opened 2026-09-02
docs(systemmap): derive the Darkbloom system map from source

ethenotethan:
Generates the Darkbloom system map **deterministically** from source — no agentic
analysis — and publishes it as a self-contained explorer. The coordinator is the
first service extracted; the IR is service-agnostic, so the Swift provider and the
two consoles land as additional clusters later without a schema change.

`tools/systemmap` is its own Go module, so `go/packages` never reaches the
coordinator's build and `go test ./...` at the repo root is unaffected.

Merged with current master and re-derived against it: the branch now also carries
the **history slider** (a second, opt-in artifact that steps the map through the
service's whole lifespan), one root-cause fix the merge surfaced in the SQL reader,
and the curation master's App Attest work needs (see *Merging master again*).

## The map is generated, not committed

CI builds it for every pull request and uploads `system-map.html` as a job artifact;
`pages-systemmap.yml` builds and publishes it from master. Nothing generated is in
git — only the curated `overlay.json` and the README are. A stored copy of a derived
artifact is a second source of truth that has to be kept honest; building it means
the map cannot be stale, reviewing it means opening the HTML rather than reading a
diff of 34k lines of JSON, and a coordinator change costs no diff at all.

What survives is the gate that actually matters: generation fails when source
contains something the curated half does not explain.

## What it derives

Type-checked with `go/packages` and never hand-edited: the route table out of
`(*api.Server).routes`, the middleware chain each handler was wrapped in
(recognized by the `HandlerFunc → HandlerFunc` shape), the authorization gates a
route's code calls, the dependency nodes its reachable code touches (struct fields,
SQL tables through interface dispatch into the Postgres store, URL/endpoint
literals), the R/W/RW access mode per association, every `pg.*` table's definition
(`CREATE TABLE` **plus** the `ALTER TABLE … ADD COLUMN` migrations that grew it),
and a `file:line` citation for all of it.

Counts today, at `51b3e962f`: **108 routes, 116 nodes, 20 groups, 295
associations, 990 wiring steps, 58 table definitions (756 columns, 33
migration-added), 15 drawn foreign keys, 1,216 citations**.

Curated in `overlay.json`, because source cannot state it: node identity/labels,
which cluster each kind of node lives in, namespace and auth-class names, callers,
and prose. That split is the point — a reader can tell which facts a compiler
stands behind.

## Behavior: before → after

```mermaid
flowchart LR
  subgraph Before
    A1[change a route, a field,<br/>a query, an outbound host] --> B1[hand-written prose in docs/]
    B1 --> C1{is it still true?}
    C1 -->|nobody knows| D1[silent drift]
    A1 --> E1[CI: builds + tests]
    E1 --> F1[green]
  end
  subgraph After
    A2[change a route, a field,<br/>a query, an outbound host] --> E2["CI System Map job:<br/>build the map at the PR head"]
    E2 --> H2{overlay explains<br/>everything source found?}
    H2 -->|no| I2[red: report.md names<br/>the unexplained thing]
    H2 -->|yes| F2[green + system-map.html<br/>uploaded for review]
    F2 --> P2[merge → Pages rebuilds<br/>and publishes from master]
    A2 -.->|nothing to commit,<br/>nothing to go stale| G2[every claim carries file:line]
  end
```

Unexplained source is a build failure, in both directions: a new `Server` field with
no node, a query naming an unknown table, a host with no boundary, a route no auth
rule matches, a node with no prose, a `pg.*` node with no `CREATE TABLE`, a struct
carrying a mutex that only a package-wide default explains — and, reversed, prose for
a route that no longer exists or an endpoint literal that vanished from its client
package.

## Code: before → after

```mermaid
flowchart TB
  subgraph Before
    S1[coordinator source] -.->|read by a human| P1[docs/*.md prose]
    P1 --> R1[reader]
  end
  subgraph After
    S2[coordinator source] --> L2["extract.Load<br/>go/packages + go/types"]
    L2 --> RT["extract.Routes<br/>route table as syntax"]
    L2 --> WK["extract.Walker<br/>transitive call graph,<br/>interface dispatch, modes"]
    L2 --> SC["extract.SchemaDefinitions<br/>CREATE + ADD COLUMN → ir.Table"]
    RT --> AS[assemble.Build]
    WK --> AS
    SC --> AS
    OV[overlay.json<br/>labels, clusters, prose<br/>the only committed input] --> AS
    AS --> IR["ir.Graph<br/>routes · nodes · groups · clusters ·<br/>edges · tables · citations"]
    AS --> RP["report.Report<br/>12 drift gates"]
    IR --> HT[render.HTML]
    IR --> JS[inventory.json]
    RP --> MD[report.md]
    HT --> OUT["ephemeral output<br/>(git-ignored)"]
    JS --> OUT
    MD --> OUT
    OUT --> CI[CI job artifact]
    OUT --> PG[Pages site]
    RP -->|not clean| FAIL[exit 1]
  end
```

`-check` runs that entire path — including the renderer, so a broken template fails
it — and writes nothing.

## Reading the map: three provenance views

The split between derived and curated is only useful if a reader can see it, so the
page reads three ways.

```mermaid
flowchart LR
  subgraph Before["Before — one provenance checkbox"]
    B1[reader opens the map] --> B2["Provenance: on"]
    B2 --> B3["curated words get tinted<br/>and labelled 'prose'"]
    B3 --> B4{"which of these facts<br/>does the compiler stand behind?"}
    B4 -->|"take the page's word for it"| B5[unfalsifiable claim]
  end
  subgraph After["After — a subtractive view"]
    A1[reader opens the map] --> A2{choose a view}
    A2 -->|All| A3[every fact, nothing marked]
    A2 -->|Code only| A4["withhold every string<br/>no gate can contradict"]
    A2 -->|Overlay marked| A5["mark each curated layer<br/>where it appears"]
    A4 --> A6["what is left IS the derived map:<br/>node ids, edges, modes, citations,<br/>table definitions, gate-checked identifiers"]
    A3 --> A7["topology fingerprint<br/>read back off the drawn DOM"]
    A5 --> A7
    A6 --> A7
    A7 --> A8["identical in all three views &rarr;<br/>'the overlay is additive' is checked,<br/>not asserted"]
  end
```

`Code only` is mechanical rather than curated-by-hand: it withholds *every overlay
string whose text no gate constrains* — node labels, endpoint descriptions and
details, all `depDocs` fields, callers, auth detail sentences, category blurbs,
cluster titles, actors, credentials — and keeps the derived facts plus the curated
identifiers the drift gates prove against source in both directions (node ids,
namespaces, auth classes). The banner counts what it removed: *116 node labels, 108
endpoint descriptions, 928 node prose fields, 14 actors, 13 credentials*. Search
follows the same rule, so the view never selects a row for a word it refuses to show.

## The ontological axis: who named each node

Whether source *reaches* a node is derived. How much of a node's *identity* a person
invented is a separate question, and it now has an answer in the IR: `namedBy`
records which mapping table names the node, computed from the overlay's own tables
and the derived schema. The identity filter selects on it — **source** (the name is
in the `CREATE TABLE`: 54 nodes), **literal** (a curated name bound to a string the
compiler found: 17), **symbol** (bound to a Go symbol: 45), **unreached** (declared
but reached by no endpoint: 6, all background-worker surfaces).

That last value is a node-level question, so it selects those nodes directly rather
than through endpoint reachability — otherwise the one filter whose purpose is to
reveal them hides them, which is the regression `TestPageExplorerAffordances` pins.

## The explorer

The graph is the page: viewport-sized by default, **Full screen** on demand (the
info panel, schema drawer and zoom controls live inside the fullscreen element, so
they survive the transition). Three nested levels, all derived — node (endpoint
square / dependency circle) → group (one per endpoint namespace, one per dependency
category) → cluster (a process, a datastore it owns, a third party). Membership is
indirect: a category names its cluster, a namespace joins its service's, so a second
extractor adds a boundary without touching the renderer. The layout packs each
cluster's group discs with a hard separation pass and clamps every node to its
group's disc, so containment is geometric rather than hoped for.

It is an explorer rather than a poster:

```mermaid
flowchart LR
  subgraph Before["Before"]
    C1["labels drawn inside the scaled scene"] --> C2["text scales with the zoom &rarr;<br/>collides identically at every zoom"]
    C2 --> C3[193 long curated labels<br/>bury the picture]
    C4[click a node] --> C5["filter the tables;<br/>graph unchanged"]
    C6[click a pg table] --> C7["440px drawer<br/>vs a 79-column table"]
    C8[click an in-memory node] --> C9["1 of 8 overlay prose<br/>fields rendered"]
  end
  subgraph After["After"]
    D1["#glabels outside #scene"] --> D2["constant screen size,<br/>placed per frame, greedy<br/>declutter + budget"]
    D2 --> D3["short ids on the canvas,<br/>0 overlaps at any zoom"]
    D4[click a node] --> D5["shadow everything off its edges<br/>+ pin the panel"]
    D5 --> D6["Esc / empty canvas clears"]
    D7[click a pg table] --> D8["560px drawer, widenable<br/>to full width; reached-by chips"]
    D9[click an in-memory node] --> D10["derived reach + citations +<br/>all 8 prose fields (~2.7KB)"]
  end
```

Clicking an endpoint opens its inventory row; clicking a dependency filters the page;
clicking a `pg.*` node opens that table's derived definition — every column with its
type, the rest of its declaration, and its own source link, with `ALTER`-added ones
tagged `migration`. `users` is the case that makes the point: 6 columns in its
`CREATE`, 12 in production.

The page's stylesheet and behaviour are separate source files — `render/page.css`
and three script parts, `render/page.js` (data, filters, layout, drawing),
`render/page.timeline.js` (the history replay and the slider) and
`render/page.boot.js` (`draw`, the control wiring, the bootstrap) — injected verbatim
at render time, because a couple of thousand lines of layout, provenance and replay
logic inside a `<script>` element is not reviewable. Their concatenation order is a
correctness contract rather than a preference, since the bootstrap's first `draw()`
reads a `const` the timeline part declares. The published artifact is still one
self-contained file with no remote asset.

## Reading a wire, and getting the corner back

A picture of 990 wiring steps is only worth having if a single wire can be read. The
graph now says three things per wire that it previously left to the panel — the
order the request meets it, the indirection the handler reaches through, and, for
the tables themselves, the foreign keys they declare — and the idle legend no longer
holds a corner of the canvas it is not using.

```mermaid
flowchart LR
  subgraph Before["Before"]
    A1[focus an endpoint] --> A2["its wires promoted,<br/>but in no stated order"]
    A3[a wire's dash] --> A4["one vocabulary for<br/>every kind of call"]
    A5[table &rarr; table] --> A6["no relationship drawn;<br/>FKs only in the schema drawer"]
    A7[enter full screen] --> A8["300px legend rides along,<br/>unaskable and unclosable"]
  end
  subgraph After["After"]
    B1[focus an endpoint] --> B2["wires numbered in the derived<br/>order the request meets them,<br/>with a lead tick to the curve"]
    B3[a wire's dash] --> B4["4 kinds: direct, interface<br/>dispatch, deferred, goroutine —<br/>legend printed from the generator"]
    B5[table &rarr; table] --> B6["15 dotted arcs from the<br/>REFERENCES the service issues,<br/>+ a relationships drawer"]
    B7[enter full screen] --> B8["legend folds to its heading<br/>+ a &plus; button; leaving reopens it;<br/>foldable by hand at any size"]
    B8 --> B9["a hover still unfolds for the<br/>node answer, then refolds"]
  end
```

The fold is one flag (`state.key`), so the two paths that change it — the button and
`fullscreenchange` — cannot disagree, and folded the panel drops its fixed width
rather than keeping a 300px box around one line of text.

## The history slider: the same map at every commit

The map answers *what the coordinator looks like now*. The next question a reader
asks is *when did it come to look like this*, and nothing in the artifact could
answer it. `cmd/history` does, without inventing a second extractor: it re-runs
**this** one at every commit that touched the service — one full `go/packages`
type-check each, so the walk costs minutes rather than seconds — and writes one
delta-encoded `history.json`. On this branch today that is **397 snapshots spanning
2026-03-21 → 2026-09-15 in 195 KiB**, and `-history` embeds it as a slider over the
service's whole lifespan.

```mermaid
flowchart LR
  subgraph Before["Before — when did this appear?"]
    A1["reader wants the shape<br/>of the service at some commit"] --> A2["git log -S / blame<br/>one file at a time"]
    A2 --> A3["build the map by hand<br/>at that commit, if at all"]
    A3 --> A4["no answer for<br/>'what did the wires look like'"]
    A5["a route or table that<br/>has since been deleted"] --> A6["invisible: the map only<br/>ever holds HEAD"]
  end
  subgraph After["After — one artifact, every commit"]
    B1["make history:<br/>walk every commit that<br/>touched the service"] --> B2["one full type-check per commit<br/>(1 of 397 unextractable —<br/>recorded in failed, not skipped silently)"]
    B2 --> B3["history.json: delta-encoded<br/>routes · nodes · wires · modes"]
    B3 --> B4["go run . -history<br/>embeds it"]
    B4 --> B5["slider + play: step the<br/>drawn map through the lifespan"]
    B5 --> B6["what a revision lacks is absent;<br/>what THIS commit deleted is ghosted"]
    B5 --> B7["Now restores the head map<br/>byte-for-byte in the fingerprint"]
  end
```

Two invariants make it safe to ship inside the same page:

- **Embedding is the flag, never the file.** A map built without `-history` is the
  map of one commit whatever is lying in the output directory (`main.go` reads the
  timeline only under the flag, and requires ≥2 snapshots for a slider to mean
  anything).
- **The slider is inert until touched.** The layout is the union's, computed once, so
  positions never reshuffle mid-slide — but what the head revision does not contain
  is drawn `.absent` and excluded from the topology fingerprint. Measured on this
  branch: the plain page draws 224 nodes / 990 wires, the `-history` page carries 15
  more nodes and 154 more wires in the DOM, and **both fingerprint `f6c65e9c`** — in
  all three provenance views. Returning to Now restores it.

Code-side, the walk is its own package rather than a flag threaded through the
generator:

```mermaid
flowchart TB
  subgraph Before["Before"]
    P1[main.go] --> P2[extract → assemble → render.HTML]
    P2 --> P3["one commit: HEAD"]
    P4["render/page.js<br/>(one script part)"] --> P2
  end
  subgraph After["After"]
    Q1["cmd/history/main.go"] --> Q2["history.Walk:<br/>git log --follow per module era,<br/>overlay retargeted per era"]
    Q2 --> Q3["extract.Load per commit<br/>(all-or-nothing type-check:<br/>a partial program is not a map)"]
    Q3 --> Q4["history.Shape:<br/>routes · nodes · links · modes"]
    Q4 --> Q5["history.Delta:<br/>one snapshot = diff vs previous"]
    Q5 --> Q6[history.json]
    Q6 -.->|only with -history| R1[main.go readTimeline]
    R1 --> R2["render.HTML(g, inventory, timeline)"]
    R3["page.js + page.timeline.js + page.boot.js<br/>concatenated in that order"] --> R2
    R2 --> R4["one page, slider inert<br/>until the reader moves it"]
  end
```

## Merging master: a rule where there was a list

Master's code-attestation coverage upsert reads `jsonb_to_recordset($1::jsonb)` in
`FROM` position, and the map went red on it: an unknown table with no
`CREATE TABLE`. The gate was right to fire and the reader was wrong — `sqlNoise`
listed *function names* (`unnest`, `generate_series`), so it was a census of the
set-returning functions the store happened to call, and the first new one broke the
build. The fix is a syntactic test, `isCallAt`, in **both** readers (the statement
scan and the fragment audit, so the two agree about what the text names).

```mermaid
flowchart LR
  subgraph Before["Before — a list of names"]
    S1["FROM jsonb_to_recordset($1::jsonb)"] --> S2{"name in sqlNoise?"}
    S2 -->|"no — nobody added it"| S3["read as a table"]
    S3 --> S4["red: table with no CREATE TABLE"]
    S5["FROM unnest($1::text[])"] --> S2
    S2 -->|"yes"| S6["skipped"]
  end
  subgraph After["After — a property of the text"]
    T1["name in a from-list position<br/>FROM · JOIN · DELETE … USING"] --> T2{"isCallAt: next<br/>non-space byte is '('?"}
    T2 -->|yes| T3["set-returning function:<br/>names no table"]
    T2 -->|no| T4["table"]
    T5["INSERT INTO usage (id, tokens)<br/>CREATE TABLE models (...)"] --> T6["untouched: the rule reaches<br/>from-list keywords only"]
    T7["FROM %s(...) — a splice"] --> T8["still reported:<br/>readability is checked first"]
  end
```

It is decidable rather than heuristic: PostgreSQL takes a column-alias list only
*after* an alias, so a bare table name directly before `(` is illegal — while a
write target and a `CREATE TABLE` are very often followed by a parenthesis, which is
why the rule reaches only the keywords that introduce a from-list item. Which is all
three of them, and getting that wrong was the regression review caught on this branch:
the first version guarded `FROM` and `JOIN` while taking the function names out of
`sqlNoise`, so `DELETE FROM usage u USING unnest($1::text[]) AS ids` read `unnest` as
a table nothing declares — the same failure the rule replaced, one keyword over. The
call-or-table distinction now travels with the matcher rather than with the read/write
mode, because `USING` is scored as a write and still reads a function's rows. An index
method (`CREATE INDEX ... USING gin (tags)`) falls out for free, and no list of
function names would ever have covered it. Seven `TestTables` cases pin both
directions and two fixture store methods prove it end to end through the real
pipeline.

Review of the merged branch then found the other half of the same mistake, and it was
worse than the first: reaching the call rule to `USING` in the statement scan without
reaching `USING` to the *readability audit* left the gate with a hole it did not have
before. `Tables` had learned that a `USING` operand may be a function; `auditText`'s
keyword set was still `FROM|JOIN|INTO|UPDATE`, so a `USING` whose table was spliced in
at run time named a table neither reader could see and **nothing at all was reported** —
counts balanced, map green, one table missing. `reTableName` and `reTrailingKeyword` now
carry `USING`, and a `fromListKeyword` set draws the same call-or-table line the scan
draws, so the false positives stay quiet.

```mermaid
flowchart LR
  subgraph Was["6bf0c1856 — one reader widened"]
    A1["DELETE FROM models m USING &quot;+other"] --> A2["Tables: USING may be a call ✓"]
    A1 --> A3["auditText: USING not a keyword"]
    A3 --> A4["no table read, nothing reported<br/>silent hole"]
  end
  subgraph Now["51b3e962f — both readers"]
    B1["DELETE FROM models m USING &quot;+other"] --> B2["auditText: USING is a keyword"]
    B2 --> B3{"fromListKeyword?<br/>bare ident + '('?"}
    B3 -->|"spliced / trailing"| B4["reported: unreadable statement"]
    B3 -->|"USING gin (tags)<br/>JOIN x USING (id)"| B5["quiet: call or column list"]
  end
```

Three fixture store methods and three pipeline cases cover both directions; the two
opaque ones report nothing without the regex change, which is how they were verified.

That merge's other drift was curated, not derived: the registry's new
`cacheDemandTracker` joins its four sibling sub-trackers on `mem.cache_routing`.
Across every merge with master on this branch, textual conflicts have been confined
to `> Last updated:` freshness stamps — three of them, each resolved to master's
newer line.

## Merging master again: App Attest and machine identities

Master's #995 landed the App Attest shadow archive and stable machine identities — 13
new `pg.*` tables, two new in-memory surfaces, 9 declared foreign keys of which 8 are
drawn (`darkbloom_machines.merged_into` points at its own table, and the graph draws no
self-loop). That is
exactly the change this gate exists for, and it failed the build in three places at
once rather than drifting quietly: **10 unmapped `Server` fields**, **2 concurrent
types absorbed by a package-wide default** (`appAttestShadowSession` and
`machineInventorySession`, both mutex- or channel-bearing, so a package default is not
allowed to explain them), and **13 `pg.*` nodes with no prose**.

```mermaid
flowchart LR
  subgraph Before["Before — the merge lands"]
    M1["master #995:<br/>app_attest_* + darkbloom_machine_*<br/>+ appAttestShadow* fields<br/>+ machineInventorySlots"] --> M2["git merge into the branch"]
    M2 --> M3{"make -C tools/systemmap check"}
    M3 --> M4["red: 10 unmapped fields"]
    M3 --> M5["red: 2 concurrent types<br/>absorbed by packageDefault"]
    M3 --> M6["red: 13 pg nodes<br/>with no depDocs"]
  end
  subgraph After["After — the overlay explains it"]
    N1["deps.fields: 5 keys<br/>(shadow slots, storage once/slots,<br/>machineInventorySlots, @skip)"] --> N7
    N2["deps.types: the two session<br/>structs named explicitly"] --> N7
    N3["deps.packageDefault:<br/>coordinator/appattest = @skip<br/>(Key/Receipt are parsed values,<br/>not shared state)"] --> N7
    N4["labels + depDocs:<br/>mem.app_attest_shadow<br/>mem.machine_inventory"] --> N7
    N5["depDocs: 13 pg tables,<br/>hand-written, each with<br/>file:line sources"] --> N7
    N7{"make check"} --> N8["108 routes, 116 nodes,<br/>295 associations —<br/>overlay complete, no drift"]
  end
```

Nothing derived was hand-fed: the tables, their 75 new columns and all 8 keys come out
of `machine_inventory_schema.go`, `app_attest_archive.go`, `app_attest_receipts.go`,
`app_attest_shadow.go` and `app_attest_enrollment.go` as written. What the overlay adds
is the two names for state the compiler can only see as fields (`mem.app_attest_shadow`
covers the per-connection shadow session plus its verifier and storage semaphores;
`mem.machine_inventory` covers the per-connection identity session and its write
permits) and the sentence explaining each node. `coordinator/appattest` is skipped by
package default the way every other pure-verification package is — `appattest.Key` and
`appattest.Receipt` are values parsed out of a proof, not shared state a route mutates.

The timeline was rebuilt over the merged tree: 397 snapshots, head `5b3251167` — the
merge commit itself, since the walk only stops at commits that touched the service — and
the two new clusters of tables appear there rather than being backfilled. The page says
so itself when the two differ, naming both revisions.

## Verification

- `make -C tools/systemmap test` — fixture pipeline over a real miniature service
  type-checked exactly as the coordinator is (routes, middleware, gates, dispatch to
  the preferred impl, cycle evidence, derived schema, one test per drift gate proving
  it fires, and one asserting `-check` renders everything yet leaves no file behind).
- `TestCoordinatorMapHasNoDrift` — the real coordinator, in `go test`: the overlay
  must still explain all of it and the map must render. `make -C tools/systemmap
  check` on the merged tree: **overlay complete, no drift**.
- `history/timeline_test.go` — the Go side of the walk: delta encoding replayed both
  ways, the three module eras, overlay retargeting, and the wire-mode rule that must
  match the page's `epMode`.
- `TestPageProvenanceViews`, `TestPageTopologyFingerprint`,
  `TestPageExplorerAffordances`, `TestOverlayLegendFieldsMatchSchema` and
  `TestNodeIdentityOrigin` — the three views and their withholding primitives, the
  DOM-read fingerprint, the label/focus/drawer/identity affordances, the legend
  reading only fields the overlay actually carries, and `namedBy` per fixture node.
- Headless harnesses over the real inventory: the layout reports 224 nodes / 990
  wires / 15 foreign-key arcs / 20 groups / 6 clusters with 0 NaN, 0 containment
  violations, 0 group-disc overlaps; the schema drawer renders every column,
  citation, constraint and statement for all 54 pg-node tables.
- jsdom probe over the published page: 0 script errors, 0 label overlaps at fit and
  through zoom churn, focusing `ext.datadog` shadows 196 of 224 nodes and 963 of 990
  wires, the topology fingerprint is `f6c65e9c` in all three views and under every
  filter, and `Code only` leaks no withheld label or prose into the graph, the
  tables, the boundaries or the panel.
- `make -C tools/systemmap webtest` — the same `*.test.mjs` suite driven against the
  real 108-route map rather than the fixture: **86 pass, 0 fail** (47 before this
  branch; the timeline and foreign-key suites are the difference). The fixture's only
  crowded moment is one 420×320 viewport, so two claims — that the label budget ever
  binds, and that a cluster name ever climbs its rung ladder — are only exercised
  here and in CI.
- `make docs-check` — 289 files OK; the four docs pages this branch changes are restamped.

The README's head-crowding paragraph now records the method alongside the numbers,
because this branch has had two goes at it and neither should be taken on trust. The
original measurement compared head positions read raw off the path `d` attribute —
scene units — against midpoints in screen units, and scored overlap with the foreign-key
head's 9.4px box rather than the access wire's 11.25 × 9px one. `6f03472f8` called the
resulting figures unreproducible and replaced them; `6bf0c1856` called that wrong and
put figures back. Re-measured in one space with `ARROW_PX`, at the zoom that fits the
map in the default 1200×620 viewport, both of those were half right, and here is which
half. At the revision the original numbers were written for (857 wires, k ≈ 0.43) its
head count and its ratio reproduce — **846 of 857** heads touch inside a 9-pixel square,
and midpoints are **5.4×** less crowded, against "846 of the 857" and "five times" —
while its other two figures do not: the average is 101, not 88, and the densest 9-pixel
square holds 36 heads, not 107. So `6f03472f8` was right that two of the four figures
were wrong and wrong to discard the other two, and `6bf0c1856` did not restore them, it
re-measured at that tree's own 975 wires (971 of 975, 122 on average, worst 316).

Today, at 990 wires and k ≈ 0.40, the paragraph reads **986 of the 990 heads overlap
another head, 111 on average**; at the wires' midpoints **985 still touch, but 25 on
average — four and a half times less crowded**. Worst case is one head overlapping 301,
67 at the midpoints. The old "156 in a 9px square" clause is dropped rather than
reasserted, since that measurement gives 36 — the same 36 the original revision gives,
which is the clearest sign that clause was never a measurement of this.

## Known gaps

Named rather than hidden, none of them blocking:

- the walk's outer layer — `history.Build`, `Commits`, `Snapshot` — has no test; what
  `timeline_test.go` covers is the encoding, the module eras, the overlay retargeting
  and the wire-mode rule, all of which are the parts a change is likely to break.
- `encode` exists twice, in Go and in the page's replay, joined only by the fixture
  the timeline webtest builds. A third representation would need both edited.
- `Commits`' pathspec hardcodes `internal` and `go.mod` beside the service directory,
  which is right for the coordinator's three module eras and would need a line for a
  service whose history moved differently.
- `render/page.js` is 2,660 lines. The page's script is already three files by
  contract (`page.js`, `page.timeline.js`, `page.boot.js`); the first is the one still
  worth splitting.
- **node prose is gated on presence, not content.** `assemble.go:468-471` requires a
  `depDocs` entry for every drawn node, but the entry is a `json.RawMessage`
  (`config/config.go:29`) the gate never opens — so a sentence that contradicts the
  source it describes passes. That is not hypothetical: review of this branch found
  seven of them among the App Attest and machine-identity nodes (who calls
  `ObserveMachine` and how long a capture waits for a permit, the enrollment table's
  primary key, which path writes a receipt first, which session index is partial, the
  two semaphores' sizes), all fixed in `51b3e962f`. Route prose is hashed against the
  facts it describes; node prose is not.
- **the timeline draws 66 of the 116 nodes as present at every one of the 397 commits.**
  `assemble.go:363-366` unions every `overlay.json` label key into the node set whether
  the code reached it or not, and the walk retargets today's overlay onto each revision
  — so `mem.app_attest_shadow` and `mem.machine_inventory`, which #995 created in
  September 2026, are drawn at the coordinator's first commit. Routes are not affected
  (they come only from what was reached: 7 of 108 span the whole axis, which is a fact
  about `/health` and its neighbours rather than an artefact). The slider is honest
  about routes, wires and access modes, and about *arrival* for the 50 nodes it does
  encode; it overstates how long the other 66 have existed. The fix is to derive
  presence from reach rather than from the label table, which changes what the plain
  map draws too — hence a follow-up, not a patch inside this PR.
- **no test renders a real `-history` page.** `timeline.test.mjs` drives a synthetic
  timeline built from the map's own inventory, and `history/timeline_test.go` covers the
  encoding, module eras, retargeting and wire-mode rule. Nothing exercises the two
  meeting over a real walk, which is where `history.json`'s head disagreeing with the
  map's revision, or a 397-snapshot replay's cost, would show up. Both were checked by
  hand here.
- **the SQL reader has gaps this branch did not introduce and did not close.**
  `maskKeywordCalls` counts parentheses without regard to string literals, so
  `substring(x, ')')` ends its mask early (`extract/sql.go:252-267`); `reFrom` and
  `reJoin` have no leading `\b`, so a word merely *ending* in `from` or `join` scans as
  the keyword; `reUsing` sits in `writeMatchers`, so a `DELETE … USING t` read is scored
  as a write to `t`; and a comma from-list (`FROM a, b`) yields only `a`. None
  hides a table in the coordinator today — `make check` is clean — and each is a
  separate small change with its own fixture, so they are named here rather than
  bundled in.

## Human steps left

Enabling Pages (Settings → Pages → Source = "GitHub Actions") is a one-time manual
step. Until then the CI artifact is the way to read the map.

Publishing the history slider is a deliberate choice, not a default: the walk costs
minutes, so `pages-systemmap.yml` builds the plain map unless asked for the slider
(`make -C tools/systemmap history`, or `HISTORY_ARGS="-every 25"` for a sampled
look).

Three checks were red for reasons outside this diff, all three reproducible on the
branch before its last commits:

- `Threat Model Review` — `anthropic.AuthenticationError: 401 API key is invalid` in
  `.github/scripts/threat-model-review.py:187`. The repository's `ANTHROPIC_API_KEY`
  secret needs rotating; the job never reads the diff.
- `Vercel – d-inference` and `Vercel – d-inference-landing` — "Deployment failed."
  in 0s with `vercel.com/docs/accounts/team-members-and-roles` as the only link,
  i.e. a Vercel account permission, not a build. Both were already failing on
  `a34e8994c` and `f8dcd5aef`; `console-ui-dev` passes on all three commits.


ethenotethan:
## The overlay is two layers, and only one of them is verifiable

Worth naming explicitly before this merges, because it is the one maintenance regression this PR introduces.

`overlay.json` reads like one curated file, but it is two different things with two different failure modes. Measured on the current file (3,174 lines / 188 KB):

| Layer | Share | Contents | What a wrong entry does | Gate |
|---|---|---|---|---|
| **Structural** | 13% (22.7 KB) | `deps.{fields,types,functions,hosts,endpoints,messages,traverse,inherit,packageDefault,strict,preferImpl}`, `clusters`, `categories`, `namespaces`, `authRules`, `gates`, `gateDepth` | **moves an edge** — changes which node exists, which cluster it sits in, how far the walk goes, what auth class a route is filed under | bidirectional: unmapped source is drift, and an overlay entry whose source is gone is drift |
| **Enrichment** | 87% (145.9 KB) | `depDocs` (96.8 KB, 8 prose fields × 92 nodes), `routes` descriptions (17.9 KB), `credentials` + `roles` (22.3 KB), `categoryDocs`, `cacheSemantics`, `labels` | **changes nothing about the graph** — only whether a reader can understand it | presence only |

The structural layer cannot rot: it is checked against the type-checked source in both directions, which is the whole point of the drift gates. The enrichment layer is where the regression lives.

### The regression, precisely

The gates ask *"is there prose for this node?"*, never *"is this prose still true?"*:

- `assemble.go:439` requires `depDocs[<node>]` to exist; any non-empty object passes. Rewrite a cache's invalidation policy and its eight prose fields keep describing the old one, forever, with a green build.
- `cacheSemantics` is validated only for *node existence* (`assemble.go:450`) — the TTL and invalidation strings themselves are unchecked.
- `roles` and `credentials` are `[]json.RawMessage` (`config.go:30-31`) passed straight through to the graph (`assemble.go:112-113`). **Zero gates.** 22 KB describing the auth model — issuers, lifetimes, client/server storage, custody — could describe a credential that no longer exists and nothing in CI would notice.

So the honest statement of what this PR ships: a base layer that *cannot* be stale by construction, wrapped in an enrichment layer that has no staleness detection at all. The base layer's guarantee does not extend to the prose, and today nothing tells a reader which sentences a compiler stands behind. The derived/curated split in the README is that distinction, but it is documentation, not enforcement.

### Where synthesis is required, and where it must not be allowed

```mermaid
flowchart TB
  subgraph base["Deterministic base layer — go/packages + go/types"]
    B1["routes · middleware · gates<br/>reached state · SQL tables · modes<br/>table definitions · citations"]
  end
  subgraph struct["Structural curation — human-reviewed, gate-checked both ways"]
    S1["node identity · cluster membership<br/>traversal scope · auth classes"]
  end
  subgraph enrich["Enrichment — synthesizable, presence-checked only"]
    E1["depDocs · route descriptions<br/>roles/credentials · cacheSemantics · labels"]
  end
  B1 --> S1 --> E1
  E1 -. "must never feed back" .-> B1
```

The rule that falls out: **LLM synthesis may only occupy fields that no gate depends on and no topology derives from.** Enrichment is safe to generate because a wrong sentence is a comprehension bug; structural curation is not, because a wrong `deps.types` entry silently relocates an edge and the build stays green — the map would be confidently wrong rather than incomplete, which is strictly worse than having no map.

Two things make that rule enforceable rather than aspirational:

1. **`TestProseDoesNotAffectTopology`** — build the graph twice, once with every enrichment section stripped, and assert the nodes, edges, modes and citations are byte-identical. That turns "prose can't change the picture" from a convention into a test, and it is the precondition for letting anything non-deterministic write those fields.
2. **Fingerprint the prose against the facts it describes.** For each node, hash the derived facts the prose is *about*: its category, cluster, the set of routes that reach it, their merged modes, and the symbols in its citations. Store that hash beside the prose. When the fingerprint changes, the prose is flagged **stale — needs re-review**, and `-check` fails. That is deterministic staleness detection over non-deterministic content: you cannot verify a sentence, but you can prove that the thing it described moved.

With (2) in place, PR-time synthesis becomes cheap and bounded: only nodes whose fingerprint changed need regenerating, so a typical PR re-synthesizes one or two `depDocs` blocks instead of 92, and the diff a human reviews is exactly the prose whose subject actually changed. It also fixes the review ergonomics — the reviewer reads a handful of regenerated paragraphs, not a 3,174-line JSON file.

### What the semantic shape is actually good for

This is the part worth building next, and it does not need the prose at all: `inventory.json` is a machine-readable statement of the architecture, so architectural rules stop being prose in `docs/` and become assertions that fail CI. A policy file plus a walk of the graph, run in the existing `System Map` job:

- **Unauthenticated blast radius.** A route whose derived auth class is `Public` may reach only an allowlisted set of nodes. Someone hanging a new handler off an unauthenticated path that reaches the ledger fails the build instead of a review.
- **Write authority.** No `W` on `pg.balances` / `pg.ledger_entries` from a route outside the declared billing auth classes.
- **Gate coupling.** Every route reaching admin-classified state must carry the `isAdminAuthorized` gate — derived, not asserted by a comment.
- **Boundary crossings.** No coordinator endpoint reaches the provider or sidecar cluster except through its declared surfaces; a new `ext.*` host reachable from an unauthenticated route is a finding by construction.
- **Orphan state.** A new `pg.*` table no endpoint reaches must be explicitly declared background-only (this is informational in `report.md` today; it should be a rule).

Two notes on making that trustworthy:

**The base layer has to be right first.** This PR just fixed a case where it wasn't: a write reaching its node through an unnamed chain link (`c.entries[k] = v` behind a `@skip`ped field, node attached to the receiver's type) was published as a **read**. Sixteen associations were understated, including every `ext.datadog` edge and `mem.rpm`; `mem.read_cache` under Administration was `RW` when the admin path only invalidates it. Any invariant keyed on `W` would have been silently wrong on all sixteen. Fixed in `extract/walk.go` with a fixture regression (`TestFixtureWriteThroughSkippedField`) — but the lesson is that invariants are only as good as the mode derivation underneath them, so each new rule should land with a fixture that proves it fires.

**Diff the graph, not the page.** Since the map is built per PR anyway, comparing this PR's `inventory.json` against master's yields exactly the delta worth reviewing: new edges, edges whose mode widened, new hosts, new tables, auth classes that changed. That belongs in a PR comment — it is deterministic, it is short, and it answers the "how do I review a generated artifact" question better than serving the HTML does.

Sequencing I'd suggest: `TestProseDoesNotAffectTopology` and the prose/structure file split first (both small, both make everything after them safe), then the fingerprint staleness gate, then PR-time synthesis for changed fingerprints only, then the invariant framework and the inventory diff comment. Happy to take these in that order — the first two can land on this PR, the rest are better as follow-ups so this one stays reviewable.

Issue #689, opened 2026-08-24
Provider hangs on any Mac with a display: HardwareDetector/attestation pipe-buffer deadlock


agustif:
## Summary

Every command that loads the runtime snapshot (`login`, `start`, `status`, `doctor`, `verify`, `models …`) hangs forever on a Mac **with a display**. Root cause: `HardwareDetector` and the attestation/security posture code spawn subprocesses (`system_profiler`, `ioreg`, `diskutil`, `csrutil`, `codesign`) with a `Pipe` on stdout and call `Process.waitUntilExit()` **before draining the pipe**.

macOS pipe buffers start at **~512 bytes** (verified empirically). The child fills the buffer, blocks in `write()`, and never exits; the parent blocks in `waitUntilExit()` waiting for a child that is waiting for the parent to read. Classic deadlock.

`ProviderCore/Hardware/HardwareDetector.swift → detectGPUInfo()` runs `system_profiler SPDisplaysDataType -json`, whose output is **1260 bytes** on this M3 Max with one built-in Liquid Retina XDR display — already over 512. **Headless Macs escape the bug** because with no display the `SPDisplaysDataType` output stays under 512 bytes, which is likely why it passes fleet/CI testing but breaks on laptops and desktops.

## Environment
- darkbloom 0.8.10 (macos-arm64), M3 Max / 128 GB, macOS 26.6
- Reproduces on any Mac with an attached display.

## It also blocks earning (confirmed)

Even after working around hardware detection, the serial-number readers deadlock the same way (`AttestationBuilder.detectSerialNumberFromIOReg`, `Auth/Enrollment.swift` — `ioreg -c IOPlatformExpertDevice -d 2`). With those subprocesses reaped to force completion, `verify` reports:

```
[PASS] binary hash: b6e60f58…          (genuine, unmodified binary)
[PASS] mdm enrollment: Darkbloom profile installed
[PASS] coordinator health: https://api.darkbloom.dev
[WARN] coordinator trust: local serial number unavailable   ← attestation can't complete
```

So the deadlock, not signing or enrollment, is what prevents a healthy provider from earning.

## Repro (standalone, no provider needed)
```c
int p[2]; pipe(p); pid_t c = fork();
if (c==0){ dup2(p[1],1); dup2(p[1],2); close(p[0]);
  execl("/usr/sbin/system_profiler","system_profiler","SPDisplaysDataType","-json",0); }
close(p[1]); alarm(15); int st;
[...]

PR #837, opened 2026-09-05
fix(provider-swift): bound the doctor/verify contention and sleep probes (#811)

### LINKED ISSUE #811: [bug] doctor and verify deadlock when ps output exceeds pipe capacity
### Component

provider-swift (Swift CLI)

### What happened?

On Darkbloom v0.8.15, both `darkbloom doctor` and `darkbloom verify` produce no output and never exit on a Mac with a large process table.

A live sample shows the CLI blocked here:

```text
Doctor.run()
  LocalContentionSnapshot.live()
    LocalContentionSnapshot.runCapture(_:args:)
      -[NSConcreteTask waitUntilExit]
```

The remaining child is:

```text
/bin/ps -axo comm=
```

The cause appears to be the ordering in `CompetingInferenceDiagnostics.swift`: `runCapture` calls `waitUntilExit()` before reading stdout from its `Pipe`. When `ps` emits more than the pipe capacity, the child blocks writing while the parent blocks waiting for the child to exit.

On the affected machine:

```text
$ /bin/ps -axo comm= | wc -c
78689
```

The child pipe reported a capacity of 65,536 bytes. Running the same `ps` command with concurrently drained stdout exits immediately.

The same implementation is still present on current `master`:

https://github.com/Layr-Labs/d-inference/blob/master/provider-swift/Sources/darkbloom/Diagnostics/CompetingInferenceDiagnostics.swift#L43-L57

### Expected behavior

`darkbloom doctor` and `darkbloom verify` should complete regardless of process-table output size. Best-effort contention diagnostics should also have a bounded runtime.

### Steps to reproduce

1. Run enough processes for `/bin/ps -axo comm=` to emit more than 65,536 bytes.
2. Confirm with `/bin/ps -axo comm= | wc -c`.
3. Run `darkbloom doctor` or `darkbloom verify`.
4. Observe that the command produces no output and remains blocked until interrupted.
5. Sample the CLI process and observe `LocalContentionSnapshot.runCapture` blocked in `waitUntilExit`, with `/bin/ps -axo comm=` as its child.

### Environment

- Darkbloom: v0.8.15, currently reported as the latest release
- macOS: 26
- Hardware: Apple M5 Pro, 64 GB RAM
- Process count during reproduction: 1,041
- Reproduced repeatedly with a 5-second timeout; both commands were still blocked each time

### Suggested fix and regression coverage

Drain stdout while the subprocess is running instead of waiting for exit first, and preferably add a timeout because this is a best-effort diagnostic.

A regression test can run a helper that writes more than 65,536 bytes to stdout before exiting, then assert that `runCapture` returns the complete output within a short timeout.

### Confirmations

- [x] I searched existing issues and this is not a duplicate.

chethanuk:
## Summary

`darkbloom doctor` and `darkbloom verify` hang forever on a Mac with a large process
table. `LocalContentionSnapshot.runCapture` and `DoctorRunner.systemSleepPrevented`
each hand their child a `Pipe`, call `waitUntilExit()`, and only read the pipe
afterwards. Once the child writes past the ~64 KiB Darwin pipe buffer it blocks in
`write(2)` and can never exit, so the parent waits on a child that is waiting on
the parent. On the reporter's Mac, `/bin/ps -axo comm=` emitted 78,689 bytes
against a 65,536-byte pipe.

Both probes now go through `FanProcessRunner.run`, which drains the pipe on a
readability handler while the child runs and enforces a SIGTERM-then-SIGKILL
deadline. Reordering the read above the wait (as #786 does) isn't sufficient on
its own: `readDataToEndOfFile()` still blocks until the child closes the stream,
so a child that never closes stdout keeps wedging the caller. The issue also asks
for a *bounded* runtime, which is the deadline, not just the drain — 5 seconds is
the budget for probes already documented as degrading to empty on failure.

`run` gains `discardStandardError` and `maximumOutputBytes`, both defaulted to
current behavior so the nine existing call sites are unchanged. `runCapture` needs
the first because the runner otherwise folds stderr into stdout, and an `lsof`
warning naming `~/.ollama` would text-match as a running Ollama — the false
positive the probe is documented to avoid. It needs the second because the
runner's 64 KiB cap is sized for fan install logs: `ps` was 78,689 bytes on the
reporter's Mac, so keeping that cap would drop every process name past it and
report "no competitors" on exactly the machines this fixes.

### Before / after

```mermaid
flowchart LR
  subgraph Before
    A1["doctor / verify"] --> B1["runCapture / systemSleepPrevented<br/>own Process + Pipe"]
    B1 --> C1["waitUntilExit()"]
    C1 --> D1["child blocked in write(2)<br/>pipe full at 64 KiB"]
    D1 --> E1["hang forever, no output"]
  end
  subgraph After
    A2["doctor / verify"] --> B2["runCapture / systemSleepPrevented"]
    B2 --> C2["FanProcessRunner.run<br/>timeout: 5, stderr discarded, 1 MiB cap"]
    C2 --> D2["readabilityHandler drains<br/>while the child runs"]
    D2 --> F2["exit -> full output"]
    C2 --> G2["deadline -> SIGTERM/SIGKILL<br/>-> status -1 -> nil"]
    F2 --> H2["check reported"]
    G2 --> I2["check degrades to empty"]
  end
```

### Relationship to the other open PRs

- Supersedes #786: same function, but that PR only reorders the read above the
  wait and adds no deadline, so a child that never closes stdout still hangs it.
- Overlaps #690's `DoctorRunner` hunk (a much larger `ProviderCore` sweep fixing
  #689). If #690 lands first, drop this PR's `DoctorRunner.swift` hunk and keep
  the rest — the two are compatible. #690 doesn't touch
  `CompetingInferenceDiagnostics.swift`, so it doesn't close this issue on its own.

## Linked issue

Closes #811

## Test plan

`Tests/DarkbloomCLITests/FanProcessRunnerTests.swift` adds seven cases:

- [x] `runCaptureSurvivesChattyChild` — `sh -c 'yes darkbloom | head -c 200000'`,
      200,000 bytes to clear the 64 KiB pipe buffer.
- [x] `runCaptureKeepsOutputPastTheDefaultCap` — a hint emitted past 64 KiB still
      reaches the caller.
- [x] `defaultOutputCapIsUnchanged` — the other nine callers keep the 64 KiB bound.
- [x] `runCaptureReturnsNilWhenTheChildNeverExits` — `sh -c 'sleep 120'`; draining
      alone doesn't fix this, only the 5s deadline does.
- [x] `stderrIsDiscarded` / `stderrIsMergedByDefault` — both sides of the new
      flag, pinning the nine existing callers' behavior.
- [x] `spawnFailureIsDistinguishable` — a missing executable stays `status == -1`
      and `nil`, not empty output.

Run on macOS 26.6.2 / arm64 / Swift 6.3.3:

```
$ swift build --build-tests
Build complete! (753.13s)

$ swift test --filter FanProcessRunnerTests
✔ Suite "Fan process runner" passed after 5.215 seconds.
✔ Test run with 7 tests in 1 suite passed after 5.215 seconds.
```

Not run: the rest of `swift test`. Those suites need the source-matched
`mlx.metallib` that `test-provider` stages, which I skipped — no claim is made
about them.

The suite is `.serialized`. `FanProcessRunner` drains through
`FileHandle.readabilityHandler`, and Foundation runs every file handle's handler on
one process-wide serial queue; run in parallel these cases jam that queue for each
other (a trivial `echo` child took 20s and the two chatty cases never returned,
because clearing the handler at the end of `run` waits on that same queue).
Nothing in production runs two of these at once — `doctor` probes sequentially.

## Components touched

- [ ] coordinator (Go)
- [ ] provider (Rust, legacy)
- [ ] console-ui (Next.js)
- [ ] enclave (Swift)
- [ ] infra / CI / release
- [ ] docs

## Notes for reviewers

- `runCapture` loses `private` so `@testable import darkbloom` can reach it (#786
  does the same).
- Output is decoded lossily, where the old code returned `nil` on invalid UTF-8.
  The probe raises the runner's byte cap to 1 MiB so a large `ps` is not truncated
  into a missed WARN; every other caller keeps 64 KiB.
- `runCapture` maps only `status == -1` (spawn failure or deadline) to `nil`; a
  non-zero exit still yields text, matching the old code, which never inspected
  the status. `systemSleepPrevented` keeps its stricter `== 0` guard.
- `systemSleepPrevented` stays `private` with a hardcoded path and has no test of
  its own; the shared runner is what the new cases exercise.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/837"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791219676&installation_model_id=21733&pr_number=837&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F837&signature=b96a4b0d177891e315e92eb9752179e7e6ad194ce96713b8ab7a98a85b9a3483"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/provider-swift/Sources/darkbloom/Diagnostics/CompetingInferenceDiagnostics.swift
+++ b/provider-swift/Sources/darkbloom/Diagnostics/CompetingInferenceDiagnostics.swift
@@ -40,21 +40,40 @@ struct LocalContentionSnapshot: Equatable, Sendable {
         )
     }
 
-    private static func runCapture(_ path: String, args: [String]) -> String? {
-        let p = Process()
-        p.executableURL = URL(fileURLWithPath: path)
-        p.arguments = args
-        let pipe = Pipe()
-        p.standardOutput = pipe
-        p.standardError = FileHandle.nullDevice
-        do {
-            try p.run()
-            p.waitUntilExit()
-        } catch {
-            return nil
-        }
-        let data = pipe.fileHandleForReading.readDataToEndOfFile()
-        return String(data: data, encoding: .utf8)
+    /// Runs `path` and returns its stdout, or nil if the probe could not
+    /// produce any.
+    ///
+    /// This used to spawn the child itself and call `waitUntilExit()` before
+    /// reading the pipe, which deadlocks the moment the child outgrows the
+    /// ~64 KiB Darwin pipe buffer: the child blocks in `write(2)` with a full
+    /// pipe while the parent waits for it to exit. `ps -axo comm=` clears
+    /// 64 KiB on a busy Mac, so `doctor` and `verify` hung outright (#811).
+    ///
+    /// `FanProcessRunner` drains concurrently *and* enforces a deadline, which
+    /// is the part draining alone does not give: a child that never closes
+    /// stdout still holds `readDataToEndOfFile()` forever. Five seconds is the
+    /// budget for a probe documented above as degrading to empty, so a slow
+    /// `lsof` now costs a missing hint rather than a hung command.
+    static func runCapture(_ path: String, args: [String]) -> String? {
+        let result = FanProcessRunner.run(
+            path,
+            arguments: args,
+            timeout: 5,
+            discardStandardError: true,
+            // The runner's 64 KiB default is sized for fan install logs. `ps`
+            // was 78,689 bytes on the reporter's Mac, and a busy machine is
+            // exactly where this probe has to work, so capping at 64 KiB would
+            // drop the hints past the cap and report "no competitors" for a
+            // running Ollama. 1 MiB is ~13x the reported table and still bounds
+            // a runaway child.
+            maximumOutputBytes: 1024 * 1024
+        )
+        // -1 is a spawn failure or the deadline. A non-zero *exit* still yields
+        // usable text and the previous implementation never checked the status
+        // either, so `lsof` exiting 1 with no listener keeps degrading to no
+        // hint. Output is decoded lossily where `String(data:encoding:)` used
+        // to return nil on invalid UTF-8.
+        return result.status == -1 ? nil : result.output
     }
 }
 
--- a/provider-swift/Sources/darkbloom/Diagnostics/DoctorRunner.swift
+++ b/provider-swift/Sources/darkbloom/Diagnostics/DoctorRunner.swift
@@ -245,17 +245,18 @@ enum DoctorRunner {
     /// via `pmset -g assertions`. Informational only (the provider
     /// self-caffeinates while serving), so nil/UNKNOWN on any failure is fine.
     private static func systemSleepPrevented() -> Bool? {
-        let p = Process()
-        p.executableURL = URL(fileURLWithPath: "/usr/bin/pmset")
-        p.arguments = ["-g", "assertions"]
-        let out = Pipe()
-        p.standardOutput = out
-        p.standardError = Pipe()
-        guard (try? p.run()) != nil else { return nil }
-        p.waitUntilExit()
-        guard p.terminationStatus == 0 else { return nil }
-        let data = out.fileHandleForReading.readDataToEndOfFile()
-        guard let text = String(data: data, encoding: .utf8) else { return nil }
+        // Same pipe-buffer deadlock as #811's contention probe: this waited on
+        // the child before draining stdout, and its stderr `Pipe()` was never
+        // read at all. Route it through the bounded runner so a chatty or wedged
+        // `pmset` degrades to UNKNOWN instead of hanging `doctor`.
+        let result = FanProcessRunner.run(
+            "/usr/bin/pmset",
+            arguments: ["-g", "assertions"],
+            timeout: 5,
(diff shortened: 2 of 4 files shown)


Gajesh2007:
@codex review this

PR #896, opened 2026-09-11
feat: lifetime earnings per machine on the provider earnings page

EigenMustafa:
## What

Adds a lifetime **Earnings by Machine** breakdown to the provider earnings page (`/providers/earnings`). Providers running multiple Macs had no way to see which machine earned what: the earnings page aggregates "across all linked provider nodes," and the fleet dashboard's machine cards deliberately show no dollars. Asked for by a provider (Mark M) in #support, 2026-09-11.

## Behavior — before / after

```mermaid
flowchart LR
  subgraph Before
    A1[Provider asks: which Mac earned what?] --> B1[Earnings page: account totals only]
    A1 --> C1[Fleet dashboard: tokens/reputation, no dollars]
    B1 --> D1[No answer in-product — community menu-bar app or nothing]
  end
  subgraph After
    A2[Provider opens /providers/earnings] --> B2[Earnings by Machine table:<br/>lifetime earned + jobs per machine,<br/>sorted highest first]
  end
```

## Code — before / after

```mermaid
flowchart LR
  subgraph Before
    E1[EarningsContent] --> P1["/api/me/earnings proxy"] --> H1[handleAccountEarnings] --> S1[GetAccountEarnings + Summary]
  end
  subgraph After
    E2[EarningsContent] --> X2[ByMachineList — new]
    E2 --> P2["/api/me/earnings proxy (unchanged)"] --> H2[handleAccountEarnings<br/>+ by_provider array] --> S2[GetAccountEarningsByProvider — new<br/>postgres GROUP BY + memory impl]
    H2 --> R2[ListProvidersByAccount<br/>hardware labels via persisted PublicKey]
  end
```

## Design notes (what we learned building this)

- **Group by `provider_key`, never `provider_id`.** The earnings rows carry both; `provider_id` is the per-connection UUID that rotates on restart and would scatter one machine across dozens of buckets. `provider_key` is the persisted X25519 hardware identity — the store documents it as the sessions↔earnings unifier.
- **Aggregation must be server-side.** The account-earnings endpoint returns lifetime totals but caps history rows (default 100, max 1000). Grouping rows client-side would silently undercount per-machine totals on any active fleet. The new store method is a single `GROUP BY` over `provider_earnings`, reusing the endpoint's existing 20s per-account cache.
- **`base_reward` rows count in money, not in jobs/tokens** — matching the existing `earnings_summary` semantics.
- **Machine labels resolve server-side** from the account's provider records (`PublicKey` is persisted, so offline machines still get chip/memory names).
- **Every row shows an always-visible machine ID** ("ID …aaaaaa") — the tail of the machine's stable hardware key. Without it, two identical Macs ("Apple M4 Max · 64 GB" twice) would be indistinguishable; showing it always, rather than only on clashes, keeps the meaning obvious. The nicer long-term fix is user-nameable machines — follow-up below.
- **The "Other" bucket**: rows whose `provider_key` is empty (the column has existed since table creation with `DEFAULT ''`, so this covers any period before the crediting path populated it). Those rows cannot be attributed retroactively — nothing durable maps them to hardware — but the dollars must still appear somewhere. The bucket renders only if such rows exist. **Open question for someone with replica access:** `SELECT COUNT(*), MIN(created_at), MAX(created_at) FROM provider_earnings WHERE provider_key = '';` — if 0, per-machine history is complete back to day one and "Other" never renders.
- **No "Last active" column, deliberately.** The obvious source (`MAX(created_at)` of earnings rows) means "last *earned*", which reads wrongly stale for an online-but-idle machine. Cut rather than mislead; the fleet dashboard is the place for liveness.

## Follow-ups (not in this PR)

- **Per-window breakdowns (24h / 7d)** — this PR ships lifetime only; windows multiply the query and UI and deserve their own pass.
- **User-nameable machines** — would let the ID line become a friendly name.
- **Echo an "earned" stat onto fleet dashboard machine cards** — currently deliberately dollar-free; product call.
- **Product/IA note (from support):** Fleet ("Your fleet") and Earnings ("Your earnings") pages now both carry per-machine information with different slices (operational vs financial). The split is starting to feel inefficient — worth rethinking whether these merge into one per-machine view. Flagging for discussion rather than acting unilaterally.

## Tests

- `coordinator/store/account_earnings_by_provider_test.go` — grouping across rotating connection IDs, base-reward money-not-jobs, account isolation, sort order, empty account.
- `coordinator/api/billing_earnings_test.go` — new `TestAccountEarningsGroupsByMachine`: full handler path, hardware enrichment, no-record machine, legacy empty-key bucket.
- `console-ui/__tests__/by-machine-list.test.tsx` — labels, identical-hardware disambiguation, Other bucket, empty state (7 tests).

`go build` + `go vet` + `gofmt` clean; `npm run build` + eslint clean (0 errors). Full `./store/` suite passes. Local `./api/` run has 6 pre-existing failures that reproduce identically on master — environmental only (sparse checkout missing `scripts/`+`provider-swift/` for two file-reading tests; local Go 1.27 vs pinned 1.25 for four `encoding/json` byte-exactness tests). CI on pinned Go is authoritative.

## Screenshots

> Simulated preview: the real page components rendered by the local dev server with fixture data (the live page needs an authenticated session and a coordinator that already serves `by_provider`). Fixtures deliberately include two identical machines (distinguishable by their IDs) and the "Other" bucket.

### Before
before

### After
after

### DIFF
--- a/console-ui/__tests__/by-machine-list.test.tsx
+++ b/console-ui/__tests__/by-machine-list.test.tsx
@@ -1,0 +1,81 @@
+import { describe, it, expect } from "vitest";
+import { render, screen } from "@testing-library/react";
+import {
+  ByMachineList,
+  machineLabel,
+  displayLabels,
+  type MachineEarnings,
+} from "@/app/providers/earnings/ByMachineList";
+
+const machine = (overrides: Partial<MachineEarnings>): MachineEarnings => ({
+  provider_key: "k".repeat(44),
+  total_micro_usd: 1_500_000,
+  total_usd: "1.500000",
+  job_count: 42,
+  prompt_tokens: 1000,
+  completion_tokens: 500,
+  last_earned_at: "2026-09-10T12:00:00Z",
+  ...overrides,
+});
+
+describe("machineLabel", () => {
+  it("uses chip and memory when known", () => {
+    expect(machineLabel(machine({ chip_name: "Apple M4 Max", memory_gb: 64 }))).toBe(
+      "Apple M4 Max · 64 GB",
+    );
+  });
+
+  it("falls back to a key suffix without hardware info", () => {
+    expect(machineLabel(machine({ provider_key: "abcdef123456" }))).toBe("Machine …123456");
+  });
+
+  it("labels the empty key as Other", () => {
+    expect(machineLabel(machine({ provider_key: "" }))).toBe("Other");
+  });
+});
+
+describe("displayLabels", () => {
+  it("disambiguates identical hardware with key suffixes", () => {
+    const labels = displayLabels([
+      machine({ provider_key: "key-one-aaaaaa", chip_name: "Apple M4 Max", memory_gb: 64 }),
+      machine({ provider_key: "key-two-bbbbbb", chip_name: "Apple M4 Max", memory_gb: 64 }),
+      machine({ provider_key: "key-three-cccccc", chip_name: "Apple M3 Pro", memory_gb: 36 }),
+    ]);
+    expect(labels).toEqual([
+      "Apple M4 Max · 64 GB (…aaaaaa)",
+      "Apple M4 Max · 64 GB (…bbbbbb)",
+      "Apple M3 Pro · 36 GB",
+    ]);
+  });
+});
+
+describe("ByMachineList", () => {
+  it("renders one row per machine with earnings and jobs", () => {
+    render(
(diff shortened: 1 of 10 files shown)


0xkydo (approved):

0xkydo:
@codex review this, check where files are where they are suppose to

PR #924, opened 2026-09-11
Share provider release signing checks and safely encode tag metadata

Gajesh2007:
The release workflow duplicated weaker versions of the signing-validation checks: it accepted unrelated app identifiers, ignored the macOS identifier key, and only warned about missing profile expiry. Both workflows now use the existing parsed profile and signed CLI entitlement validators. Release registration also reads tag text as environment data when encoding JSON, so quotes, backslashes and multiline notes retain their values.

This removes 99 lines from the release workflow while preserving its build, signature verification, runtime smoke, notarization and publication order. Invalid profile identities and missing expiry now stop the release path before signing; both debug-task entitlement spellings are checked on the signed CLI. No release was dispatched or published.

```mermaid
flowchart LR
  subgraph Before
    P[Decoded profile] --> W[Inline release validator: weaker identity and expiry checks]
    S[Signed CLI] --> G[Separate grep and PlistBuddy checks]
    T[Tag text] --> I[Interpolate into Python source]
    I --> F[Quoted notes can break JSON generation]
  end
  subgraph After
    P2[Decoded profile] --> V[provider-signing-validation.py profile]
    S2[Signed CLI] --> C[provider-signing-validation.py cli-entitlements]
    V --> N[Existing signing and notarization sequence]
    C --> N
    T2[Tag text] --> E[Environment values and json.dumps]
    E --> R[Valid registration JSON with original note text]
  end
```

Validation: nine signing-helper tests and two workflow fixture tests pass. The workflow fixtures exercise seven profile scenarios and two tag-text scenarios; five assertions fail against master and pass with this change. They execute only profile/payload steps with local CMS fixtures, with no signing, model execution, network request or publication. YAML and changed shell steps parse; docs lint passes (278 pages). Release Integrity CI now runs these checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,8 @@ jobs:
         run: |
           ./scripts/check-release-version.sh
           python3 scripts/test-provider-release-resolution.py
+          python3 scripts/test-provider-signing-validation.py
+          python3 scripts/test_release_workflow.py
           ./scripts/sync-install-embed.sh check
           ./scripts/test-prod-env-refresh.sh
       - name: Test startup observer against local stubs
--- a/.github/workflows/release-swift.yml
+++ b/.github/workflows/release-swift.yml
@@ -345,69 +345,7 @@ jobs:
           # passes codesign but fails at runtime → silent attestation break.
           security cms -D -i /tmp/embedded.provisionprofile > /tmp/profile.plist
 
-          python3 - <<'PY'
-          import plistlib, sys
-          from datetime import datetime, timezone
-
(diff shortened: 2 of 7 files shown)


Gajesh2007:
Consolidated into #1096, the combined tooling, benchmark and E2E refactor PR. Your latest head `35d8c5ae64c4b95edfd70792cdc5ca014d4cd17a` is preserved in aggregate `9d2160bb370e8a0dbf3810e61a5e48fe66345651`, along with the relevant review fixes. Continue review and merge at https://github.com/Layr-Labs/d-inference/pull/1096 (after provider #1094 and web #1095). Closing this duplicate merge path; the source branch and review history are retained.

PR #931, opened 2026-09-12
fix(coordinator): preserve model aliases after validation read failures

Gajesh2007:
Transient model-registry reads currently fall through as if the alias or concrete ID did not exist. A following write can succeed, dropping rollout history, replacing an OpenRouter-only alias through the standard endpoint, creating an alias over a concrete model ID without takeover, or registering a concrete build over an existing alias.

Both alias upsert endpoints and concrete model registration now stop before writing when namespace validation fails. Registration also stops before fetching artifacts. Standard upserts also require a successful prior-alias read, and distinguish failed member reads (500) from confirmed missing builds (400). The duplicated desired/previous validation follows one ordered loop; the error-swallowing `priorAlias` helper is removed. Existing successful updates, takeover rules, endpoint ownership and missing-ID behavior remain covered by the alias tests.

```mermaid
flowchart LR
  subgraph Before
    A[Alias upsert or model registration] --> B[Namespace or priorAlias read fails]
    B --> C[Assume missing]
    C --> D[Alias or concrete model write succeeds]
    D --> E[200 with lost lineage or conflicting ownership]
  end
  subgraph After
    F[Alias upsert or model registration] --> G[Check namespace and prior alias errors]
    G -->|Store failure| H[500 before any write]
    G -->|Successful read or confirmed namespace miss| I[Validate ordered desired and previous members]
    I -->|Store failure| H
    I -->|Missing member| J[400]
    I -->|Valid| K[Existing ownership checks then appropriate write]
  end
```

Validation:
- Added seven fault-injection cases with an available backing store. On the baseline, four return 200 and mutate persisted aliases; two return a misleading 400. All now return 500. The reverse registration regression also previously fetched two artifacts and created the colliding concrete model; it now preserves the namespace and performs no fetch.
- `GOTOOLCHAIN=go1.25.0 go test -race ./api -run 'Test.*(Alias|DesiredModels)' -count=1` passed (5.406s), including existing creation, takeover, missing-member, lineage, OpenRouter and desired-model publication cases.
- Additional registration/alias/desired-model race coverage passed after the reverse-guard fix (3.674s).
- `scripts/docs-check.sh --all` passed (278 pages); current Markdown has no references to the removed helper.
- `GOTOOLCHAIN=go1.25.0 go test ./...` passed for the full coordinator (API package 194.207s); CI also validates the published branch.

This change addresses validation-read failures. It does not change alias deletion ordering or claim transactional serialization across coordinator processes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased — alias validation recovery
+
+- Return a server error before changing model aliases when namespace, rollout-history or build validation reads fail. Preserve existing alias ownership and retired-build lineage through transient store errors.
+
 ## Release candidate v0.9.2 — Gemma QAT caching, adaptive MTP and Nemotron Lightning (not shipped; 2026-09-10)
 
 Source changes since `v0.9.1`. Provider changes require a new signed bundle.
--- a/coordinator/api/model_alias_handlers.go
+++ b/coordinator/api/model_alias_handlers.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"encoding/json"
+	"errors"
 	"net/http"
 	"strings"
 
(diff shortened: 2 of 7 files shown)


Gajesh2007:
@codex review

Gajesh2007:
@codex review — also includes reverse model registration namespace guard and its reproduced write-through regression at source 88e9baeed.

Gajesh2007:
Consolidated into #1056 for review and merge after #1094, #1095 and #1096. #1056 contains this branch's exact head (`a6de519a164a9d6b0c04229d26e7d68369032ad1`), including its reviewed fixes, with implementations and regressions reconciled into the coordinator subsystem owners. Closing the component PR to collapse the stack; its branch, commits and review history are retained. Follow #1056 for combined validation and merge status.

PR #940, opened 2026-09-12
fix(coordinator): make key and account token admission atomic

Gajesh2007:
Concurrent inference requests can both pass token-quota peeks before either charges the buckets. The later `Commit` ignores a failed charge, so two requests can proceed against a one-token account burst. The regression reproduces that outcome in the API admission helper.

The key and account checks/charges now run under one bounded account-sharded lock in `admitTokenBuckets`; extra output-token reconciliation uses the same lock. HTTP headers and error writes stay outside it. A private key-limit record replaces the six-value parameter tuple, and `server.go` retains tier selection and response handling.

Both token limiters now obtain availability and a retry hint through a non-consuming bucket check. A refill between the old `CanN` and consuming `AllowN` hint call can no longer debit a rejected request. Unlimited dimensions, oversized-charge clamps, service/admin behavior, output estimates and header/error shapes retain their existing rules.

```mermaid
flowchart LR
  subgraph Before
    A[Concurrent requests] --> B[Separate key and account Peek]
    B --> C[Locks released before Commit]
    C --> D[Failed charge ignored: quota exceeded]
    E[Failed CanN] --> F[Consuming AllowN for retry hint]
  end
  subgraph After
    G[Concurrent requests] --> H[admitTokenBuckets account lock]
    H --> I[Non-consuming checks for key and account]
    I -->|All fit| J[Charge both before unlock]
    I -->|Rejected| K[Unlock then existing 429]
    L[Additional output debt] --> H
  end
```

Validation:
- API concurrency regression failed on master: two admissions from a one-token burst in round 3.
- 100 contention rounds each for account-only, key-only, combined limits and different keys sharing an account pass after the fix (400 rounds total).
- Fixed-time refill tests verify rejected and successful checks never consume quota. Fixed/per-key tests preserve caps, oversized-charge behavior and unlimited dimensions.
- Focused API race tests passed (1.897s); limiter check tests passed (1.256s). The preceding broader token-related race run also passed (API 4.507s, limiter 1.325s), including existing header/rejection/estimation cases.
- Docs lint passed (278 pages). No current Markdown references the removed `keyTokenParams` helper.
- Full coordinator `GOTOOLCHAIN=go1.25.0 go test ./...` passed (API 195.247s); CI also validates the published branch.

The locks coordinate one coordinator process. They do not introduce distributed quota storage or change the existing per-process limiter model.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased — concurrent token quota enforcement
+
+- Prevent simultaneous inference requests from exceeding account or API-key token quotas through separate availability checks and charges. Extra output-token reconciliation shares the admission lock, and retry-hint checks no longer consume quota.
+
 ## Release candidate v0.9.2 — Gemma QAT caching, adaptive MTP and Nemotron Lightning (not shipped; 2026-09-10)
 
 Source changes since `v0.9.1`. Provider changes require a new signed bundle.
--- a/coordinator/api/server.go
+++ b/coordinator/api/server.go
@@ -205,6 +205,7 @@ type Server struct {
 	baseRewards                   *baserewards.Engine
 	logger                        *slog.Logger
 	mux                           *http.ServeMux
+	tokenAdmissionMu              [64]sync.Mutex  // account-wide key/tier admission and output reconciliation
 	modelAliasMutationMu          sync.Mutex      // serializes cross-endpoint alias validation + persistence
 	challengeInterval             time.Duration   // 0 means use DefaultChallengeInterval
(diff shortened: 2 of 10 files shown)


Gajesh2007:
@codex review

Gajesh2007:
Consolidated into #1056 for review and merge after #1094, #1095 and #1096. #1056 contains this branch's exact head (`aaa274d8e11bf5c7ad31aeea1588bab9d955c04f`), including its reviewed fixes, with implementations and regressions reconciled into the coordinator subsystem owners. Closing the component PR to collapse the stack; its branch, commits and review history are retained. Follow #1056 for combined validation and merge status.

PR #951, opened 2026-09-12
Refuse non-regular attention replay inputs without blocking

Gajesh2007:
The standalone replay CLI opens inputs before checking their file type. A FIFO supplied as the transfer or native buffer path therefore waits for a writer instead of reaching the existing regular-file refusal.

`readBounded` now opens with `O_NONBLOCK` alongside its existing read-only/no-follow/close-on-exec flags. The same descriptor is still checked for regular-file type, size and exact bytes. A host fixture supplies an owned FIFO without a writer; its failure recovery keeps a read/write descriptor open and joins the owned reader even if scheduling delayed its initial open. No operator, tensor math, allocation bound, hash, report schema or package pin changes.

**Before**
```mermaid
flowchart LR
  A[ReplayTransfer.load] --> B[readBounded opens input]
  B --> C[FIFO waits for writer]
  C -. cannot reach .-> D[fstat regular-file refusal]
```

**After**
```mermaid
flowchart LR
  A[ReplayTransfer.load] --> B[readBounded opens without blocking or following symlinks]
  B --> C[fstat rejects FIFO or other non-regular input]
  B --> D[Regular file retains exact size and byte checks]
  E[ReplayHostTests owned FIFO] --> C
```

Validation completed here: an owned Python/POSIX reproduction using the exact old flags times out before fstat; the new flags open immediately and identify the FIFO for refusal. The subprocess is killed/reaped by its timeout. The recovery descriptor also opens successfully on Darwin before any reader exists. Docs lint passes.

Root validation also completed at head `2b4be4e8831930edd719b150cbf1bfe214020092` on the dedicated M5: release build with tests passed in 131.97s, and all **5 ReplayHostTests passed with zero skips**. The test executable and replay binary hashes stayed identical before/after the run. All 7 package files and 5,093 reviewed dependency/source files were verified; all 31 resolved dependency pins matched the existing provider lockfile. No operator or model tests were run.

The first wrapper rejected Swift's “5 tests in 1 suite” summary even though the tests passed. That failure is retained. A corrected summary parser and fresh guarded verification reran the same compiled host suite successfully (exit 0); no source or package pins changed. Replay binary SHA-256: `94cea11bdf321493444388f9639847b737d21171975878a5ea837d0d6a1f48c8`.

### DIFF
--- a/docs/developer/build.md
+++ b/docs/developer/build.md
@@ -1,6 +1,6 @@
 # Build
 
-> Last updated: 2026-09-10 · commit `4f29957d2`
+> Last updated: 2026-09-12 · commit `1c4670fdf`
 
 How to build every component of Darkbloom from a fresh clone: the Go
 coordinator, the Rust prompt-contract sidecar, the Swift provider CLI (with its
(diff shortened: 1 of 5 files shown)


Gajesh2007:
@codex review

Gajesh2007:
Consolidated into #1096, the combined tooling, benchmark and E2E refactor PR. Your latest head `2b4be4e8831930edd719b150cbf1bfe214020092` is preserved in aggregate `9d2160bb370e8a0dbf3810e61a5e48fe66345651`, along with the relevant review fixes. Continue review and merge at https://github.com/Layr-Labs/d-inference/pull/1096 (after provider #1094 and web #1095). Closing this duplicate merge path; the source branch and review history are retained.

PR #952, opened 2026-09-12
Validate replay artifact ownership and simplify benchmark evidence operations

Gajesh2007:
Replay case IDs were used directly as artifact filenames, so a retained input could escape the new output directory, replace `warmup.json`/`report.json`, or silently collapse duplicate rows. An empty replay also fell back to generating a new workload, and a wrong model was rejected only after the warmup request. Separately, an owned process disappearing between `poll` and `killpg` could abort cleanup before it was reaped.

`load_replay` now validates the nonempty plan, model, non-null earlier comparison targets and unique safe artifact names (including the filesystem byte limit and `.json` suffix) before output creation or HTTP work. The owned-provider wrapper runs this preflight before host probes or model startup; the HTTP child revalidates before using the replay. Case/Unicode filename collisions are rejected for the Mac filesystem as well. Request bodies and ordered cases are preserved verbatim. Shared process cleanup treats a disappeared group as an exit race and still waits on the owned process; permission failures and unconfirmed termination still fail.

Pairwise generation comparisons now compute each participating observation's token/prompt identity hashes once, retaining independent dictionaries in every ordered comparison. Strict versus record policy, cancellation-prefix semantics, report schemas, generated-token gates and measured timing formulas are unchanged. No native source or dependency changes are in this PR; the separate native FIFO guard is #951.

**Before**
```mermaid
flowchart TD
  A[radix_prefix_cache.run] --> B[Create output and collapse replay rows into map]
  B --> C[Warmup before model validation]
  C --> D[Case ID becomes unchecked artifact path]
  E[run_radix_http.terminate] --> F[poll then signal]
  F --> G[Exit race raises before wait]
  H[comparison_records pair loop] --> I[Hash both identities for every pair]
```

**After**
```mermaid
flowchart TD
  A[Wrapper and HTTP run delegate load_replay] --> B[Validate model order comparison targets and bounded filenames]
  B --> C[Create output then warmup and exact replay requests]
  B --> D[Invalid plan refuses before HTTP or writes]
  E[terminate] --> F[Signal owned group]
  F --> G[Disappeared group still reaches owned wait]
  H[comparison_records pair loop] --> I[Lazy per-observation identity cache]
  I --> J[Independent ordered comparison values]
```

Validation: all 134 benchmark CPU tests pass using pinned Python 3.12/NumPy 2.4.2, including 95 scoped radix tests. New regressions reproduce the original replay failures, TERM/KILL exit races and repeated hashing. Fixtures use temporary owned directories and mocked HTTP/process operations; no real endpoint, `pgrep`, provider, macmon or signal is invoked. A 1,000-report differential matches exact prior JSON, including record/key order. Four same-prompt observations now use eight hashes instead of twenty-four. Docs lint passes on 278 documents.

Source review covers the remaining native benchmark packages and Python radix helpers; the reviewed native implementation and existing operator tests are unchanged here and were not executed. These CPU checks establish helper behavior, not model performance or release qualification.

All existing review findings are addressed. Final Codex approval is pending because the connector reports its review usage limit has been reached.

### DIFF
--- a/docs/developer/build.md
+++ b/docs/developer/build.md
@@ -1,6 +1,6 @@
 # Build
 
-> Last updated: 2026-09-10 · commit `4f29957d2`
+> Last updated: 2026-09-12 · commit `2619389bc`
 
 How to build every component of Darkbloom from a fresh clone: the Go
 coordinator, the Rust prompt-contract sidecar, the Swift provider CLI (with its
@@ -244,6 +244,13 @@ and retain their hashes with the build's source/dependency inventory. See
 <a id="resident-prefix-benchmark-executable"></a>
 
 #### Prefix-cache benchmark executable
+
+The radix Python replay and process-cleanup fixtures run with the standard
+library alone, without building their native executables
+(`scripts/benchmarks/test_radix_prefix_cache.py`, `ReplayInputTests`;
+`scripts/benchmarks/test_radix_process_cleanup.py`, `ProcessCleanupTests`).
+Use the [prefix-cache benchmark checks](test.md#prefix-cache-benchmark-validation)
+before running an authorized live measurement.
 
 [`scripts/benchmarks/radix-engine`](../../scripts/benchmarks/radix-engine/Package.swift)
 links the real provider factory and MLX packages from an explicitly selected
--- a/docs/developer/build.md
+++ b/docs/developer/build.md
@@ -1,6 +1,6 @@
 # Build
 
-> Last updated: 2026-09-12 · commit `2619389bc`
+> Last updated: 2026-09-12 · commit `9e79160f1`
 
 How to build every component of Darkbloom from a fresh clone: the Go
 coordinator, the Rust prompt-contract sidecar, the Swift provider CLI (with its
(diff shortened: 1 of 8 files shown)


Gajesh2007:
@codex review

Gajesh2007:
Addressed both static review comments in 3b6567cb6: documented the owned process-group disappearance race on escalation and standardized generation-comparison imports. All 93 radix CPU tests pass after this behavior-preserving change. @codex review

Gajesh2007:
Addressed all three Codex findings in source 9e79160f1 (documented at d506c6758). Non-null comparison targets now require nonempty earlier-case IDs, filename validation includes the .json suffix and destination filesystem byte limit, and the provider wrapper invokes shared replay validation before host probes/output/model startup. New fixtures fail baseline with 7 assertion failures and 1 filename error; all 95 scoped radix CPU tests now pass, including valid byte boundaries and null optional comparison. Docs lint passes. The earlier 3b6567cb6 comment overstated its narrow test command count; that command ran 59 tests, while this final run uses the complete test_*radix*.py pattern (95). Codex currently reports review usage limits, so final automated approval is pending.

Gajesh2007:
@codex review

Please review the current head. Earlier requests reached the review quota; reviews are now completing again. The fixes and scoped validation are recorded in the PR description.

Gajesh2007:
Consolidated into #1096, the combined tooling, benchmark and E2E refactor PR. Your latest head `d506c6758d8a5d7914ae75047817601f4f28b499` is preserved in aggregate `9d2160bb370e8a0dbf3810e61a5e48fe66345651`, along with the relevant review fixes. Continue review and merge at https://github.com/Layr-Labs/d-inference/pull/1096 (after provider #1094 and web #1095). Closing this duplicate merge path; the source branch and review history are retained.

PR #966, opened 2026-09-12
Prevent stale API-key cache writes after key mutations

Gajesh2007:
A slow API-key lookup can finish after a successful update, revoke or rotation and restore its old permissions in the coordinator's auth cache. The old cache code stamped the result with the generation at insertion. It could also restore a negative result after a disabled key was re-enabled.

Capture the cache generation before reading the store and publish the result only while that generation still matches. Legacy raw-token revocation uses the same invalidation fence as managed keys. The cache implementation now lives in `api_key_cache.go`; TTL, size bound, provider-token handling and an already-running request's snapshot remain unchanged. Invalidation is local to the coordinator handling the mutation; other processes and direct store edits still rely on the ordinary cache lifetime.

```mermaid
flowchart LR
  subgraph Before
    A1[requireAuth cache miss] --> B1[AuthenticateKey reads old record]
    B1 --> C1[Management mutation commits and invalidates]
    C1 --> D1[storeAPIKeyCache stamps current generation]
    D1 --> E1[Later requests reuse revoked key or old limits]
  end
  subgraph After
    A2[lookupAPIKeyCache captures generation] --> B2[AuthenticateKey reads record]
    B2 --> C2[Management mutation commits and advances generation]
    C2 --> D2[storeAPIKeyCache rejects old generation]
    D2 --> E2[Later requests read current key state and cache it]
  end
```

Validation:
- Six real HTTP interleavings cover disable, delete, rotation, legacy revoke, limit restriction and re-enable using a real memory store and locally signed Privy session. All six fail on the baseline; the focused Go 1.25 race run passes. Each case checks that the next request refreshes from the store and the following request uses the cache.
- Full Go 1.25 API race suite passes (227.015 s).
- Docs check passes for 278 pages; source stamp points to reachable commit `f176a28ac`.
- Two independent source/test peer reviews are clean.


Gajesh2007:
Consolidated into #1056 for review and merge after #1094, #1095 and #1096. #1056 contains this branch's exact head (`7258c3e1d39469d90936269b7eb91c50dde26387`), including its reviewed fixes, with implementations and regressions reconciled into the coordinator subsystem owners. Closing the component PR to collapse the stack; its branch, commits and review history are retained. Follow #1056 for combined validation and merge status.

PR #969, opened 2026-09-12
fix(api): include exact histogram upper bounds

Gajesh2007:
Histogram samples exactly on an upper bound, such as a 5 ms HTTP duration, were counted in the next bucket. Adding `1e-9` before the bucket search also misplaced values immediately below a boundary.

Search using the original sample so cumulative JSON counts and Prometheus `le` buckets include their upper bounds. Bucket configuration, sample totals and sums retain their existing behavior.

```mermaid
flowchart LR
  subgraph Before
    A1[Observe a 5 ms sample] --> B1[Search upper bounds with sample plus epsilon]
    B1 --> C1[JSON and Prometheus 5 ms bucket count stays zero]
  end
  subgraph After
    A2[Observe a 5 ms sample] --> B2[Search upper bounds with original sample]
    B2 --> C2[JSON and Prometheus 5 ms bucket count includes sample]
  end
```

Validation:
- Regressions exercise all 13 default boundaries and the adjacent representable floats, plus exact exported Prometheus lines. Baseline fails (0.532 s).
- Go 1.25 race run passes (1.870 s), covering the new cases, existing registry/snapshot/rendering checks and relevant API metric emission checks.
- Documentation check passes for 278 pages; source stamp points to reachable commit `ea75820bf`.
- Independent peer review is clean.


Gajesh2007:
Consolidated into #1056 for review and merge after #1094, #1095 and #1096. #1056 contains this branch's exact head (`3308662b224b0fdd72b743959c09701e8a19ff35`), including its reviewed fixes, with implementations and regressions reconciled into the coordinator subsystem owners. Closing the component PR to collapse the stack; its branch, commits and review history are retained. Follow #1056 for combined validation and merge status.

PR #970, opened 2026-09-12
fix(api): omit keyed geo URLs from failure logs

Gajesh2007:
A failed PRO geolocation request was logged with Go's `url.Error`, which embeds the complete request URL including the API key query parameter. This could expose the key when debug logging is enabled.

Log the underlying transport cause without that URL wrapper. The resolver still returns no location on failure; successful requests and their PRO/free-tier selection retain their existing behavior.

```mermaid
flowchart LR
  subgraph Before
    A1[PRO lookup fails] --> B1[http.Client wraps cause with keyed URL]
    B1 --> C1[Debug log includes query API key]
  end
  subgraph After
    A2[PRO lookup fails] --> B2[lookupIPAPI extracts underlying cause]
    B2 --> C2[Debug log retains cause and tier without request URL]
  end
```

Validation:
- Two local stock-http.Client transport failures, including timeout, reproduce the leaked escaped fixture key on the baseline (0.593 s). No live service or real credential is used.
- Focused Go 1.25 geo/IP/header race tests pass (1.676 s); diagnostics retain the cause and tier while omitting the raw/escaped key and request URL.
- Docs check passes for 278 pages; stamp references reachable source `8e51192fe`.
- Independent source/test/docs peer review is clean.


Gajesh2007:
Consolidated into #1056 for review and merge after #1094, #1095 and #1096. #1056 contains this branch's exact head (`efc30d1425837fedeaa3f6c9d305c7d9e4d6f3ca`), including its reviewed fixes, with implementations and regressions reconciled into the coordinator subsystem owners. Closing the component PR to collapse the stack; its branch, commits and review history are retained. Follow #1056 for combined validation and merge status.

PR #984, opened 2026-09-12
Ignore provider completions before registration

Gajesh2007:
An `inference_complete` frame received before `register` dereferences a nil provider while recording ingress time, so the WebSocket closes before registration can finish. Ignore that frame at the read-loop boundary, matching the existing accepted/chunk/error handlers. Registered completions keep their existing timestamp and asynchronous settlement path.

The regression uses a real local WebSocket: send a completion, then register on the same connection and require its initial attestation nonce challenge. It fails on master and passes with the guard.

### Before

```mermaid
flowchart LR
  A[Completion before register] --> B[providerReadLoop]
  B --> C[MarkPendingCompletionIngressNow on nil provider]
  C --> D[Handler panic and deferred socket close]
  D --> E[Following registration cannot complete]
```

### After

```mermaid
flowchart LR
  A[Completion before register] --> B[providerReadLoop registration guard]
  B --> C[Ignore frame]
  C --> D[Same connection registers and receives challenge]
  E[Completion after register] --> F[Existing ingress stamp and async handleCompleteAt]
```

Validation at source `beca0ca591e657dfd2b19d2ce3a2b2a606a4d319`:

- Baseline regression fails in 0.530s because the connection closes before registration.
- Focused Go 1.25 race checks pass in 1.958s: new regression plus existing heartbeat-before-registration, duplicate-register/reconnect, and oversized-version cases.
- Full Go 1.25 API race suite passes in **228.895s** with `DATABASE_URL` and `EIGENINFERENCE_DATABASE_URL` unset.
- Independent source review is clean. `make docs-check` passes for all 278 documents; signed commits and `git diff --check` pass.

No protocol fields, trust rules, billing behavior for registered requests, or provider dependencies change.


Gajesh2007:
@codex review

Gajesh2007:
Consolidated into #1056 for review and merge after #1094, #1095 and #1096. #1056 contains this branch's exact head (`7900857dfd52f9f7a16d7c69912e3483ec239652`), including its reviewed fixes, with implementations and regressions reconciled into the coordinator subsystem owners. Closing the component PR to collapse the stack; its branch, commits and review history are retained. Follow #1056 for combined validation and merge status.

PR #990, opened 2026-09-12
test: verify blue-green sessions and reward overlap

Gajesh2007:
The blue-green session fixtures could pass when a store result duplicated one session, corrupted its heartbeat, or the reward engine double-counted overlapping durations and capped the result at 100%.

The existing store fixture now requires each expected session exactly once, including its heartbeat, account/provider identity, open state and connection window. The existing settlement fixture covers `[0%, 60%]` and `[50%, 95%]` of an epoch and requires the payout for their 95% union. Summing those durations yields 105%, so a 100% cap alone cannot pass. The redundant test-only union implementation is removed; all 21 existing test entries remain, with 47 fewer source lines.

```mermaid
flowchart LR
  subgraph Before
    B1[ListProviderSessionsOverlapping] --> B2[Count two rows]
    B2 --> B3[Test helper unions and clamps its own intervals]
    B3 --> B4[Duplicate or corrupted session can pass]
    B5[SettleEpoch with full-period overlap] --> B6[Union or capped sum pays full floor]
    B6 --> B7[Double counting can pass]
  end
  subgraph After
    A1[Same Memory and PostgreSQL calls] --> A2[Consume both expected session IDs]
    A2 --> A3[Check heartbeat and open identity]
    A3 --> A4[Duplicate or corrupted session fails]
    A5[SettleEpoch with 95 percent union] --> A6[Actual uptimeByProviderKey and PeriodFloor]
    A6 --> A7[Only payout for 95 percent uptime passes]
  end
```

Validation used `GOTOOLCHAIN=go1.25.0` and an owned disposable PostgreSQL 16 database. Temporary Go overlays introduced faults at the actual result comparison or interval accumulation:

| Temporary fault | Baseline | Candidate |
| --- | --- | --- |
| Replace second session with first, preserving row count | Both backends incorrectly pass | Both reject duplicate ID |
| Corrupt returned `LastSeen` | Both backends incorrectly pass | Both reject heartbeat mismatch |
| Sum intervals while retaining production 100% cap | Settlement incorrectly passes | Rejects 2,016 µUSD instead of 1,008 µUSD |

Both unmodified controls pass before and after. Under the existing availability ramp, 95% uptime earns half the full-uptime floor, so the settlement fault would double this fixture's payout.

The unmodified race checks passed: **19 top-level tests / 25 including subtests, zero failures or skips**. The three session-read tests exercised Memory and PostgreSQL in 1.814s; 16 reward/epoch checks passed in 1.280s. The validation overlays are temporary artifacts; production source is unchanged.

`gofmt`, `git diff --check`, and `make docs-check` (278 pages) pass. Both commits are signed, and the owned PostgreSQL database is stopped. Evidence: `/private/tmp/session-overlap-oracles-20260912/`, with source hashes and review scope in `/private/tmp/repository-session-overlap-oracles-review.json`.

Independent source review is clean at `63a40326bce0d84ffda424b78218a6ca1932d3b2`, including both store backends and the actual settlement/uptime/floor paths. Fresh Codex review is clean at head `891d3691d6f4a16f1124442d1df7c4eac1541440`, with an actual thumbs-up and zero review threads. This checks store reads and actual reward settlement; it does not exercise a live coordinator deployment or provider/model runtime.


Gajesh2007:
Consolidated into #1056 for review and merge after #1094, #1095 and #1096. #1056 contains this branch's exact head (`891d3691d6f4a16f1124442d1df7c4eac1541440`), including its reviewed fixes, with implementations and regressions reconciled into the coordinator subsystem owners. Closing the component PR to collapse the stack; its branch, commits and review history are retained. Follow #1056 for combined validation and merge status.

PR #1098, opened 2026-09-16
fix(datadog): submit histograms as distributions over HTTPS

ethenotethan:
## Problem

The production coordinator runs with `DD_API_KEY` set and **no Datadog agent on the host**. Counters and gauges already noticed this and divert to the HTTPS `/api/v1/series` intake (`metrics_http.go`). `Client.Histogram` never did — it only wrote DogStatsD UDP datagrams to `localhost:8125`.

Connecting a UDP socket succeeds whether or not anything is listening, so `statsd.New` returned a live client, every `Histogram` call returned `nil`, and the kernel discarded the datagrams. There is no dropped-packet counter on that path, which is why ~39 histogram metric names have been silently missing rather than visibly failing — including `http.latency_ms`, `inference.ttft_ms`, `inference.timing.*`, `store.{debit,credit}.latency_ms` and the billing latency families.

Two consequences, both fixed here:

1. The values never reached Datadog at all.
2. Even where an agent *does* exist (dev), the dashboard queried `avg:<metric>.95percentile{…}` — an agent-computed per-window percentile submitted as a plain gauge. Averaging percentiles is not a percentile of anything, and it gets more wrong as the time range widens. (`.50percentile` never existed at all; the agent names that one `.median`.)

## Fix

Histograms take the HTTPS `/api/v1/distribution_points` intake, on the same **replace-not-tee** rule the series path already uses — so an agent appearing later cannot double-count. Datadog stores the raw values and computes percentiles server-side over whatever range is queried, so the dashboard asks for `pNN:<metric>` instead.

### Before / After — behavior

```mermaid
flowchart LR
  subgraph Before
    A1["ddHistogram(&quot;http.latency_ms&quot;, 12)"] --> B1["Client.Histogram"]
    B1 --> C1["statsd UDP → localhost:8125"]
    C1 --> D1{"agent listening?"}
    D1 -->|"dev: yes"| E1["agent aggregates →<br/>http.latency_ms.95percentile (gauge)"]
    D1 -->|"prod: no"| F1["kernel discards datagram<br/>send returns nil, nothing logged"]
    E1 --> G1["widget: avg of percentiles<br/>(wrong, and empty for p50)"]
    F1 --> H1["widget: No data"]
  end
```

```mermaid
flowchart LR
  subgraph After
    A2["ddHistogram(&quot;http.latency_ms&quot;, 12)"] --> B2["Client.Histogram"]
    B2 --> C2{"httpDistributions()<br/>API key + buffer?"}
    C2 -->|"yes (every deployed env)"| D2["distBuffer: raw values<br/>per (metric, tags)"]
    D2 --> E2["flushDistributions, every 5 s"]
    E2 --> F2["POST /api/v1/distribution_points<br/>type: distribution"]
    F2 --> G2["Datadog computes pNN<br/>over the queried range"]
    G2 --> H2["widget: p50/p95/p99 render<br/>on any host, agent or not"]
    C2 -->|"no API key"| I2["statsd UDP (unchanged)"]
  end
```

### Before / After — code

```mermaid
flowchart TB
  subgraph Before
    direction TB
    CB["Count / Gauge"] --> HB{"httpMetrics()"}
    HB -->|true| SB["seriesBuffer"] --> FSB["flushSeries → /api/v1/series"]
    HB -->|false| UB["Statsd"]
    HGB["Histogram"] --> UB2["Statsd — no branch, no HTTPS leg"]
    HOGB["HistogramOrGauge"] --> HB
  end
  subgraph After
    direction TB
    CA["Count / Gauge"] --> HA{"httpMetrics()"}
    HA -->|true| SA["seriesBuffer"] --> FSA["flushSeries → /api/v1/series"]
    HA -->|false| UA["Statsd"]
    HGA["Histogram"] --> HDA{"httpDistributions()"}
    HDA -->|true| DA["distBuffer<br/>(reservoir-capped)"] --> FDA["flushDistributions →<br/>/api/v1/distribution_points"]
    HDA -->|false| UA
    HOGA["HistogramOrGauge"] --> HOGG{"either HTTPS leg?"}
    HOGG -->|true| GA["Gauge — never a distribution"]
    HOGG -->|false| UA
  end
```

`httpDistributions()` is deliberately a separate predicate from `httpMetrics()` rather than a widened one: the two buffers are independent, and a hand-assembled `Client` with only a `seriesBuffer` must fall back to DogStatsD for histograms rather than dereference a nil buffer.

`HistogramOrGauge` (19 call sites, heartbeat snapshots) keeps submitting gauges. It no longer exists to survive agentlessness, but a gauge is still the right type for "what does the allocator read now", and a metric name already stored as a gauge cannot be reinterpreted as a distribution without breaking every query on it. It now takes the gauge path when *either* HTTPS leg is live: gated on `httpMetrics()` alone, a `Client` holding only a `distBuffer` would have fallen through to `Histogram` and submitted these names as distributions — the exact reinterpretation the comment warns against.

### Bounded memory

`distBuffer` cannot pre-aggregate — the point of a distribution is that the intake sees every observation — so an unbounded slice per series would be a heap leak on a traffic burst. Past 2000 values in one 5 s window a series is reservoir-sampled (Vitter Algorithm R), which keeps percentiles unbiased while making `count`/`sum` under-report. That trade-off is logged when it engages rather than left to look like a quiet window — including the part that is easy to overclaim: percentiles *over that window* stay unbiased, but a sampled window then contributes the same 2000 values as an unsampled one, so a wide time range weighs a busy window like a quiet one.

One flush is split across POSTs on whichever of two bounds comes first: 50 000 values, or an estimated 2 MB. Value count alone is not enough — 50 000 values spread over 50 000 one-value series is ~14 MB of metric names and tags. The v1 spec documents a 3.2 MB body cap for `/api/v1/series` and none for `distribution_points`; 3.2 MB is assumed here as the conservative figure. The per-entry estimate uses the widest a `json.Marshal`'d float64 can get (24 bytes, scientific notation) rather than the ~16 a latency reading takes, so the estimate stays above the real body.

## Deploy side

Two changes outside the Go code, both required for the widgets to actually render:

1. **`deploy/datadog/dev-network-dashboard.json`** — 19 queries rewritten from `avg:<metric>.NNpercentile{…}` to `pNN:<metric>{…}`, across 7 metrics. The 4 `trace.http.request.latency.*` APM queries are left alone; those really are trace-side aggregates.
2. **`deploy/datadog/enable-distribution-percentiles.sh`** (new) — a distribution answers only avg/sum/min/max/count until percentile aggregators are explicitly enabled on it, per metric, per org. The script derives the metric names from the dashboard JSON (so a new `pNN:` widget cannot be added without the script covering it), reports current state by default, and writes only with `--apply`. Each write sets `include_percentiles: true` with `exclude_tags_mode: true` and an empty tag list — "exclude nothing". The other form of a tag configuration is a `tags` **allowlist**, which silently stops resolving every key not on it; `http.latency_ms` alone is emitted with `method`, `path` and `status_code`, so an allowlist derived from the dashboard would have quietly cost "p95 latency for 5xx". `metric_type` goes on the create only — it is not in the update schema, so sending it on a re-run would fail every re-run.

Percentiles are deliberately **not** enabled for all ~35 histogram names, only the ones something queries that way, because enabling percentiles roughly doubles what that metric bills. Any other name can be added as an argument: `--apply d_inference.inference.ttft_ms`. Everything outside that set still answers `avg:`/`count:`/`max:`.

3. **`docs/operations/datadog-dashboard.md`** (new runbook) — the order is deploy → wait for one flush → `--check` → `--apply` → apply the dashboard. Datadog derives a metric's type from submitted data and cannot configure one it has never received, so running the script first achieves nothing; the widgets stay empty even though the data is arriving. The runbook also carries the cost check, the verification queries, and the known-empty widgets below.

## Known consequences

- **Dev (which has an agent):** the `.95percentile` series stop advancing and the bare distribution starts. That is the point — one behavior in both environments — but it is a visible cutover, not a silent one.
- **`routing.cost_ms` stays empty.** Its only `ddHistogram` calls are in `routing_metrics_test.go` — no production code emits it. Pre-existing, but this PR rewrote the widget's query, so it is now written down as known-empty in the runbook rather than left to look like a transport failure. Fixing it means emitting from the scheduler's cost computation, which is out of scope here.
- **APM widgets stay empty.** `ddtracer.Start()` already runs on prod, but no code creates spans, so `trace.http.request.*` needs instrumentation, not a sidecar. Also pre-existing, also written down.
- **Cost.** Every histogram becomes a distribution on deploy, script or no script, and a distribution is not billed as one custom metric: ~5 per timeseries (count/sum/min/max/avg), ~10 with percentiles. So the deploy is the 5× step and this script is only 2× on top of it, against a production baseline of zero — histograms were being discarded for want of an agent. Timeseries count is driven by tag cardinality, and the dominant contributor is `http.latency_ms` (`path` is bounded to registered route patterns by `httpPathLabel`, so a few hundred series). Order of magnitude for the coordinator: hundreds to low thousands of distribution timeseries. The runbook says to read the real number from **Plan & Usage → Custom Metrics** after deploying, and that whoever ran the deploy owns that check.

## Tests

`coordinator/datadog/metrics_distribution_test.go` — 15 tests, real UDP sockets and a real `statsd` client rather than mocks, `httptest` for the intake:

- histograms reach the distribution intake with the right metric name, type, host, tags and **raw un-aggregated values**, and **do not leak to UDP** (replace, not tee)
- without an API key, DogStatsD still carries `d_inference.http.latency_ms:12|h|#path:/health` and nothing is POSTed
- series split by tag set; a flushed window is not resubmitted
- the reservoir bounds the window, submits only real observations, and provably samples rather than truncates
- an oversized flush spans requests without losing a series, on the value bound *and* independently on the byte bound (many thin series, which the value bound alone cannot catch)
- the timestamp goes out as a JSON number, which the intake requires and nothing else here would notice
- a 403 is reported with its status and the intake's own reason; a 202 is silent
- `NewClient` really wires `distURL` and the buffer (the other tests stub those, so this is what catches a deleted assignment)
- nil receiver, missing buffer, empty client, and a client with no logger at all are safe on both reporting paths — including `NewClient` itself, whose DogStatsD-failure warning is the most likely thing to fire on an agentless host
- `HistogramOrGauge` never emits a distribution, even for a `Client` that has only a `distBuffer`
- both buffers copy the caller's tag slice, so a caller that reuses one cannot retag a buffered point

Reverting the `Histogram` rewrite fails 9 of the 15; each of the three new tests fails on reverting just its own fix.

Verified: `gofmt -l` clean, `go build ./...`, `go vet ./datadog/`, `go test ./datadog/... -race`, full `coordinator` suite exit 0, `make docs-check` clean, `bash -n` under bash 3.2.57 plus an end-to-end run against a stubbed intake exercising every script branch in both `--check` and `--apply` modes.

## Docs

New: `docs/operations/datadog-dashboard.md` (indexed in `docs/operations/README.md`). Updated: `docs/architecture/telemetry.md` (transport table, the percentile explanation and what percentiles are *not* enabled for, 4 new failure-mode rows, code map), `docs/architecture/system-profiler.md` (its "histograms do not survive HTTPS" claim was the old truth), `docs/architecture/routing.md`, `docs/reference/telemetry-inventory.md` (the two `histogram (DogStatsD only)` rows, plus how to query a histogram now — `avg:` always, `pNN:` only once enabled), `docs/reference/configuration.md`, `docs/operations/dev-environment.md` (its Datadog row implied the agent carried metrics). Also corrected the dashboard's own header note, which still told readers metrics arrived "via DogStatsD (agent on 8125)". All restamped; `make docs-check` passes.

## Review follow-up (commit 2)

Both review passes ran on the first commit; the Go pass came back PASS, the diff pass flagged the cost figures as blocking. Fixed in a second commit:

- **Cost was wrong by 5×** in the one section whose job is to prevent a bill surprise — corrected in the runbook, `telemetry.md`, the script header and above.
- `NewClient` warned through the raw logger, a nil dereference for an embedder passing no logger.
- `HistogramOrGauge` could submit gauge names as distributions (above).
- `seriesBuffer` retained the caller's tag slice, the same hazard `distBuffer` had just fixed; the codebase does `ddIncr(a, tags)` then `ddIncr(b, append(tags, extra))`, safe only while those literals have no spare capacity.
- `distValueBytes` 16 → 25, since a marshalled float64 reaches 24 bytes plus its comma.
- Script: reads `metric_type` and says "is a gauge, not a distribution" instead of letting Datadog answer with a 400; POST 409 counts as unchanged rather than failing the run; only calls a non-empty tag list an *allowlist* when exclude mode is off; `pNN:` extraction no longer requires scope braces and also reads `q` keys.
- Runbook: `--apply` marked **human-only** (an org-wide write that changes production billing), `DD_SITE`/`DD_APPLICATION_KEY` fixed in the step-1 curl, `app.datadoghq.com` → `app.<DD_SITE>`, and the rollback note corrected — unsetting `DD_API_KEY` stops log and event forwarding outright, since those have no UDP leg.

## Note on the first commit

This branch also carries `fix(datadog): tag forwarded logs with env and service` — the other half of the same investigation. Forwarded logs were arriving over HTTPS but carried no `env` tag, and every dashboard widget scopes by `env`/`service`, so nothing matched. Same root cause shape: on a host with no agent, nothing downstream adds what an agent would have added.


ethenotethan:
Superseded: we're adding a Datadog agent to the prod host instead of hand-rolling agent-side aggregation in the coordinator. `metrics_distribution.go` (reservoir sampling, payload chunking, distribution_points POSTs) is the wrong place for that logic when a real agent can do it.

Closing rather than merging. Carrying forward into the agent PR, because these are transport-independent and still needed:

- `e22d49019` — forwarded logs carried no `env`/`service` tag, so every dashboard widget scoped by `env` matched nothing. Independent bug, unrelated to histograms.
- `dev-network-dashboard.json` — the 19 `avg:<metric>.NNpercentile{…}` → `pNN:<metric>{…}` rewrites. Still correct under the agent: DogStatsD `Distribution` forwards raw values to the same `distribution_points` intake, so the metrics are bare distributions either way. `avg:` of an agent-computed percentile gauge is wrong regardless of how the values get there.
- `enable-distribution-percentiles.sh` + `docs/operations/datadog-dashboard.md` — percentile aggregators are per-metric, per-organization Datadog config. Nothing about the transport changes that.

Dropped: the HTTPS distribution leg and its 15 tests.

Note for the agent PR: the fix is `statsd.Distribution()`, not `statsd.Histogram()`. `Histogram` + agent is what dev has today and it produces `.95percentile` **gauges** computed per flush window, which can't be re-aggregated — that was the second half of the original bug, not just the missing agent.

PR #1140, opened 2026-09-19
fix(coordinator): 503 from /v1/leaderboard when the ranking query fails

brandon-eigenlabs:
## Summary

`store.Leaderboard` swallowed every query error and returned `nil`, so a timed-out ranking over `provider_earnings` was served as `200 {"entries":[]}` and cached for five minutes as if no provider had earned anything — while `/v1/network/totals` hitting the same 10 s timeout on the same table answered 503. This makes `Leaderboard` return `([]LeaderboardRow, error)` on both backends and has `handleLeaderboard` answer 503 `service_unavailable` without caching, matching `handleNetworkTotals` and `handleNetworkSeries`. A genuinely empty window is still a successful, cacheable 200.

## Linked issue

Closes #1139

## Test plan

Local, on this branch (`coordinator/`):

- [x] `gofmt -l ./store ./api` — clean
- [x] `go build ./...` (root module) — ok
- [x] `go vet ./store/ ./api/` — clean
- [x] `go test ./store/ -run 'Leaderboard|NetworkTotals|IsRewardLedgerType' -count=1` — pass
- [x] `go test ./api/ -run 'Leaderboard|Totals|StatsRefresh|CacheRefresh' -count=1` — pass (includes the two new regression tests below)
- [x] `go test $(go list ./... | grep -v /internal/api) -count=1` — the pre-push set, pass
- [x] `make docs-check` — 309 files OK
- [x] `make docs-impact-check BASE=origin/master` — see *Documentation impact*
- [ ] CI Postgres job runs `TestLeaderboardReturnsErrorWhenUnavailable` (needs `DATABASE_URL`; skipped locally, no Postgres on this box)

New tests:

- `coordinator/api/leaderboard_test.go`
  - `TestLeaderboardStoreErrorIs503NotEmptyBoard` — failing store → 503 `service_unavailable`, nothing cached under `leaderboard:earnings:24h:50`; after recovery → 200 with the seeded row and the body is cached. **Fails without the fix** (previously 200 with `entries: []`, cached).
  - `TestLeaderboardEmptyWindowIs200` — pins the contract that no earnings is a successful, cached empty board.
- `coordinator/store/analytics_workmem_test.go`
  - `TestLeaderboardReturnsErrorWhenUnavailable` — Postgres closed-pool test mirroring `TestNetworkTotalsReturnsErrorWhenUnavailable`.
[...]

Issue #1139, opened 2026-09-19
[bug] /v1/leaderboard serves a timed-out ranking query as 200 {"entries":[]} and caches it for 5 min


brandon-eigenlabs:
## What happens

`store.Leaderboard` returns a bare `[]LeaderboardRow` and swallows every query error (`coordinator/store/postgres.go`, `if err != nil { return nil }`). When the three-CTE aggregate over `provider_earnings` exceeds its 10 s context timeout, the handler receives an empty slice, serializes it as `200 {"entries":[]}` and caches that body for five minutes as a successful board.

Observed live on production on 2026-09-18 against `api.darkbloom.dev`:

```
GET /v1/leaderboard?metric=earnings&window=24h -> 200  0.04s  50 entries   (cache hit)
GET /v1/leaderboard?metric=earnings&window=all -> 200 10.04s   0 entries   (cache miss -> timeout)
GET /v1/leaderboard?metric=jobs&window=24h     -> 200 10.04s   0 entries   (cache miss -> timeout)
GET /v1/network/totals?window=all              -> 503 17.02s               (same table, same instant)
```

The 10.04 s latency is the store timeout hit to the millisecond. An empty leaderboard is therefore currently ambiguous between "no provider earned anything" and "the query timed out", and any downstream consumer will plot the latter as zero.

## Why it matters

- `/v1/network/totals` hitting the same timeout on the same table answers 503 `service_unavailable` (`handleNetworkTotals`: "say so rather than serve a zero row as if it were data"). The two public endpoints disagree about the same outage.
- The failure is silent: the 503 got investigated; the empty board was served as a measurement for days.
- Scan errors are also swallowed (`continue`), so a partially-read board can be published as complete.

## Expected

`store.Leaderboard` returns `([]LeaderboardRow, error)` on both backends; `handleLeaderboard` answers 503 `service_unavailable` on error and does not cache it, matching `handleNetworkTotals` and `handleNetworkSeries`. A genuinely empty window is still a successful, cacheable 200 with empty entries.

## Not in scope here

The underlying cost of the aggregate (~140M-row `provider_earnings` with no `created_at`-leading index, so 24h/7d windows are parallel seq scans and `all` is a full index scan) is a separate performance issue. This issue is only about reporting the failure honestly.

owizdom:
nice one, was about pushing a pr for this. (leaderboard querying)

thanks @brandon-eigenlabs

PR #1142, opened 2026-09-19
perf(coordinator): BRIN index on provider_earnings(created_at)

### LINKED ISSUE #1141: perf: provider_earnings has no created_at-leading index, so windowed network totals and leaderboard queries seq-scan ~140M rows and time out
## What happens

`/v1/network/totals` has answered 503 for every window since 2026-09-10, and `/v1/leaderboard` cache misses hit the same 10 s store timeout (#1139 made that failure honest; this issue is about making the query fast enough not to fail).

Both `NetworkTotals` and `Leaderboard` (`coordinator/store/postgres_analytics.go`, `coordinator/store/postgres.go`) filter `provider_earnings` on `created_at >= $1` across all accounts. The table's indexes all lead with an equality column (`id`, `job_id`, `(account_id, created_at)`, `(provider_key, created_at)`), so none can serve that predicate. Measured on the production read replica on 2026-09-19:

| | |
|---|---|
| `provider_earnings` | ~140M rows, 33 GB heap, 78 GB with indexes, growing ~4M rows/day |
| rows in 24h / 7d / 30d | ~4.0M (2.9%) / ~26M (19%) / ~106M (76%) |
| 24h plan | `Parallel Seq Scan` of the whole heap, estimated **6,021** rows (actual ~4.0M — 670× under) |
| 7d / 30d plans | `Parallel Seq Scan` |
| `all` plan | full `Index Scan` over `idx_provider_earnings_account` (~144M entries) |
| `pg_stats.correlation(created_at)` | **0.9974** — the heap is physically in insertion order |

The 670× misestimate is because the `created_at` histogram's last bucket ends hours before "now"; at 4M rows/day the default autovacuum analyze threshold (10% of the table ≈ 14M rows) means statistics lag by days.

`Leaderboard` additionally runs on a bare `pool.Query` (session `work_mem` = 4 MB) while `NetworkTotals` runs under `withAnalyticsTx` (`work_mem` = 1 GB), so its two `GROUP BY account_id` sorts over millions of rows spill to disk.

## Proposed fix (first of a series)

1. A **BRIN** index on `provider_earnings(created_at)`, built `CONCURRENTLY` at boot the way `ensureProviderEarningsJobIndex` / `ensureProviderRestoreIndex` already do. With correlation 0.997 BRIN prunes to the block ranges that can contain the window (~3% of pages for 24h, ~19% for 7d), is kilobytes in size, and builds in about a minute; a btree would be 2–3 GB and the covering btree previously proposed ~15 GB.
2. `ALTER TABLE provider_earnings SET (autovacuum_analyze_scale_factor = 0.005)` so the `created_at` histogram is refreshed every ~0.7M rows instead of ~14M (`SHARE UPDATE EXCLUSIVE`, non-blocking).
3. Run `Leaderboard` under `withAnalyticsTx` like the other analytics statements.

## Explicitly not fixed by this

- `window=30d` touches 76% of the table; no index helps. Needs a time-bucketed rollup.
- `window=all` should be served from the already-maintained `earnings_summary` table; that requires one additive column to preserve the API's work/reward split. Separate issue/PR.

brandon-eigenlabs:
## Summary

The windowed `NetworkTotals` and `Leaderboard` aggregates filter `provider_earnings` on `created_at >= $1` across all accounts, but every index on the table leads with an equality column, so the 24h/7d/30d windows plan as parallel seq scans of a ~140M-row / 33 GB heap and exceed the 10 s store timeout (the cause of the standing `/v1/network/totals` 503 and the leaderboard cache-miss timeouts behind #1139). This builds a BRIN index on `created_at` at boot through the existing concurrent-index helper, pins the table's analyze cadence so the `created_at` histogram stops trailing ingestion by days, and runs the leaderboard query under `withAnalyticsTx` like the other stats statements.

## Linked issue

Closes #1141

## Test plan

Local, on this branch:

- [x] `cd coordinator && gofmt -l ./store` — clean
- [x] `go build ./...` (root module) — ok
- [x] `cd coordinator && go vet ./store/` — clean
- [x] `cd coordinator && go test ./store/ -count=1` — pass (Postgres-backed tests skip without `DATABASE_URL`)
- [x] `cd coordinator && go test $(go list ./... | grep -v /internal/api) -count=1` — the pre-push set, pass
- [x] `make docs-stamp FILES=docs/architecture/storage.md`, `make docs-impact-check BASE=origin/master` — "7 changed file(s), canonical documentation coverage OK"
- [x] `make docs-check` — 309 files OK
- [ ] CI Postgres job (`DATABASE_URL`) runs the two new/extended Postgres-backed tests below — no local Postgres on this box

Tests:

- `coordinator/store/provider_earnings_window_index_test.go` — `TestProviderEarningsWindowIndex_BootSafe`: after `NewPostgres`, `idx_provider_earnings_created_at_brin` is valid and `pg_am.amname = 'brin'`, `provider_earnings.reloptions` contains `autovacuum_analyze_scale_factor=0.005`; re-running `migrate()` (a restart) is a no-op that leaves both in place; the index definition is `USING brin (created_at)` and `brin_summarize_new_values()` runs on it. (Planner choice is not asserted — on a test-sized table it is scale-dependent; see *Deployment* for the prod `EXPLAIN`.)
- `coordinator/store/analytics_workmem_test.go` — `TestUsageAnalyticsRunInWorkMemTransaction` now also asserts `Leaderboard` runs `BEGIN READ ONLY → SET LOCAL work_mem → statement → COMMIT` on one connection via the existing statement tracer.
- Existing `provider_restore_index_test.go` covers the renamed helper through `ensureProviderRestoreIndexes`.

## Components touched

- [x] coordinator (Go)
- [ ] provider (Rust, legacy)
- [ ] console-ui (Next.js)
- [ ] enclave (Swift)
- [ ] infra / CI / release
- [x] docs

No HTTP, WebSocket, config or CLI change. **Schema/migration effect (no version bump):** a new index `idx_provider_earnings_created_at_brin` and the table storage parameter `autovacuum_analyze_scale_factor = 0.005` on `provider_earnings`, both applied at boot, idempotently. See *Deployment* below for the one-time build cost.

## Documentation impact

- [x] Canonical documentation updated
- [ ] No documentation needed — reason:

`docs/architecture/storage.md`: boot flowchart gains the new phase; invariant 4 ("boot never holds a long lock on a hot table") now covers this index and the analyze pin; a failure-mode row for the totals/leaderboard 503; code-map row for the boot-time concurrent index builds; freshness stamp refreshed. `CHANGELOG.md`: *Unreleased — earnings window index*.

### Why BRIN

Measured on the production read replica (2026-09-19): `pg_stats.correlation` for `provider_earnings.created_at` is **0.9974** — the heap is physically in insertion order. That is the case BRIN exists for: block-range summaries prune to the ~3% (24h) / ~19% (7d) of pages that can contain the window, the index is kilobytes (a btree would be 2–3 GB; the covering btree floated earlier ~15 GB), it builds in about a minute, and — because BRIN has no useful selectivity for `window=all` — the planner never trades the current full index scan for something worse on that window. Default `pages_per_range = 128` (1 MB ranges) is kept; `autosummarize` is on by default in PG 17.

### Why the analyze pin

The `created_at` histogram's last bucket ended hours before "now" on the replica, and the planner costed the 24h window as **6,021 rows** (actual ~4.0M, 670× under). At ~4M inserts/day the default `autovacuum_analyze_scale_factor = 0.1` (~14M rows) means days between analyzes. `0.005` (~0.7M rows) re-analyzes several times a day. `ALTER TABLE … SET (autovacuum_…)` takes `SHARE UPDATE EXCLUSIVE` only, and the write is skipped when the option already holds.

### Before / after — observable behavior

```mermaid
sequenceDiagram
    participant R as totals refresher / leaderboard miss
    participant S as store (NetworkTotals / Leaderboard)
    participant PG as provider_earnings (33 GB)
    Note over R,PG: BEFORE — window=24h
    R->>S: aggregate since now-24h (10 s ctx)
    S->>PG: Parallel Seq Scan, whole heap (planner: 6,021 rows)
    PG--xS: context deadline exceeded (>10 s)
    S-->>R: error → 503 / stale cache latch
```

```mermaid
sequenceDiagram
    participant R as totals refresher / leaderboard miss
    participant S as store (NetworkTotals / Leaderboard)
    participant PG as provider_earnings (33 GB)
    Note over R,PG: AFTER — window=24h
    R->>S: aggregate since now-24h (10 s ctx)
    S->>PG: Bitmap Index Scan on created_at BRIN → ~3% of block ranges
    PG-->>S: rows (fresh histogram: ~4M est.)
    S-->>R: totals / board → 200, cached
```

### Before / after — code flow

```mermaid
flowchart TD
    subgraph BEFORE
        A1[migrate] --> B1[ensureProviderEarningsJobIndex]
        B1 --> C1[serve]
        L1[Leaderboard] --> M1["pool.Query (session work_mem 4 MB)"]
    end
```

```mermaid
flowchart TD
    subgraph AFTER
        A2[migrate] --> B2[ensureProviderEarningsJobIndex]
        B2 --> W2["ensureProviderEarningsWindowIndex\nensureConcurrentIndex(BRIN, CONCURRENTLY)\n+ ALTER TABLE SET analyze_scale_factor (if differs)"]
        W2 --> C2[serve]
        L2[Leaderboard] --> M2["withAnalyticsTx: BEGIN READ ONLY\nSET LOCAL work_mem = 1GB → query → COMMIT"]
    end
```

### Deployment

- **First boot after deploy is slower.** `CREATE INDEX CONCURRENTLY` on a 33 GB heap makes two passes; expect the new coordinator to spend a few minutes in `migrate` before it binds `:8080`, exactly as the job index did on its first boot. The build takes no lock that blocks the old blue-green coordinator's writes. Every later boot hits the valid-index fast path.
- **Optional pre-flight to make boot a no-op:** an operator can run the same statement by hand first — `CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_provider_earnings_created_at_brin ON provider_earnings USING brin (created_at);` followed by `ANALYZE provider_earnings;` — after which boot only sets the reloption.
- **Verify on prod after deploy:** `EXPLAIN (ANALYZE, BUFFERS) SELECT count(*) FROM provider_earnings WHERE created_at >= now() - interval '24 hours';` should show `Bitmap Index Scan on idx_provider_earnings_created_at_brin` and an estimate within ~2× of ~4M; `SELECT reloptions FROM pg_class WHERE relname = 'provider_earnings';` should include `autovacuum_analyze_scale_factor=0.005`; `pg_stat_user_tables.last_autoanalyze` should advance several times a day.

### Overlap with #1140

#1140 changes the body of `PostgresStore.Leaderboard` (error propagation); this PR wraps the same body in `withAnalyticsTx`. The two are independent in intent and small; I will rebase whichever lands second.

### Helper rename

`ensureProviderRestoreIndex` → `ensureConcurrentIndex` (and its error strings drop "restore"): the helper was already generic, only its name was restore-specific. Call sites and behavior unchanged; `provider_restore_index_test.go` still covers it.

### Deliberately out of scope

- `window=30d` touches ~76% of the table; no index helps. Needs a time-bucketed rollup.
- `window=all` should read the already-maintained `earnings_summary` table, which needs one additive column to preserve the API's work/reward split. Separate issue/PR.
- Decoupling inbound requests from the refresher's `queryMu`, and alerting on `cache.refresh_failed`.

### Limitations

- Index effectiveness was inferred from the replica's statistics and plans; an index cannot be created on the read replica, so the first real plan confirmation is on `darkbloom-dev` or after the prod deploy (commands above).
- The Postgres-backed tests only run in CI.
- No milestone set: open milestones are `v0.3.6` / `v0.4.0` with zero open items — please assign if a release is targeted.

---
<a href="https://app.blacksmith.sh/Layr-Labs/codesmith/d-inference/pr/1142?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792390562&installation_model_id=21733&pr_number=1142&ref=codesmith_pr_footer&repository=Layr-Labs%2Fd-inference&return_to=https%3A%2F%2Fgithub.com%2FLayr-Labs%2Fd-inference%2Fpull%2F1142&signature=f35d2170fdcf83f2c7acba387181d18c9c158b94dac11fa7522c8274fa358bd1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


brandon-eigenlabs:
Rebased onto `master` at `76a8f03d9` after the v0.9.7 release. Only `CHANGELOG.md` (the entry now sits under the new `## Unreleased` list) and a docs freshness stamp conflicted; the Go changes merged cleanly and `store`/`api` tests, `docs-check` and `docs-impact-check` were re-run on the rebased branch. No functional change.

blocksek:
Rollout question, rather than a merge blocker: will the BRIN index be prebuilt while the current coordinator is still serving, so the replacement can take the valid-index fast path at startup? `CONCURRENTLY` allows writes during the build, but the build still adds database I/O and blocks the new coordinator from becoming ready until it completes. It would be useful to confirm the intended rollout sequence here so we can minimize downtime.

blocksek:
Two questions on the implementation—I may be missing some context:

1. **Leaderboard concurrency with the higher `work_mem`:** Should we add a concurrency bound across leaderboard cache keys before routing these queries through `withAnalyticsTx`? My reading is that the helper's 1 GB per sort/hash budget relies on serialized analytics queries, while `handleLeaderboard` calls the store directly on cache misses, including distinct keys for different `limit` values. Could concurrent misses therefore create significant database memory pressure when the query uses large sorts? I realize `work_mem` is a ceiling rather than an upfront allocation, so the actual impact depends on the plan and workload. Is there an existing concurrency guard or production measurement that makes this safe?

2. **BRIN summarization as the table grows:** Should the index explicitly set `WITH (autosummarize = on)`? The [PostgreSQL 17 documentation](https://www.postgresql.org/docs/17/sql-createindex.html#INDEX-RELOPTION-AUTOSUMMARIZE) lists the default as off, whereas the PR description says it is on. My understanding is that more frequent `ANALYZE` refreshes planner statistics but doesn't summarize new BRIN ranges; those would wait for vacuum or explicit summarization. Do we already have maintenance that keeps those ranges summarized frequently enough for the 24h queries? If not, would it make sense to enable the option for both newly created and precreated indexes, and assert it in the test? The current test manually calls `brin_summarize_new_values`, so it wouldn't catch a missing automatic-maintenance setting.

